PR #59, opened 2020-11-17
Ensure the letters of an all uppercase word are converted to lowercas…

vivien-apple:
…e when using the 'asCamelCased' helper method

#### Problem

When generating declarations/definitions for the `IAS Zone` cluster, the methods name contains `IASZone` while the documentation state `IasZone`.

Similarly the various types in `zcl-builtin/silabs/` expects `IasZone` and not `IASZone`.

####
 * Change the `asCamelCased` helper method to convert the remaining letters after the first one as lowercased.


tecimovic:
This pull request fixes the problem that was also seen in with "ZLL Commissioning" and was fixed on the `silabs` branch earlier. So it's a duplicate at this point.

PR #85, opened 2021-01-12
Add initial network provisoning cluster

erjiaqing:


tecimovic:
Unit tests report:
```
test Silabs zcl data loading in memory

    expect(received).toEqual(expected) // deep equality

    Expected: 109
    Received: 110

      60 |     .then((rows) => expect(rows.length).toEqual(1))
      61 |     .then(() => queryZcl.selectAllClusters(db, packageId))
    > 62 |     .then((x) => expect(x.length).toEqual(109))
         |                                   ^
      63 |     .then(() => queryZcl.selectAllDomains(db, packageId))
      64 |     .then((x) => expect(x.length).toEqual(22))
      65 |     .then(() => queryZcl.selectAllEnums(db, packageId))
```

When you add more data to the test files, you usually need to adjust unit tests, since the unit tests actually validate what is the number of clusters loaded and such.

Please, be aware that the XML files in zcl-builtin are NOT your production XML files, but only the ones used for unit testing and the ones provided by default. Your SDK itself needs to provide final XML files. So you also have a possibility to just add this chip.xml to your SDK, instead of zap itself.

But if you would indeed like to add chip.xml to the unit test suite, then that's ok as well, you just have to adjust the actual numbers in unit tests to the reality of how much of clusters/attributes/commands/etc are now available in the XML files.

erjiaqing:
> Unit tests report:
> 
> ```
> test Silabs zcl data loading in memory
> 
>     expect(received).toEqual(expected) // deep equality
> 
>     Expected: 109
>     Received: 110
> 
>       60 |     .then((rows) => expect(rows.length).toEqual(1))
>       61 |     .then(() => queryZcl.selectAllClusters(db, packageId))
>     > 62 |     .then((x) => expect(x.length).toEqual(109))
>          |                                   ^
>       63 |     .then(() => queryZcl.selectAllDomains(db, packageId))
>       64 |     .then((x) => expect(x.length).toEqual(22))
>       65 |     .then(() => queryZcl.selectAllEnums(db, packageId))
> ```
> 
> When you add more data to the test files, you usually need to adjust unit tests, since the unit tests actually validate what is the number of clusters loaded and such.
> 
> Please, be aware that the XML files in zcl-builtin are NOT your production XML files, but only the ones used for unit testing and the ones provided by default. Your SDK itself needs to provide final XML files. So you also have a possibility to just add this chip.xml to your SDK, instead of zap itself.
> 
> But if you would indeed like to add chip.xml to the unit test suite, then that's ok as well, you just have to adjust the actual numbers in unit tests to the reality of how much of clusters/attributes/commands/etc are now available in the XML files.

ACK, however, since CHIP is still using the xml files inside ZAP to generate code, I guess this is the place for chip related clusters for now.

PR #117, opened 2021-03-04
Update the output format of GENERATED_DEFAULTS from helper-endpointco…

vivien-apple:
…nfig.js to be easier to read


tecimovic:
This change was also in PR 119, which included the unit test fix, so I'm closing this PR as it's redundant at this point.

PR #123, opened 2021-03-24
Sort clusters, commands and attributes in helper-endpointconfig.js

vivien-apple:
#### Problem

While generating `endpoint_config.h` the attributes order is sometimes moving around. This is annoying since it creates some useless noises in PRs and it may hides some local changes.

This PR sorts clusters, attributes and commands to make sure the order stay the same between multiple runs.


vivien-apple:
Closing since @tecimovic did better thing by using `ORDER BY` directly on the SQL queries.

vivien-apple:
I'm reopening this PR since it is still not fixed. `endpoint_config.h` still has some differences between different runs.

PR #130, opened 2021-04-12
Add missing httpPort argument to doOpen in menu.js

vivien-apple:
#### Problem
 * Trying to open an existing ZAP file using the UI menu: `Electron -> Open File...` results into an `undefined httpPort` error.


tecimovic:
This has been fixed in commit 2945775 as well, which came in as part of a different PR.
I'll close this PR.

PR #132, opened 2021-04-12
Fix the github workflow.

tecimovic:


tecimovic:
workflow requires xvfb-run for selfCheck

PR #233, opened 2021-09-22
Use max length for sizing the attribute store for attributes that are…

vivien-apple:
… not strings and does not have a defined atomic size

#### Problem

Some attributes does not have a defined atomic size, but expect strings, this case is ignored into the query used to generate `endpoint_config.h`


vivien-apple:
For reference, this is blocking https://github.com/project-chip/connectedhomeip/pull/9725

brdandu:
> … not strings and does not have a defined atomic size
> 
> #### Problem
> Some attributes does not have a defined atomic size, but expect strings, this case is ignored into the query used to generate `endpoint_config.h`

Could you point me to some use case attributes which do not have atomic size defined?

vivien-apple:
> > … not strings and does not have a defined atomic size
> > #### Problem
> > Some attributes does not have a defined atomic size, but expect strings, this case is ignored into the query used to generate `endpoint_config.h`
> 
> Could you point me to some use case attributes which do not have atomic size defined?

As I mentioned array. While using this code, I also realised that I have to do the same changes for `query-clusters.js` and `query-endpoints.js` otherwise the {{clusterSize}} is wrong.

brdandu:
> > > … not strings and does not have a defined atomic size
> > > #### Problem
> > > Some attributes does not have a defined atomic size, but expect strings, this case is ignored into the query used to generate `endpoint_config.h`
> > 
> > 
> > Could you point me to some use case attributes which do not have atomic size defined?
> 
> As I mentioned array. While using this code, I also realised that I have to do the same changes for `query-clusters.js` and `query-endpoints.js` otherwise the {{clusterSize}} is wrong.

That is correct fixing size of attributes for arrays can lead to changes in the cluster and endpoint queries for calculating size as well. Another question that will need to be answered is how will attribute defaults and bounds be treated for array attributes?

jmeg-sfy (approved):

Issue #243, opened 2021-09-30
Need to check if command contains array argument and check if the particular command argument is array, and there is also array inside


yunhanw-google:
regarding array inside command, we need similar support as you have done for array in struct item,

1. may we have  {{#if command_contains_array}},   where command have multiple argument, and wanna check if there exists array type for arg,  this is pretty similar for what you have done for {{#if struct_contains_array}}

yunhanw-google:
{{#if command_contains_array}}
struct DecodableType {
public:
    {{#zcl_command_arguments checkForDoubleNestedArray="true"}}
    {{#if isArray}}
    DataModel::DecodableList<{{#if struct_item_contains_nested_array}}{{type}}::DecodableType{{else}}{{#if_is_enum type}}{{type}}{{else}}{{asChipZapType type}}{{/if_is_enum}}{{/if}}> {{asLowerCamelCase label}};
    {{else}}
    {{#if command_argument_contains_nested_array}}{{type}}::DecodableType{{else}}{{#if_is_enum type}}{{type}}{{else}}{{asChipZapType type}}{{/if_is_enum}}{{/if}} {{asLowerCamelCase label}};
    {{/if}}
    {{/zcl_command_arguments}}
    CHIP_ERROR Decode(TLV::TLVReader &reader);
};

PR #245, opened 2021-10-02
Add command_contains_array

yunhanw-google:
### Problems:
https://github.com/project-chip/zap/issues/243
### Summary of Changes

--Add {{#if command_contains_array}}, where command have multiple argument, and check if there exists array type for arg


src-electron/db/query-command.js
```diff
@@ -1013,6 +1013,76 @@ ORDER BY COMMAND_REF, FIELD_IDENTIFIER`,
     .then((rows) => rows.map(dbMapping.map.commandArgument))
 }
 
+/**
+ * Retrieves filtered commands for a given cluster Id
```
bzbarsky-apple, line 1017: Filtered by what?

PR #257, opened 2021-10-14
Silabs

tecimovic:
Add helper and documentation for the access control and fabric sensitivity feature.

### DIFF
--- a/docs/access.md
+++ b/docs/access.md
@@ -66,3 +66,35 @@ Example:
 
 ```
 
+## Template helpers
+
+The basic template helper to use is the `{{#access}} ... {{/access}}` iterator. This iterator iterates over all the given access triplets.
+It supports 2 options:
+   * `entity="attribute/command/event"` - if the entity can't be determined from context, this sets the entity type.
+   * `includeDefault="true/false"` - determines if default values are included or not.
+
+Example:
+```
+{{#zcl_clusters}}
+Cluster: {{name}} [{{code}}]
+{{#zcl_attributes}}
+  - attribute: {{name}} [{{code}}]
+  {{#access entity="attribute"}}
+    * Op: {{operation}} / Role: {{role}} / Modifier: {{accessModifier}}
+  {{/access}}
+{{/zcl_attributes}}
+{{#zcl_commands}}
+  - command: {{name}} [{{code}}]
+  {{#access entity="command"}}
+    * Op: {{operation}} / Role: {{role}} / Modifier: {{accessModifier}}
+  {{/access}}
+{{/zcl_commands}}
+{{#zcl_events}}
+  - event: {{name}} [{{code}}]
+  {{#access entity="event"}}
+    * Op: {{operation}} / Role: {{role}} / Modifier: {{accessModifier}}
+  {{/access}}
+{{/zcl_events}}
+
+{{/zcl_clusters}}
+```


tecimovic:
@tcarmelveilleux : This PR completes all the access control work for now.
There is an example in `access.zapt`. There is documentation in `access.md`.

Test it out, and tell me how it does or does not meet your needs.

docs/access.md
```diff
@@ -0,0 +1,100 @@
...
+## Default permissions
+
+ZCL entities can define their own individual permissions. However, there is also a global definition of default permissions for given types. These are assumed for the given entity, unless is provides any specific permissions of its own.
+
+Default permissions are declared via a `<defaultAccess>` tag at the toplevel of the XML.
+
+Example:
+```
+  <defaultAccess type="command">
+    <access op="invoke"/>
```
tcarmelveilleux, line 56:
```
<access op="invoke" role="operate"/>
```

docs/access.md
```diff
@@ -0,0 +1,100 @@
...
+  </accessControl>
+
+```
+
+This example defines three operations, _read_, _write_ and _invoke_, two modifiers and four roles.
+
+## Access triplets
+
+Each individual access condition can be defined with an access triplet in the XML.
+Access triplet is a combination of an _operation_, _role_ and _modifier_. They are optional, so you can only have one of these. A missing part of triplet generally means permissivenes, but that's implementation specific for the given SDK. An entity, that defines it's access can have one or more access triplets.
```
tcarmelveilleux, line 36: Add a note that not all combinations are valid

docs/access.md
```diff
@@ -0,0 +1,100 @@
...
+    <attribute side="server" code="0x0000" define="AT1" type="INT64U" writable="false" optional="true">
+      <description>at1</description>
+      <access op="write" role="manage" modifier="fabric-scoped"/>
+    </attribute>
+```
+This is a definition of an attribute that has an access triplet, declaring it allows _write_ operation by a _manage_ role, with _fabric-scoped_ modifier applied.
+
+## Default permissions
+
+ZCL entities can define their own individual permissions. However, there is also a global definition of default permissions for given types. These are assumed for the given entity, unless is provides any specific permissions of its own.
```
tcarmelveilleux, line 49: `ZCL entities can define their own individual acces privilege requirements`

docs/access.md
```diff
@@ -0,0 +1,100 @@
...
+
+ZAP supports access control on all ZCL entities. It's down to the implementation of the SDK to map
+these features to the required and supported access control SDK features. ZAP generally just provides a data model and a mechanism to encode it in the meta-info files, and propagate that data to the generation templates, without assigning specific meanings to the data points.
+
+## Base terms
+
+ZAP access control defines three base terms:
+
+   1. **operation** : it's something that can be done. Example: read, write, invoke.
+   2. **role**: it's a privilege of an actor. Such as "View privilege", "Administrative rol", etc.
```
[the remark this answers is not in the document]
tecimovic: Note that ZAP is not the tool that 100% maps to Matter. Zap is too that needs to map to a superset of Matter + Zigbee + UIC project + any other projects that use ZCL. So if Matter chooses to call this privilege, and UIC maybe calls it "management level" and something else calls it something else.

I can appreciate that Matter calls it "privilege", but for a non-English native speaker, there is about 200 ways how `privilege` can be misspelled, but `role` is a very simple word, with much less abilities to be misspelled.

I can give aliases to this in `XML` so you can type "privilege". 

But again: Zap is NOT just a Matter tool. It's ALSO a Matter tool. It needs to suport of superset of all it's uses.

docs/access.md
tecimovic: Again, this was just example. No specific relationships to Matter spec here.

docs/access.md
```diff
@@ -0,0 +1,100 @@
...
+ZCL entities can define their own individual permissions. However, there is also a global definition of default permissions for given types. These are assumed for the given entity, unless is provides any specific permissions of its own.
+
+Default permissions are declared via a `<defaultAccess>` tag at the toplevel of the XML.
+
+Example:
+```
+  <defaultAccess type="command">
+    <access op="invoke"/>
+  </defaultAccess>
+  <defaultAccess type="cluster">
```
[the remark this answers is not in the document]
tecimovic: Same thing: Zap supports default access for clusters. But if Matter chooses to not apply them, it's free to do so.

Issue #258, opened 2021-10-14
Need support for nullable types


bzbarsky-apple:
Matter has nullable types.  ZAP does not support this, and needs to.

tecimovic:
@bzbarsky-apple : if the `isNullable="true/false"` is missing in the XML, what is the default? False or true?

PR #259, opened 2021-10-16
Silabs

tecimovic:
1.) https://github.com/project-chip/zap/issues/258

Implement nullability.

2.) Update versioning to YYYY.MM.DD for main package.json version.

3.) Some pre-commit git hook work.


src-electron/db/db-mapping.js
```diff
@@ -155,6 +157,7 @@ exports.map = {
       argIsArray: dbApi.fromDbBool(x.ARG_IS_ARRAY),
       argPresentIf: x.ARG_PRESENT_IF,
       argCountArg: x.ARG_COUNT_ARG,
+      argIsNullable: x.ARG_IS_NULLABLE,
```
bzbarsky-apple, line 160: This is true if any arg is nullable?

Not sure we will need this, but ok.

src-electron/db/query-command.js
```diff
@@ -724,6 +724,7 @@ async function selectAllCommandsWithArguments(db, packageId) {
       argFieldId: x.FIELD_IDENTIFIER,
       argIsArray: x.ARG_IS_ARRAY,
       argPresentIf: x.ARG_PRESENT_IF,
+      argIsNullable: x.ARG_IS_NULLABLE,
```
bzbarsky-apple, line 727: Why do we not need `dbApi.fromDbBool` here?

src-electron/db/zap-schema.sql
```diff
@@ -186,6 +186,7 @@ CREATE TABLE IF NOT EXISTS "COMMAND_ARG" (
   "TYPE" text,
   "IS_ARRAY" integer,
   "PRESENT_IF" text,
+  "IS_NULLABLE" integer,
```
bzbarsky-apple, line 189: Why integer here, not boolean?

src-electron/db/zap-schema.sql
```diff
@@ -257,6 +259,7 @@ CREATE TABLE IF NOT EXISTS "ATTRIBUTE" (
   "IS_SCENE_REQUIRED" integer,
   "IS_OPTIONAL" integer,
   "IS_REPORTABLE" integer,
+  "IS_NULLABLE" integer,
```
bzbarsky-apple, line 262: And here, why integer?  Os is the boolean in the EVENT_FIELD table the wrong one?

bzbarsky-apple (approved):
Generally looks sane.  I don't have a clear enough grasp of the DB schema and queries to tell whether all the places that needed updating got updated, but at least the use cases I am interested in seem to be covered in the mapping functions....

I figure once this is in we can try adding things to TestCluster that exercise this, and if there's something missing we will notice at that point.

src-electron/db/db-mapping.js
tecimovic: This is part of a join, and it's note called "IS_NULLABLE", but "ARG_IS_NULLABLE" to bypass possible naming conflicts.

src-electron/db/query-command.js
tecimovic: We do need it. We're not the most disciplined, because in all reality, the `1` and `0` that is the database boolean, automagically maps into JavaScript world just fine, because JavaScript was designed that way, but it makes me uncomfortable still, so I like to have `fromDbBool()` wherever I can.
I'll add it in.

src-electron/db/zap-schema.sql
tecimovic: There is no BOOLEAN types in SQLITE. Boolean are by convention integers saved as 1 or 0.

src-electron/db/zap-schema.sql
tecimovic: Booleans in the EVENT_FIELD are wrong. I'll change them to 'integer'. Why did they work in the first place, no idea.

Reading https://www.sqlite.org/datatype3.html, there is no separate boolean storage class. However sqlite is very fuzzy and supports a lot of data type aliases and such. I suspect 'boolean' magically works as an alias for `integer`.

Either way, I like to keep the raw storage classes, not the aliase in the schema, so I'll replace the booleans there into integers.

tecimovic:
Added fix to https://github.com/project-chip/zap/issues/260 to this PR and added @yunhanw-google as a reviewer.

src-electron/db/zap-schema.sql
tecimovic: Even though, now that I'm thinking about this some more.... Maybe it's more correct to go and fix all of the actual "boolean" fields to be 'boolean', and not 'integer', and trust the database? It's probably more portable against other DB engines to use the "data type" and not the "storage class" from sqlite.....

PR #264, opened 2021-10-20
Stop assuming that attribute types are always lowercased.

bzbarsky-apple:
I've verified that with these changes Matter codegen is unaffected by
dropping our forced lowercasing of attribute types from
zcl-loader-silabs.js.  Not including that change for now, since it
might break Zigbee bits.


tecimovic:
In dotdot loading, there are four types that are not all lowercase:
   `ToD`, `UTC`, `bacOID` amd `EUI64`.

PR #274, opened 2021-10-27
Silabs

tecimovic:
- Performance enhancements: this shaves of about 25% of generation time for the ZigbeePro stack.
- Add cascading deletes to many tables that didn't have this set correctly


tecimovic:
Rebased against master before merge.

PR #295, opened 2021-11-04
Fix little-endian long default values.

bzbarsky-apple:
The values we spit out for long defaults just take the incoming hex
string and dump it into the file.  But for little-endian we need to
reverse the hex bytes for at least integers and floats and enums and
bitmaps... and generally everything that is not strings.


tecimovic (approved):
The helper-endpointconfig.js is used solely by Matter now anyway. ZigbeePro stacks has divested itself from it, and is using templates that are not using these at all any more. UIC never needed that to begin with. So if changes here work for Matter, then they're good changes.

bzbarsky-apple:
I've verified that for Matter this is doing the right thing.

PR #301, opened 2021-11-11
Silabs

tecimovic:
- Fix an importing problem with using "changes" count mistakenly for an attribute ID.
- Update some UI for custom XML page.
- Few other bugfixes.


bzbarsky-apple (approved):
Looks reasonable, though I skimmed a lot of the changes (esp. the .vue bits).

PR #304, opened 2021-11-12
Allocate the mimimum amount of space necessary for string default values

bzbarsky-apple:
emAfLoadAttributeDefaults uses emAfReadOrWriteAttribute to write the
value from defaults into the attr store, and that ends up calling
typeSensitiveMemCopy, which for strings already checks the length and
only copies data up to that length.  That means that for strings we
only need to allocate enough space in the default store for the actual
string length+value, not for all possible values that fit in the
typeSize.

### DIFF
--- a/src-electron/generator/helper-endpointconfig.js
+++ b/src-electron/generator/helper-endpointconfig.js
@@ -335,7 +335,7 @@ function endpoint_attribute_min_max_list(options) {
       (min >= 0 ? '' : '-') + '0x' + Math.abs(min).toString(16).toUpperCase()
     let maxS =
       (max >= 0 ? '' : '-') + '0x' + Math.abs(max).toString(16).toUpperCase()
-    ret += `  { (uint8_t*)${defS}, (uint8_t*)${minS}, (uint8_t*)${maxS} }${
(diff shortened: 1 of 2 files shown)


vivien-apple (approved):

tecimovic (approved):

tecimovic (changes requested):
I think this trips one of the unit tests.

```
      193 |       '#define FIXED_ENDPOINT_ARRAY { 0x0029, 0x002A, 0x002B }'
      194 |     )
    > 195 |     expect(epc).toContain(
          |                 ^
      196 |       "17, 'V', 'e', 'r', 'y', ' ', 'l', 'o', 'n', 'g', ' ', 'u', 's', 'e', 'r', ' ', 'i', 'd', 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,"
      197 |     )
      198 |     expect(epc).toContain(

```

Probably expected change, we just have to fix the unit test.

bzbarsky-apple:
@tecimovic Indeed, the test expectations needed to be updated.  Test fixed.

tecimovic (approved):

Issue #298, opened 2021-11-05
UI cleanup within cluster cleanup


brdandu:
When we open 'configure' on a cluster, on the top we see something like:
```
< Endpoint 1 / General / Basic
Basic
Cluster ID: 0x0000, Enabled for Server
```

Here `< Endpoint 1 / General` is a hyperlink which seems weird. Would it make sense to make it look something as follows:
```
< Back
Endpoint 1 / General / Basic Cluster
Cluster ID: 0x0000, Enabled for Server
```
Where just back is a hyperlink which takes you back

tecimovic:
I didn't like what we had before, @brdandu . But I don't like what we have now either.... :(

But I don't have a better idea at this point....

brdandu:
> I didn't like what we had before, @brdandu . But I don't like what we have now either.... :(
> 
> But I don't have a better idea at this point....
> 
> Isn't there some standard practice how these should look? Doesn't every website in existence have some of these breadcrumbs? We should just follow what everyone else does....

Yeah I wish it looked cleaner. That blob looked off to me. If there are any better suggestions then I am open to it.

Issue #293, opened 2021-11-01
Inconsistent reference to manufacturer Code / ID / Id


tecimovic:
In various places where Manufacturer ID is mentioned or shown as a column heading, the text is inconsistent in its use of Code v. ID v. Id.

Per CSA's ZCL specification and Manufacturer Code document, "Code" is the correct terminology.

Issue #292, opened 2021-11-01
Cluster ID / Id


tecimovic:
In various places where Cluster ID is mentioned or shown as a column heading, the text is inconsistent in its use of ID v. Id.

SaharKargar:
@tecimovic
Dear Timotej, I had a quick question just to make sure we are on the same page.
So there are some places that we have "cluster Id" as the text of column headings.
We just have to change it to "Cluster ID" right?

tecimovic:
Looking at the ZCL spec, they actually pretty consistently use the full work "Identifier" and are not trying to abbreviate into "ID" or "Id". Let me run this by some folks that are closer to documentation and such...

tecimovic:
I ran this question with the documentation folks...

We should always use "ID" as in all uppercase.
But if we have real estate we can use a full word "identifier".

So for sake of simplicity, it should be "Cluster ID", "Command ID", "Attribute ID", etc for now.
If you are updating something that is a full sentence, as in some paragraph, then we could use "identifier" instead of the abbreviation. Make that decision based on real estate.

Issue #291, opened 2021-11-01
Cluster filter - cosmetic


tecimovic:
The Cluster Filter dropdown has values No Filter (capital F), All Clusters, and Only Enabled.

- Clicking the CLOSE ALL button sets dropdown text to "No filter" (lowercase F).
- "Only Enabled" probably should be "Enabled Clusters".

PR #306, opened 2021-11-12
resolving ui issues

Mehradml:
The resolved issues are as follows:
- https://github.com/project-chip/zap/issues/287
- https://github.com/project-chip/zap/issues/282
- https://github.com/project-chip/zap/issues/271
- https://github.com/project-chip/zap/issues/272
- https://github.com/project-chip/zap/issues/281
- https://github.com/project-chip/zap/issues/283
- https://github.com/project-chip/zap/issues/284
- https://github.com/project-chip/zap/issues/285
- https://github.com/project-chip/zap/issues/288
- https://github.com/project-chip/zap/issues/290
- https://github.com/project-chip/zap/issues/291
- https://github.com/project-chip/zap/issues/292
- https://github.com/project-chip/zap/issues/293
- https://github.com/project-chip/zap/issues/298


src/boot/ws.js
```diff
@@ -104,17 +104,21 @@ onWebSocket(dbEnum.wsCategory.init, (data) =>
   console.log(`Init message received: ${data}`)
 )
 
-onWebSocket(dbEnum.wsCategory.tick, (data) =>
-  console.log(`Tick received: ${data}`)
-)
```
brdandu, line 109: Can we tie these to some debug flag so we automatically get these when we turn on the debug level. I remeber Timotej did something like this. Same goes for all the other commented out code in this file

src/components/ZclAttributeReportingManager.vue
```diff
@@ -220,6 +332,18 @@ export default {
     isRowDisabled(attributeId) {
       return !this.editableAttributesReporting[attributeId]
     },
+    displayAttrWarning(row) {
+      // TODO display proper warnings
```
brdandu, line 336: Is there some todo here remaining?

src/components/ZclCommandManager.vue
```diff
@@ -185,7 +187,13 @@ export default {
...
                 (this.selectionServers.includes(this.selectedCluster.id) &&
                   row.source == 'server'))
-      && !this.selectionOut.includes(this.hashCommandIdClusterId(row.id, this.selectedCluster.id))
+      && !this.selectionOut.includes(this.hashCommandIdClusterId(row.id, this.selectedCluster.id)) ||
+        this.isCommandRequired(row) && (
+        (this.selectionClients.includes(this.selectedCluster.id) &&
+                  row.source == 'server') ||
+                (this.selectionServers.includes(this.selectedCluster.id) &&
+                  row.source == 'client'))
+      && !this.selectionIn.includes(this.hashCommandIdClusterId(row.id, this.selectedCluster.id))
```
brdandu, line 196: What is this above change doing?

src/boot/ws.js
tecimovic: We have a good logging on back-end tied to loglevels. I don't even know how this is done on front-end.

Would it make sense for back-end to control this? 

Or does it make more sense to this fully in front-end?

tecimovic: BTW: This is a cleanup. We should not hold up the PR for this.

tecimovic:
https://github.com/project-chip/zap/issues/287: See my comment on the bug. There is a weird jump.
https://github.com/project-chip/zap/issues/282: Same problem as above. When table is too small, there is a jump. (I noticed that the jump doesn't happen if the scrollbar is not visible, so jump is only if the table is long enough to cause a scrollbar.)

tecimovic (approved):

src/components/ZclCommandManager.vue
tecimovic: One of the bugs was that the command handling needs the warnings if you disable mandatory command. I don't know what the code does, but I clicked around there and it worked for me. So whatever the code does, is not doing too much damage. :)

src/boot/ws.js
brdandu: My main concern was that if that was added by someone for some error monitoring. So I was wondering if we will need that again. May be have the front end show this for now. We could come up with a way to show all issue in one place later?

bzbarsky-apple (approved):
Rubberstamping, generally....

PR #315, opened 2021-11-18
Add a way to configure the default reportability of attributes.

bzbarsky-apple:
This adds a "defaultReportable" property to the metadata file and uses
that to decide whether the "reportable" XML attribute defaults to true
of false when it's not set.

### DIFF
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -88,6 +88,11 @@ async function collectDataFromJsonFile(metadataFile, data) {
     returnObject.featureFlags = obj.featureFlags
   }
 
+  // Default reportability
+  if ('defaultReportable' in obj) {
+    returnObject.defaultReportable = obj.defaultReportable
+  }
+
   returnObject.version = obj.version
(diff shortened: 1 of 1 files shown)


bzbarsky-apple:
@tecimovic Review, please?

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1483,7 +1502,7 @@ async function loadSilabsZcl(db, metafile, isJson = false) {
     if (ctx.version != null) {
       await zclLoader.recordVersion(db, ctx.packageId, ctx.version)
     }
-    await parseZclFiles(db, ctx.packageId, ctx.zclFiles)
+    await parseZclFiles(db, ctx.packageId, ctx.zclFiles, ctx)
```
tecimovic, line 1505: If we're passing the ctx as a final argument, then we should change the api to just get db and ctx, no need to have extra two separate arguments for ctx.packageId and ctx.zclFiles.

Minor nitpick. Not holding up the PR.....

tecimovic (approved):

PR #333, opened 2021-11-30
Silabs

tecimovic:
- enable config equalization across endpoints: initial support for both Matter and ZigbeePro
- clean up some package ID matching.
- add xml validation to Jenkins CI
- cleanups.

### DIFF
--- a/src/util/util.js
+++ b/src/util/util.js
@@ -26,10 +26,16 @@ export function cantorPair(x, y) {
   return ((x + y) * (x + y + 1)) / 2 + y
 }
 
-export function asHex(value, padding) {
-  if (value == null) return ''
-  if (value < 0) return value.toString()
-  return '0x' + value.toString(16).padStart(padding, '0').toUpperCase()
+export function asHex(input, padding) {
+  if (input == null) return ''
+  if (input < 0) return input.toString()
+  let output = input
+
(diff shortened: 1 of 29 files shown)


src/util/util.js
```diff
@@ -26,10 +26,16 @@ export function cantorPair(x, y) {
   return ((x + y) * (x + y + 1)) / 2 + y
 }
 
-export function asHex(value, padding) {
-  if (value == null) return ''
-  if (value < 0) return value.toString()
-  return '0x' + value.toString(16).padStart(padding, '0').toUpperCase()
+export function asHex(input, padding) {
+  if (input == null) return ''
+  if (input < 0) return input.toString()
```
bzbarsky-apple, line 31: Worth a comment that the `<` will trigger a ToNumber conversion on `input`, so this is fine even if input is a string.

bzbarsky-apple (approved):

src/util/util.js
tecimovic: @jingteng25742 ^^^ Please note this comment from Boris. I am not 100% certain what you indented here, but I know we have occasional bug around showing things like `0X0x0104` and weird stuff like that.

I'll merge this in regardless for now, since this is not making things much worse, but please do check that.

PR #334, opened 2021-11-30
Add timed write/invoke booleans for attributes and commands.

bzbarsky-apple:
Adds to database schema, XML import, various queries.


tecimovic:
Are these things defined by spec, or is this implementation specific?

If the spec itself defined these "timed write"s, then ZCL XML is the right home for it.
If, however, these are purely implementation specific flags, then this should be listed as the SDK extensions.

tecimovic:
Note: we are planning to clean up the SDK extensions vs core properties sometimes in Q1 22 or so, so even if this is technically SDK extensions, and not defined by spec, we could temporarily keep it here, just like we do the "storage options" for example, which would also technically have to be SDK extensions.

So this question is not necessarily blocking the PR for expediency sake.

tecimovic (approved):
I'll regenerate the schema SVG independently.
(For the record: there is a `make-schema-diagram` script under `doc/` that regenerates the SVG, but it uses something called `schemacrawler`, which is some java package that can generate a DOT format of the schema, which is then converted to SVG using the standard `dot` tooling.... So I'm ok with generating this myself, since I don't want everyone to suffer from having to make sure all that tooling is installed locally....)

bzbarsky-apple:
> Are these things defined by spec, or is this implementation specific?

The spec defines this.  See "7.6.10. Timed Interaction" for the relevant quality.

PR #337, opened 2021-12-02
[WIP] Add struct_is_fabric_scoped to zcl_structs

erjiaqing:
In the spec, we have

Fabric scoped Attribute data SHALL be encapsulated in one of two forms:
- A struct data type containing a fabric-idx data type field
- A list of fabric scoped struct data type entries

Fabric scoped events SHALL be encapsulated in one of two forms:
- A fabric-idx data type field at the top-level of the event definition
- A struct field within the event containing a fabric-idx data type field


yunhanw-google (commented):
please add test for type: fabric_index, 
https://github.com/project-chip/zap/blob/master/test/resource/meta/test1.xml
https://github.com/project-chip/zap/blob/master/test/resource/meta/test1.zapt

> @erjiaqing: The tests are updated, struct.zapt & struct.xml are updated since fabric_scoped is for structs.

tecimovic:
This PR is failing the SonarCloud action, which is currently happening because of some missconfiguration, it's no fault of this PR itself.
I have disabled the action temporarily, so if you can rebase this PR, it should clean up.

bzbarsky-apple (approved):

PR #356, opened 2021-12-15
Allow negative defaults for attributes

vivien-apple:
#### Problem
Negative integer can not be used as defaults, since the `-` character is not taken into account.
Also ` ` is a valid `float` based on the current regex. I'm not sure if that's an expected assertion or not.

#### Changes
 * Make sure that `-` is taken into account
 * Update the `float` regex to not allow ` `, nor `.9` nor `5.` but only `1`or `1.2`.

Additionally `endpointTypeIdList` is not defined into the request body for me when I try to edit an attribute.
Since `endpointTypeId` is defined I changed the code to understand both, but I suspect that there is some other root issue.

#Fixes 327

### DIFF
--- a/src-electron/rest/user-data.js
+++ b/src-electron/rest/user-data.js
@@ -144,6 +144,7 @@ function httpPostAttributeUpdate(db) {
   return async (request, response) => {
     let {
       action,
+      endpointTypeId,
       endpointTypeIdList,
       id,
       value,
@@ -156,7 +157,11 @@ function httpPostAttributeUpdate(db) {
     } = request.body
 
     if (!Array.isArray(endpointTypeIdList) || !endpointTypeIdList.length) {
-      response.status(StatusCodes.BAD_REQUEST).json()
+      if (endpointTypeId == 'undefined') {
+          response.status(StatusCodes.BAD_REQUEST).json()
+      } else {
+          endpointTypeIdList = [endpointTypeId];
+      }
     }
 
     let paramType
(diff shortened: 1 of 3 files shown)


bzbarsky-apple (approved):

vivien-apple:
Seems like the test expects ´.0001´ to be valid. I will update the regexp then.

tecimovic:
Do we need regexes? Can't we just simply use JS native parseInt() and parseFloat() and check if they return NaN.

Something like:
 return ( parseInt(input) != NaN)
and
  return ( parseFloat(input) != NaN)

(I did not specifically look at what exactly you do there, but personally I am 100% not fan of regexes... I don't even know where the original ones came from...)

vivien-apple:
> Do we need regexes? Can't we just simply use JS native parseInt() and parseFloat() and check if they return NaN.
> 
> Something like: return ( parseInt(input) != NaN) and return ( parseFloat(input) != NaN)
> 
> (I did not specifically look at what exactly you do there, but personally I am 100% not fan of regexes... I don't even know where the original ones came from...)

I have updated the code to use `Number` instead of regexes where possible. I have kept some regexpes for hex strings, since the code allows things such as `AABBCC` without the `0x` prefix.

About `parseInt` and `parseFloat`, they comes with their own sets of issues, such as allowing `10 foxes`

tecimovic:
If someone types in `10 foxes`, I hope one of them bytes them! 
Nothing is easy..... :(

PR #357, opened 2022-01-03
Cypress init

Mehradml:
Initiated cypress end-to-end test implementation and also added two test cases.


tecimovic:
Doing a "rebase and merge", because these commits are not rebased.

PR #358, opened 2022-01-04
updated build.yml to use pull_request_target

wehale:
Testing sonarcloud integration to see if it works with this slight modification


tecimovic:
This commit already journeyed to master via the other silabs PR that happened.

PR #362, opened 2022-01-11
Silabs

tecimovic:
- remove erroneous `technotree.com` registry references.
- fix the "ADD" button on the custom XML UI.


tecimovic:
This fixes a weird problem with the tecnotree.com sneaking into the registry URLs.

PR #377, opened 2022-01-19
Fix field id defaulting for Matter.

bzbarsky-apple:
Matter field ids start at 0, not 1, but the defaulting was defaulting
to starting at 1.

The new setup defaults the first field to 0, and every following field
to 1 more than the previous one (which might have an explicitly
specified field id).

### DIFF
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -219,14 +219,15 @@ function prepareBitmap(bm) {
   }
   if ('field' in bm) {
     ret.fields = []
-    bm.field.forEach((field, index) => {
+    let lastFieldId = -1;
+    bm.field.forEach((field) => {
+      let defaultFieldId = lastFieldId + 1;
+      lastFieldId = field.$.fieldId ? parseInt(field.$.fieldId) : defaultFieldId;
       ret.fields.push({
         name: field.$.name,
         mask: parseInt(field.$.mask),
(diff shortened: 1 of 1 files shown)


src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -804,11 +809,14 @@ function prepareStruct(struct) {
   }
   if ('item' in struct) {
     ret.items = []
-    struct.item.forEach((item, index) => {
+    let lastFieldId = -1;
```
tecimovic, line 812: Not a big deal, but for some or the other reason, your editor is not picking the code formatting. We have at some point mandated the no-semicolons and at least in my case VSCode and Prettier plugins with husky git hooks are reformatting this automatically all the time. I am not sure how you avoided that....

But like I said: not a big deal. These gets weeded out sooner or later when someone saves these files in their editors.

tecimovic (approved):

tcarmelveilleux (approved):

src-electron/zcl/zcl-loader-silabs.js
bzbarsky-apple: Fwiw, the "no semicolons" format is full of terrible footguns, so I'm surprised you are using it...

Also, clearly these things are not weeded out in general: tons of files have things that husky catches as formatting errors if I run it.  It might be worth doing a single mass-reformat to fix them all.

PR #384, opened 2022-01-21
Add a storage policy for attributes.

bzbarsky-apple:
Some SDK implementations want to force some specific storage policy
for attributes depending on implementation decisions.  Add some
mechanisms for doing that:

* Add a way to force all lists to have an "attributeAccessInterface"
  storage policy.
* Add a way to provide a map of attributes that have an
  "attributeAccessInterface" storage policy.

For now this policy is mapped to "External" storage for endpoint
attributes, but later on I want to add another value to that enum and
use it instead of "External", because there are some optimizations we
can do for "attributeAccessInterface" that we can't do for things that
are just "External".

### DIFF
--- a/src-electron/db/query-impexp.js
+++ b/src-electron/db/query-impexp.js
@@ -434,7 +434,8 @@ async function importAttributeForEndpointType(
   let selectAttributeQuery = `
 SELECT 
   A.ATTRIBUTE_ID,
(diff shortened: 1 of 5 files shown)


tecimovic (approved):

tcarmelveilleux:
LGTM

PR #388, opened 2022-01-24
Bug/helpers for outgoing commands/GitHub#342

brdandu:
Block helpers for outgoing commands. This PR is needed for the non determinism issue. These changes will be needed by the CHIPClusters-src.zapt and CHIPClusters.zapt as per my understanding.
The way I understood those templates is that outgoing command signatures are generated for here.


brdandu:
This PR is required for https://github.com/project-chip/connectedhomeip/pull/13871

tecimovic (approved):

tecimovic:
@tcarmelveilleux , @bzbarsky-apple , @vivien-apple :
   Please review. This needs to get pushed into the `master` to unblock another PR @brdandu is working on on the main CHIP repo, which should be a first step towards solving the non-deterministic behavior.

PR #401, opened 2022-02-02
Adding if_command_has_response for MatterIDL template non determinism

brdandu:
Adding if_command_has_response to check if a reponse command exists for a given command

Github#342

### DIFF
--- a/src-electron/generator/helper-command.js
+++ b/src-electron/generator/helper-command.js
@@ -408,7 +408,10 @@ async function if_command_fixed_length(commandId, options) {
 }
 
 /**
- * If helper that checks if a reponse command is present for a given command
+ * If helper that checks if a response command is present for a given command
+ * Note: Deprecate helper once https://github.com/project-chip/zap/issues/402
+ * is resolved. Use {{#if responseName}} in templates once the issue is
+ * resolved.
  *
  * example:
  * {{#if_command_has_response command 'Request'}}
@@ -421,6 +424,7 @@ async function if_command_fixed_length(commandId, options) {
  * @param replaceKeyWord Check for response command after removing the string
  * from command name
  * @returns content based on whether the command response was found or not
+ *
  */
 async function if_command_has_response(commandName, replaceKeyWord, options) {
   let commandResponse = commandName.replace(replaceKeyWord, '') + 'Response'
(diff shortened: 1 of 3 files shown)


src-electron/generator/helper-command.js
```diff
@@ -407,6 +407,36 @@ async function if_command_fixed_length(commandId, options) {
   return options.fn(this)
 }
 
+/**
+ * If helper that checks if a reponse command is present for a given command
```
bzbarsky-apple, line 411:
```suggestion
 * If helper that checks if a response command is present for a given command
```

src-electron/generator/helper-command.js
```diff
@@ -407,6 +407,36 @@ async function if_command_fixed_length(commandId, options) {
...
+ * {{else}}
+ * command has no reponse
+ * {{/if_command_has_response}}
+ *
+ * @param commandName Name of the command
+ * @param replaceKeyWord Check for response command after removing the string
+ * from command name
+ * @returns content based on whether the command response was found or not
+ */
+async function if_command_has_response(commandName, replaceKeyWord, options) {
```
bzbarsky-apple, line 425: But this ignores commands that have a response defined via responseName, no?   As in, there are commands with responses that this function returns false for....

brdandu: That is correct. I added responseName to selectAllIncomingOrOutgoingCommandsForCluster such that helpers calling this function can have direct access to responseName which can be used in a #if helper. I have filed https://github.com/project-chip/zap/issues/402 for resolving command responses cleanly.

bzbarsky-apple: I guess I don't understand how one would use this helper in practice.  Is the idea that you check for `responseName` first, and if falsy _then_ use this helper?

I guess this is all kinda broken, in that this has logic that needs to match logic in another repo, etc, etc...  Ah, well.

bzbarsky-apple (approved):

src-electron/generator/helper-command.js
brdandu: Yes one needs to check responseName first and then use this.

tecimovic (approved):
Should be rebased.

PR #404, opened 2022-02-03
Remove the response-matching heuristic.

bzbarsky-apple:
Matter commands now all specify explicit response="" in XML, so
we can do all matching by RESPONSE_NAME.

Also, expose RESPONSE_NAME in command queries so consumers can use it,
now that it's reliable.


tecimovic (approved):

tecimovic:
This changes some data for ZigbeePro, but it's the data that is not really needed, since ZigbeePro never cared about what's request and response anyway. So dismantling that is ok. An explicit "response=" in the XML is cleaner and better than the speculative logic based on names anyway.

PR #414, opened 2022-02-10
Silabs

tecimovic:
- fix several issues related to custom XML loading
- add the back-end code to support UI options from zcl.json
- some cleanup and other minor fixes


src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1281,6 +1295,32 @@ async function parseFeatureFlags(db, packageId, featureFlags) {
   )
 }
 
+/**
+ * Inside the `zcl.json` can be a `featureFlags` key, which is
+ * a general purpose object. It contains keys, that map to objects.
+ * Each key is a "package option category".
+ * Key/velues of the object itself, end up in CODE/LABEL combinations.
```
bzbarsky-apple, line 1302:
```suggestion
 * Key/values of the object itself, end up in CODE/LABEL combinations.
```

bzbarsky-apple (approved):
More or less rubberstamping.  Would love to understand what the actual ui options stuff looks like, though....

PR #427, opened 2022-02-26
Silabs

tecimovic:
- add allowances for node 16 and prefer support for node 16 over continuing support for node 12. Node 14 is still most tested against. In reality, it should work for all three versions, but if we make decisions from now on, they will be made towards support of node 16 rather than continuing support for node 12. Node 14 should still remain fully supported.
- Cleanup some Cypress issues
- Clean up few minor UI issues with dialogs.
- allow for shutdown timer to be configurable, to assist CI a bit more.
- cleanup few binary artifact build issues

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ apack_zap.package
 # Backup files
 *~
 
+# Cypress screenshots and videos
+cypress/screenshots/**
+cypress/videos/**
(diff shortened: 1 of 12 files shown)


tecimovic:
Note: nothing here affects generation, so not expecting or looking for reviews from @bzbarsky-apple or @tcarmelveilleux , but if you want to review some of this, go for it.

Issue #153, opened 2021-05-16
Redesign executable layout: separate electron and backend processes


tecimovic:
**Problem**

ZAP can be used directly from source via various scripts that do NOT require DISPLAY, because they bypass 'electron' and just use plain 'node' engine to perform generation and such.

However, once you use a binary package of zap (.deb, or .rpm) as downloaded from here, you loose this ability.

Binary package ALWAYS uses Electron, so the moment you start zap in a headless environment, without DISPLAY set, you get a core dump and nothing happens.

Workaround is easy: just wrap the execution using `xvfb-run`. 
However that is a bit sad and should be fixed.

**Solution**

Make the binary build produce 2 executables: 
- zap, which is what we know now: the full zap tool including ui and all.
- zap-headless, which is the generation and conversion and all the hadless parts of zap, and NO UI bundled in, so it does not require electron.

A user will then be able to run `zap` for EVERYTHING, or `zap-headless` just for headless purpose.

tecimovic:
Ok, we have a better design now:
1.) Split up the server executable and the UI executable.
   - Server executable becomes a full daemon running all the server aspects.
   - UI executable is then an electron container, essentially just serving as a simple browser to the server.

tecimovic:
I'll use this bug as a placeholder for the redesign that will achieve the following endgoal:

Zap will be shipped as TWO separate executables.

1.) `zap` will be a pure node.js executable. There will be no electron there. The executable will perform everything zap does now, with the same command lines, except that all the UI options will be taken out. It will run as "server" by default.

2.) `zapui` will be an electron executable. It WILL NOT CONTAIN any of the logic `zap` has. It WILL NOT open a database. IT WILL NOT query the database. It will only do the following:
   a.) When needed it will execute `zap` server.
   b.) It will open the new window, or a separate window to edit the files provided.
   c.) It will run in "electron single instance mode" which will make sure only one copy of zapui runs, and the `main` process manages with dispatching of the renderer processes.
   d.) Renderer processes communicate with `zap` via HTTP.
   e.) Main process communicates with `zap` via RPC.

turon:
:+1: The plan to split out the UI features from the pure-command-line `zap` core sounds good! :+1:

tecimovic:
Here is the plan as I see it right now:

1.) Current state:
[...]

PR #460, opened 2022-03-29
Silabs

tecimovic:
- rework the entry points to split up electron and node a bit:


tecimovic:
First part of addressing: https://github.com/project-chip/zap/issues/153

tecimovic:
Second part of addressing: https://github.com/project-chip/zap/issues/153

PR #462, opened 2022-03-31
Update format for device type lists to align with the Matter SDK

mrjerryjohns:
Fixes the endpoint config helpers to emit the device type lists in the right format, even though it contains just a single device in the list for now for each endpoint.

### DIFF
--- a/src-electron/generator/helper-endpointconfig.js
+++ b/src-electron/generator/helper-endpointconfig.js
@@ -307,7 +307,8 @@ function endpoint_fixed_device_type_array(options) {
 
       ret +=
         '{' +
-        this.deviceList[i].deviceId.toString() +
+        '0x' +
+        bin.int16ToHex(this.deviceList[i].deviceId) +
         ',' +
         this.deviceList[i].deviceVersion.toString() +
         '}'


src-electron/generator/helper-endpointconfig.js
```diff
@@ -295,20 +295,39 @@ function endpoint_attribute_list(options) {
...
+        if (device != null && device.deviceId != null) {
+          return (
+            '{ ' +
+            device.deviceId.toString() +
+            ',' +
+            device.deviceVersion.toString() +
+            '}'
+          )
+        } else {
+          return '{}'
```
bzbarsky-apple, line 313: What will this init things to in practice?

src-electron/generator/helper-endpointconfig.js
```diff
@@ -295,20 +295,39 @@ function endpoint_attribute_list(options) {
...
-}
-
-function endpoint_fixed_device_version_array(options) {
+function endpoint_fixed_device_type_array(options) {
   return (
     '{ ' +
-    this.deviceList.map((device) => device.deviceVersion).join(', ') +
+    this.deviceList
+      .map(function (device) {
+        console.log(device)
```
bzbarsky-apple, line 303: Please take out the debugging output?

src-electron/generator/helper-endpointconfig.js
mrjerryjohns: An `EmberAfDefinedEndpoint` that has a `EmberAfDeviceType` instance that points to a null list.

mrjerryjohns: This initializes to just a default constructed `EmberAfDeviceType`, which is actually not what we want. What we do want is to make it such that the appropriate length for that endpoint is set to 0.

src-electron/generator/helper-endpointconfig.js
```diff
@@ -295,18 +295,67 @@ function endpoint_attribute_list(options) {
...
+    if (this.deviceList[i] != null && this.deviceList[i].deviceId != null) {
+      if (wroteItem) {
+        ret += ','
+      }
+
+      ret +=
+        '{' +
+        this.deviceList[i].deviceId.toString() +
+        ',' +
+        this.deviceList[i].deviceVersion.toString() +
```
bzbarsky-apple, line 312: These should probably use `bin.int16ToHex`?

bzbarsky-apple (approved):

tecimovic:
I'm ok with merging this whenever you wish. I restarted the Cypress tests too. Sometimes they break for no discernible reason.

PR #463, opened 2022-03-31
Handle null default values in endpoint config codegen.

bzbarsky-apple:
Without this change, null generally gets treated as 0 for integers and
throws an exception for strings.


tecimovic:
All good. Only Matter uses the endpoint config anyway, so this can't affect any other SDKs currently.

PR #466, opened 2022-04-04
Silabs

tecimovic:
This PR has no functional changes around generation. It's just build/release infrastructure work.

- add build targets for platform-specific non-ui executables
- fix Jenkinsfile with flags
- add feature level

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,7 @@
 final boolean buildForMac = false
 final boolean runCypressTests = false
 final boolean triggerAdapterPackJob = false
+final boolean sonarScan = false
 
 pipeline
 {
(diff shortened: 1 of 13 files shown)


tecimovic:
This PR also fixes the issue caused by endpoint config helper, which tripped the token generation on zigbee stack. Turns out zigbee stack DOES USE the endpoint_config helper in some side-ways way, so my earlier "guarantees" that only Matter uses that were naive.

Anyway: this is not fixed with an extra option to the endpoint_config helper which Matter doesn't use but Zigbee pro does.

tecimovic:
Fix for https://github.com/project-chip/zap/issues/467

PR #480, opened 2022-04-28
Add responseName field to clusters in endpoint config

erjiaqing:
Per spec (Spec 7.17.3.1, 7.17.3.2), the GeneratedCommandList are the commands that in response to another command in AcceptedCommandList

This PR will include a "responseName" in the cluster info in endpoint config, to generate the above attributes correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "type": "commonjs",
   "name": "zap",
-  "version": "2022.4.28",
+  "version": "2022.5.5",
   "description": "Configuration tool for the Zigbee Cluster Library",
   "productName": "zap",
   "cordovaId": "",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "type": "commonjs",
   "name": "zap",
-  "version": "2022.4.29",
+  "version": "2022.4.28",
   "description": "Configuration tool for the Zigbee Cluster Library",
   "productName": "zap",
   "cordovaId": "",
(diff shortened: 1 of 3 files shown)


bzbarsky-apple (approved):

tecimovic (commented):
Minor feedback: the RESPONSE_REF is intended to be a "reliable" reference to the response. RESPONSE_NAME can technically contain gibberish that came from XML, however it was gibberish, so the linking code during the RESPONSE_REF linking would keep that as null.

Do you wish in this case to properly get nothing (for which you would need to not use RESPONSE_NAME, but a RESPONSE_REF link), or you are ok to retrieve gibberish that was typed in the XML and didn't correspond to any real command (which is the case where RESPONSE_NAME would contain gibberish, but RESPONSE_REF would stay null)?

erjiaqing:
> Minor feedback: the RESPONSE_REF is intended to be a "reliable" reference to the response. RESPONSE_NAME can technically contain gibberish that came from XML, however it was gibberish, so the linking code during the RESPONSE_REF linking would keep that as null.
> 
> Do you wish in this case to properly get nothing (for which you would need to not use RESPONSE_NAME, but a RESPONSE_REF link), or you are ok to retrieve gibberish that was typed in the XML and didn't correspond to any real command (which is the case where RESPONSE_NAME would contain gibberish, but RESPONSE_REF would stay null)?

Thanks for the advice, after checking the SQL and the usage, what I really want here is the command id ofr the response 
 for code gen (https://github.com/project-chip/connectedhomeip/pull/17833)

I updated the code to use the response_ref to lookup for the actual command code, and use the response name as comment in the code. (since the response ref itself is meanful only in zap's scope)

erjiaqing:
Merge this since it is verified by https://github.com/project-chip/connectedhomeip/pull/17833

Issue #223, opened 2021-09-17
Migrate SETTINGS table when recreating schema


tecimovic:
When schema changes, zap will wipe out the DB and reload it from scratch.

This looses the SETTINGS table, which are local per-install settings.

Add following to the process of changing the scheme:
  - before wiping out old schema, do a full SELECT on the SETTINGS table.
  - perform the wipe-out as now.
  - load back the data from earlier SELECT query into the new SETTINGS table.

tecimovic:
The purpose of this bug, is to keep the SETTINGS that were added previously, even when schema is recreated.

Note that there are 3 "special" settings:
   APP/VERSION, APP/HASH and APP/DATE.
Those are wrritten into the SETTINGS table by design, every time zap starts, to record the version of the zap that ran against this databse. These three are ok to not persist. They are intended to keep growing.

The settings that should persist are other category/key/value triplets.

For example, if the SETTING table contains:
   category=METAFILE
   key=ZCL_XML_LOCATION
   value=/home/user/some/place/zcl.xml
(made these up as example), then the expectation is, that after the schema is recreated, the SETTING table will still contain this triplet.

There are several values we wanted to put into this table, but we mostly didn't because they all get wiped out every time schema changes.

paulr34:
merged PR https://github.com/project-chip/zap/pull/486

PR #486, opened 2022-04-30
saving settings when schema changes 

paulr34:
https://github.com/project-chip/zap/issues/223

### DIFF
--- a/src-electron/db/db-api.js
+++ b/src-electron/db/db-api.js
@@ -472,7 +472,6 @@ async function performSchemaLoad(db, schemaContent) {
 async function loadSchema(db, schemaPath, zapVersion, sqliteFile = null) {
   let schemaFileContent = await fsp.readFile(schemaPath, 'utf8')
   let rows = []
-  rows = await saveSettings(db)
   let context = {
     filePath: schemaPath,
     data: schemaFileContent,
@@ -480,6 +479,7 @@ async function loadSchema(db, schemaPath, zapVersion, sqliteFile = null) {
   }
   await determineIfSchemaShouldLoad(db, context)
   if (context.mustLoad && context.hasSchema) {
+    rows = await saveSettings(db)
     await closeDatabase(db)
     if (sqliteFile != null) util.createBackupFile(sqliteFile)
   }
@@ -490,7 +490,6 @@ async function loadSchema(db, schemaPath, zapVersion, sqliteFile = null) {
       db = await initDatabase(sqliteFile)
     }
   }
-
   if (context.mustLoad) {
     await performSchemaLoad(db, context.data)
     await updateCurrentSchemaCrc(db, context)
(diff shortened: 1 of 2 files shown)


tecimovic (changes requested):
You've implemented this with just the three tokens we have now: VERSION, HASH, and DATE.
The point of this functionality is, to propagate into the table whatever happens to be there.

This table can contain other rows, depending on future functionality.

So this should be implemented in a way where you read WHATEVER is in that table, and then writing it back out. It can't be limited to just VERSION, HASH and DATE.

tecimovic:
VERSION, HASH and DATE are actually the ones that will probably be wiped out in the process, so this feature addresses all BUT those.

paulr34:
@tecimovic it saves off version, hash, and date before the wipeout then adds them back

paulr34:
At first, I started making a function which SELECTS the entire table but since the table was already being populated in such a manner that I should keep it consistent. 

On the other hand, I agree with your requested changes but need to point out that the changes would require reorganizing how the settings table is being populated in all scenarios.

As you can see I just saved off version, hash, and date then added them back in while following the same structure.

paulr34:
Thanks for explaining here https://github.com/project-chip/zap/issues/223
Working on fixing it.

src-electron/db/db-api.js
```diff
@@ -400,6 +400,17 @@ async function insertOrReplaceSetting(db, category, key, value) {
...
 }
+/**
+ * Returns an array of containing the entire SETTING table
+ *
+ * @param {*} db
+ * @returns  An array containing the entire SETTING table
+ */
+async function saveSettings(db) {
+  let rows = []
+  rows = await dbAll(db, 'SELECT * FROM SETTING')
```
tecimovic, line 411: Move this function into the query-<xxx>.js files. It can be used outside of db-api.js.

Also:
   the name of this function is "saveSettings". Which makes me assume it "saves" the settings.
However it does not "save" the settings, it "retrieves" or "selects" the settings.

Please name it matching to what it does.

src-electron/db/db-api.js
```diff
@@ -482,9 +490,18 @@ async function loadSchema(db, schemaPath, zapVersion, sqliteFile = null) {
...
     }
   }
+
   if (context.mustLoad) {
     await performSchemaLoad(db, context.data)
     await updateCurrentSchemaCrc(db, context)
+    for (let i = 0; i < rows.length; i++) {
+      await insertOrReplaceSetting(
+        db,
+        rows[i].CATEGORY,
```
tecimovic, line 499: In principle this works, but it hardcodes the columns of the setting table here. Let's just clean this up a bit:

Preferred pattern is:
   0.) add the "selectSettings()", which will read the settings from the table and create an array of JS objects that have keys: "category", "key", "value", so add the mapping to that db-mapping.js. It should not do "SELECT *", but instead it should do "SELECT CATEGORY, KEY, VALUE" and then map the rows into the JS objects, like most of our query functions do.
   1.) Add the "updateSetting()" that receives either a single { category:, key:, value: } object, or an array of them, and performs the "INSERT OR UPDATE".
   2.) Then make the db-api.js do the same thing you've made it here, except call these methods to interact with the SETTING table.
   
Otherwise, once we clean this up, this is good to go I think.

tecimovic (changes requested):

src-electron/db/db-api.js
paulr34: Okay, I am on it

src-electron/db/db-api.js
paulr34: Okay thanks! I am on it

src-electron/db/db-api.js
paulr34: For now, the new function selectSetting() is still in db-api.js as we discussed..

tecimovic (approved):
Looks good.

jingteng25742 (approved):

PR #490, opened 2022-05-05
enforce zigbee shared cluster logic

jingteng25742:


src-electron/db/query-command.js
```diff
@@ -1385,24 +1387,26 @@ async function selectAllCommandDetailsFromEnabledClusters(
       db,
       `
   SELECT
-    COMMAND.COMMAND_ID,
```
brdandu, line 1388: Could we not re-arrange these? Will be easier to look at the changes and would be easier to rebase as well.

test/resource/isc/ha-combined-interface.zap
```diff
@@ -13,10 +13,6 @@
     {
       "key": "manufacturerCodes",
       "value": "0x1049"
-    },
-    {
-      "key": "shareConfigsAcrossEndpoints",
-      "value": "1"
```
brdandu, line 19: Do you add these and are now removing these?

src/util/editable-attributes-mixin.js
```diff
@@ -105,68 +105,66 @@ export default {
...
-            clusterRef: this.selectedCluster.id,
-            attributeSide: singleItem.side,
-            reportMinInterval: singleItem.reportMinInterval,
-            reportMaxInterval: singleItem.reportMaxInterval,
-          }
-          this.$store.dispatch('zap/updateSelectedAttribute', editContext)
-        }
-      })
-    },
+    // selectedCluster() {
```
brdandu, line 108: Can we get rid of this commented code?

src-electron/db/query-command.js
```diff
@@ -1385,24 +1387,26 @@ async function selectAllCommandDetailsFromEnabledClusters(
...
-    CLUSTER.NAME AS CLUSTER_NAME,
-    ENDPOINT_TYPE_CLUSTER.ENABLED
+    ENDPOINT_TYPE_CLUSTER.ENDPOINT_TYPE_CLUSTER_ID
   FROM COMMAND
   INNER JOIN CLUSTER
   ON COMMAND.CLUSTER_REF = CLUSTER.CLUSTER_ID
   INNER JOIN ENDPOINT_TYPE_CLUSTER
   ON CLUSTER.CLUSTER_ID = ENDPOINT_TYPE_CLUSTER.CLUSTER_REF
-  WHERE ENDPOINT_TYPE_CLUSTER.CLUSTER_REF in (${endpointTypeClusterRef})
+  WHERE ENDPOINT_TYPE_CLUSTER.ENDPOINT_TYPE_CLUSTER_ID in (${endpointTypeClusterRef})
```
brdandu, line 1409: Would be very careful with changing these. Can blow ZUC command generation.

test/gen-template/zigbee/gen-templates.json
```diff
@@ -15,8 +15,8 @@
     "types": {
       "bacnet_oid": "bacnet_type_t"
     },
-    "configOptions": {
-      "shareConfigsAcrossEndpoints": false
+    "generator": {
+      "shareClusterStatesAcrossEndpoints": "true"
```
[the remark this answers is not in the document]
jingteng25742: The Cypress test is captured via check-cluster-attributes-singleton-in-zigbee-mode.spec.js as part of this PR.

src-electron/db/query-command.js
jingteng25742: I actually re-arranged them to make the entries more organized / readable. I can restore to previous state to avoid rebase issues.

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
```diff
@@ -0,0 +1,115 @@
...
+      cy.visit(data.baseurl)
+    })
+
+    cy.gotoAttributePage('Chatting Station (0x0601)', 'General')
+  })
+  it(
+    'verify initial values on Endpoint 2 are reflected from Endpoint 1',
+    { retries: { runMode: 2, openMode: 2 } },
+    () => {
+      // Set ZCL version - 1
```
brdandu, line 66: Get from Ep1 instead of set?

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
```diff
@@ -0,0 +1,115 @@
...
+      cy.get(
+        ':nth-child(8) > [style="min-width: 180px;"] > .q-field > .q-field__inner .q-field__control > .q-field__control-container > input'
+      ).should('have.value', '22')
+
+      // Set cluster revision - 33
+      cy.get(
+        ':nth-child(19) > [style="min-width: 180px;"] > .q-field > .q-field__inner .q-field__control > .q-field__control-container > input'
+      ).should('have.value', '33')
+
+      // Set ResetToFactoryDefaults command
```
brdandu, line 81: Changes to comments

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
```diff
@@ -0,0 +1,115 @@
...
+      cy.gotoCommandsTab()
+      cy.get(':nth-child(3) > .q-mt-xs').should(
+        'have.attr',
+        'aria-checked',
+        'true'
+      )
+
+      // Uncheck ResetToFactoryDefaults command
```
brdandu, line 89: Changes to comments

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
```diff
@@ -0,0 +1,115 @@
...
+      cy.get(':nth-child(3) > .q-mt-xs')
+        .click({
+          force: true,
+        })
+        .should('have.attr', 'aria-checked', 'false')
+    }
+  )
+
+  it(
+    'Check Command changes on Endpoint 2 is reflected on Endpoint 1',
```
brdandu, line 99: How are you doing this? Not very familiar with cypress tests so looking at the code below I am confused.

cypress/support/commands.js
```diff
@@ -11,5 +11,19 @@ Cypress.Commands.add('gotoAttributePage', (endpoint, cluster) => {
   if (endpoint) cy.addEndpoint(endpoint)
   cy.get('.q-page-container > div').children().should('contain', cluster)
   cy.get('div').contains(cluster).click()
-  cy.get(`#${cluster} > .q-expansion-item__container > .q-expansion-item__content > :nth-child(1) > .q-table__container > .q-table__middle > .q-table > tbody > .text-weight-bolder > :nth-child(7) > .q-btn > .q-btn__wrapper > .q-btn__content > .notranslate`).click({force: true})
+  cy.get(
+    `#${cluster} > .q-expansion-item__container > .q-expansion-item__content > :nth-child(1) > .q-table__container > .q-table__middle > .q-table > tbody > .text-weight-bolder > :nth-child(7) > .q-btn > .q-btn__wrapper > .q-btn__content > .notranslate`
+  ).click({ force: true })
+})
+
+Cypress.Commands.add('gotoAttributesTab', () => {
```
brdandu, line 19: Is this taking you to the attributes tab of ep1?

src-electron/db/query-command.js
```diff
@@ -1396,13 +1398,15 @@ async function selectAllCommandDetailsFromEnabledClusters(
     COMMAND.MUST_USE_TIMED_INVOKE,
     ENDPOINT_TYPE_CLUSTER.SIDE,
     CLUSTER.NAME AS CLUSTER_NAME,
-    ENDPOINT_TYPE_CLUSTER.ENABLED
+    CLUSTER.CLUSTER_ID,
+    ENDPOINT_TYPE_CLUSTER.ENABLED,
```
brdandu, line 1402: Not a big issue but this can be removed from the diff as well

src-electron/db/query-command.js
brdandu: Why do we need this change? Would this not work with `ENDPOINT_TYPE_CLUSTER.CLUSTER_REF`. I will have to look at the schema but am just worried about this causing any regen diff issues in ZUC. Can we avoid changing this?

src-electron/rest/user-data.js
```diff
@@ -145,7 +147,6 @@ function httpPostAttributeUpdate(db) {
   return async (request, response) => {
     let {
       action,
-      endpointTypeId,
```
brdandu, line 148: Any reason we had both id and list before?

src/util/editable-attributes-mixin.js
brdandu: Any reason we are keeping this?

jingteng25742: Well most of this code should exist but it's buggy atm. I'm just leaving them here for the contractors to review / fix.

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
jingteng25742: The chained .type() API call actually does the set.

src-electron/db/query-command.js
jingteng25742: CLUSTER.CLUSTER_ID is needed here as part of the return result.

src-electron/db/query-command.js
jingteng25742: `ENDPOINT_TYPE_CLUSTER.CLUSTER_REF` is the Cluster id while an endpoint specific cluster id needs to be used to compare with `endpointTypeClusterRef`

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
brdandu: Is this a set either. Same for all the calls with no .type() chained to them?

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
brdandu: Are you unchecking here?

cypress/integration/attributes/check-cluster-attributes-singleton-in-zigbee-mode.spec.js
brdandu: So the `cy.gotoAttributePage` and `cy.gotoCommandsTab()` take you to Ep1? Could we name them in a way that mentions EP1 as well for less confusion.

cypress/support/commands.js
brdandu: Same as above.

src-electron/db/query-command.js
brdandu: I meant the diff on `ENDPOINT_TYPE_CLUSTER.ENABLED,`. If you move it above `CLUSTER.CLUSTER_ID,` then the diff is cleaner but I see that we have an additional , . This is fine. I missed the ,

brdandu (approved):

Issue #497, opened 2022-05-09
selectCommandDetailsFromAllEndpointTypesAndClusters should sort its output


bzbarsky-apple:
Right now it does not.

Issue #496, opened 2022-05-09
selectEndpointClusterAttributes should be using dbApi.fromDbBool


bzbarsky-apple:
It's dumping out raw column values instead.

Issue #494, opened 2022-05-09
selectAllClustersDetailsFromEndpointTypes should sort its output


bzbarsky-apple:
Right now it does not.

Issue #472, opened 2022-04-21
Add ability to access cluster lists from device types


tecimovic:
Enhance zcl_device_types to allow:
- iterating over all clusters inside the device type
- determine if clusters are mandatory or not

PR #498, opened 2022-05-10
Silabs

tecimovic:
Besides some more minor issues, this mostly fixes:

1.) https://github.com/project-chip/zap/issues/472
2.) There was a loading bug: when the DEVICE_TYPE_ATTRIBUTE and DEVICE_TYPE_COMMAND tables were being relinked back to the ATTRIBUTE and COMMAND tables as a post-load action, the queries were structured such, that the COMMAND_REF and ATTRIBUTE_REF were left empty in the device type tables. This is fixed now.

Added:
3.) https://github.com/project-chip/zap/issues/494
4.) https://github.com/project-chip/zap/issues/497
5.) https://github.com/project-chip/zap/issues/496


src-electron/db/query-command.js
```diff
@@ -1504,13 +1504,15 @@ async function selectCommandDetailsFromAllEndpointTypesAndClusters(
...
-    AND ETC.ENDPOINT_TYPE_CLUSTER_REF in (${endpointClusterIds}) `
+    AND ETC.ENDPOINT_TYPE_CLUSTER_REF in (${endpointClusterIds})`
 
   if (doGroupBy) {
     // See: https://github.com/project-chip/zap/issues/192
-    query = query + ` GROUP BY C.NAME, C.COMMAND_ID`
+    query = query + ' GROUP BY C.NAME, C.COMMAND_ID'
   }
 
+  query = query + ' ORDER BY C.MANUFACTURER_CODE, C.CODE, C.NAME'
```
bzbarsky-apple, line 1514: To keep the sort closer to what we were accidentally/usually getting out of the group by, would it make sense to sort on NAME first?

I worry about the major changes to codegen that will likely happen otherwise when this is imported into Matter, which will make it very hard to review whether the changes are intended...

If we _do_ want to change the sort order significantly, I'd like it if it happened in a separate PR so there is a separate changeset we do the Matter update to so we can easily check that all the codegen changes are just from the ordering change.

src-electron/db/query-endpoint-type.js
```diff
@@ -200,9 +201,11 @@ ON
 WHERE
   ENDPOINT_TYPE_CLUSTER.ENDPOINT_TYPE_REF IN (${endpointTypeIds})
 AND
-  ENDPOINT_TYPE_CLUSTER.SIDE IS NOT "" AND ENDPOINT_TYPE_CLUSTER.ENABLED=1
+  ENDPOINT_TYPE_CLUSTER.SIDE IS NOT "" AND ENDPOINT_TYPE_CLUSTER.ENABLED = 1
 GROUP BY
-  NAME, SIDE`
+  NAME, SIDE
+ORDER BY 
+  CLUSTER.MANUFACTURER_CODE, CLUSTER.CODE, CLUSTER.DEFINE`
```
bzbarsky-apple, line 208: Again, it would be very helpful on the Matter side if this ordering change were a separate changeset from everything else.

bzbarsky-apple (changes requested):

src-electron/db/query-endpoint-type.js
tecimovic: Ok. I'll split this up. Hold on.

tecimovic:
@bzbarsky-apple : the final commit in this PR disables the "ORDER BY" changes. I will open a separate PR, that will ONLY turn these back on and leave it for you to review.

tecimovic:
@bzbarsky-apple : the final commit in this PR disables the "ORDER BY" changes. I will open a separate PR, that will ONLY turn these back on and leave it for you to review.

tecimovic:
@bzbarsky-apple : the final commit in this PR disables the "ORDER BY" changes. I will open a separate PR, that will ONLY turn these back on and leave it for you to review.

PR #501, opened 2022-05-12
Modify helper endpoint config 4byte default

tehampson:
Allow for size of default value size to be configurable

defaultValue in the EmberAfDefaultOrMinMaxAttributeValue is uint16_t. For targets that have 32-bit addresses, there is wasted space in the union. We can let defaultValue be uint32_t and save on rodata region, by fitting the default value in the meta data structure itself.

### DIFF
--- a/src-electron/generator/helper-endpointconfig.js
+++ b/src-electron/generator/helper-endpointconfig.js
@@ -617,7 +617,7 @@ async function collectAttributes(endpointTypes, options) {
         // nonempty default value to use long defaults.
         // external strings and zero-length default values.
         if (
-          defaultSize > 2 ||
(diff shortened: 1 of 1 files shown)


src-electron/generator/helper-endpointconfig.js
```diff
@@ -617,7 +618,7 @@ async function collectAttributes(endpointTypes, options) {
         // nonempty default value to use long defaults.
         // external strings and zero-length default values.
         if (
-          defaultSize > 2 ||
+          defaultSize > spaceForDefaultValue ||
```
bzbarsky-apple, line 621: @tehampson The documentation in the code above no longer makes sense with this change....

PR #520, opened 2022-05-20
Enable order-by in two queries.

tecimovic:
Enable the ORDER-BY in two queries that should order.


tecimovic:
@bzbarsky-apple : this is the separate PR to enable those 2 "ORDER BY" issues from a week or two ago.

bzbarsky-apple (approved):
Might be nice to have these two changes as separate PRs, but this is probably OK...

Issue #532, opened 2022-06-02
Preview tab is open when we click back button in preferences page


Mehradml:
When we go to the preferences page and click back to the main page, the preview tab is open.
<img width="1497" alt="image" src="https://user-images.githubusercontent.com/31411797/171604559-a1e0a096-8733-4963-a41d-cadf3d172054.png">

Issue #527, opened 2022-05-24
ZAP Endpoint display issue


jingteng25742:
![image](https://user-images.githubusercontent.com/61332088/170096927-466e51df-f9fc-4bf7-8df8-b69b533fde99.png)

1) Within above pictures, the values of "Enabled Clusters" / etc should be left aligned, similar to Version.
2) Each Endpoint entry (Endpoint - 1, Endpoint - 2) should be able to be expanded individually. Previously, the user is able to see all endpoint attribute fields. This behavior needs to return. Behavior similar to https://quasar.dev/vue-components/expansion-item

wehale:
I'm not able to validate that this has been resolved. To me it appears that the values for Enabled Clusters, Enabled Attributes, and Enabled reporting are all right aligned still.

PR #533, opened 2022-06-02
Resolving UI issues 

Amirreza-Ghasemkhani:
These are the resolved issues: 
https://github.com/project-chip/zap/issues/532 
https://github.com/project-chip/zap/issues/527

### DIFF
--- a/src/components/ZclEndpointCard.vue
+++ b/src/components/ZclEndpointCard.vue
@@ -20,8 +20,8 @@ limitations under the License.
       :bordered="isSelectedEndpoint"
       @click="setSelectedEndpointType(endpointReference)"
     >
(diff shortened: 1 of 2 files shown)


jingteng25742 (changes requested):
1) Upon creation of an Endpoint, the details should be shown by default and should not require clicking to "expand".
To indicate "expand" or not expanded mode to user, an state icon should be shown. behavior should replicate  - https://quasar.dev/vue-components/expansion-item

2) If two endpoints are created with details showing for each, the displayed info is not correct. Both details window seem to be displaying from a single source

Mehradml:
@jingteng25742 I created a new issue for the problem you mentioned.
https://github.com/project-chip/zap/issues/553
I think it would be great if we could merge this PR for now and fix the other one in another PR.

PR #534, opened 2022-06-02
Ubuntu version

Mehradml:
Test pull request to see if ubuntu version 18 works fine with our github actions workflows.


tecimovic:
@Mehradml : you have changed all the ubuntu to ubuntu 18. So now it's no longer testing for ubuntu-latest.

Would it be possible to have at least few checks run against BOTH!

I was hoping we could run cypress and the main unit test run on BOTH ubuntu-latest and ubuntu-18, not just ubuntu-18.

tecimovic (changes requested):
I'd like this to run on BOTH ubuntu-latest and ubuntu-18.04, not just one of them.

PR #560, opened 2022-06-16
Adding helpers for zcl cli commands such that the right content is generated for Zigbee and other cleanup.

brdandu:
- Increasing the feature level pointer for apacka.json
- Checking for lowercase when querying bitmaps and enums since the atomics are in uppper case under certain occasions for eg BITMAP8 instead of bitmap8
- Getting rid of the cli helpers in helper-zcl.js and using the new data_type table after the schema change instead of the old logic. Also moving the cli helpers int helper-zigbee-zcl since it is specific to zigbee 3.0 only
- Adding helper-zigbee-zcl.js to template-engine such that the helpers are included
- Adding the new zcl cli helper in zap-cli.zapt such that testing of the new helper is covered along with the old one.
- JIRA: ZAPP-802


bzbarsky-apple (approved):

src-electron/db/query-bitmap.js
```diff
@@ -87,8 +87,8 @@ SELECT
   BITMAP.SIZE AS SIZE
 FROM BITMAP
 INNER JOIN DATA_TYPE ON BITMAP.BITMAP_ID = DATA_TYPE.DATA_TYPE_ID
-WHERE DATA_TYPE.NAME = ? AND DATA_TYPE.PACKAGE_REF = ?`,
-      [name, packageId]
+WHERE (DATA_TYPE.NAME = ? OR DATA_TYPE.NAME = ?) AND DATA_TYPE.PACKAGE_REF = ?`,
+      [name, name.toLowerCase(), packageId]
```
jingteng25742, line 91: Are both the input arg 'name' and the column attribute DATA_TYPE.NAME fields case-insensitive?
If 'name' is only lower cased but nam DATA_TYPE.NAME has some weird capitalization, they might not necessarily match.
Using `COLLATE NOCASE` in the SQL query would avoid that issue.

src-electron/db/query-enum.js
```diff
@@ -244,9 +244,9 @@ INNER JOIN
 ON
   ENUM.ENUM_ID = DATA_TYPE.DATA_TYPE_ID
 WHERE
-  NAME = ? AND PACKAGE_REF = ?
+  (DATA_TYPE.NAME = ? OR DATA_TYPE.NAME = ?)AND PACKAGE_REF = ?
 ORDER BY NAME`,
-      [name, packageId]
+      [name, name.toLowerCase(), packageId]
```
jingteng25742, line 249: same as earlier comment about case-insensitiveness.

src-electron/generator/helper-zcl.js
```diff
@@ -2614,8 +2466,6 @@ exports.if_manufacturing_specific_cluster = dep(
   if_manufacturing_specific_cluster,
   { to: 'if_mfg_specific_cluster' }
 )
-exports.zcl_command_argument_type_to_cli_data_type =
-  zcl_command_argument_type_to_cli_data_type
```
jingteng25742, line 2618: do we need to maintain backward compatibility for the zcl_command_argument_type_to_cli_data_type api under helper-zcl.js?

src-electron/db/query-bitmap.js
brdandu: So I thought about this. I used like to overcome this at one point but didn't want to promote this so just kept it to either small/uppercase. None of the other variations.

src-electron/generator/helper-zcl.js
brdandu: This helper is still present and generates the correct content in helper-zigbee-zcl.js

jingteng25742 (approved):

Issue #570, opened 2022-06-25
Apply changes on packages database into front-end


Mehradml:
There are some changes made on packages table and we have make changes in the front-end accordingly.
Mostly on package preferences page.

brdandu:
This seems to have been closed with https://github.com/project-chip/zap/pull/581

Issue #557, opened 2022-06-14
UI display issues


jingteng25742:
1)
<img width="936" alt="Screen Shot 2022-06-14 at 10 46 19 AM" src="https://user-images.githubusercontent.com/61332088/173607017-2a901edb-2560-408e-8624-31f521ae149d.png">
 
The circled box (container for Zigbee Cluster)  should expand all the way horizontally similar to the ZCL tab

2) 
<img width="896" alt="Screen Shot 2022-06-14 at 10 49 53 AM" src="https://user-images.githubusercontent.com/61332088/173607985-7a323376-31aa-4a22-a244-070b94f6c470.png">

The "Zigbee Cluster Configuration" display has higher priority than the "ZCL GLOBAL" / "ZCL Extension" Tab. When the screen is small horizontally, the "Zigbee Cluster Configuration" display should stay. The text of "ZCL GLOBAL" / "ZCL Extension" should disappear and only display the corresponding logos.

3)
<img width="835" alt="Screen Shot 2022-06-14 at 10 54 04 AM" src="https://user-images.githubusercontent.com/61332088/173608853-bc381fdc-0433-459c-869f-393a34c3ca00.png">
With the cursor over the main content page, the user should be able to scroll up and down. Currently, the display cannot move.

Issue #556, opened 2022-06-14
Show to user whether Zigbee or Matter is used.


tecimovic:
![image](https://user-images.githubusercontent.com/7923190/173588357-8f555a68-0ff8-4b85-a1e7-96186babacf4.png)

Show which specific toplevel zcl-properties is used for this configuratio.

You can show a string:
   "Cluster configurator: Zigbee"
   or
     "Cluster configurator: Matter"
 
 or something like that.

The string to use, is in the database, in the VERSION field of the PACKAGE table. So depending on which toplevel package is chosen by the users, the UI shows what is being used.

jingteng25742:
For the text, perhaps we can standardize / follow the format as to the rest of the bar ( .e.g. ZCL GLOBAL OPTOINS / ZCL Extensions)
.e.g. "#logo [Zigbee | Matter] Cluster Configurator"
#logo being either the Zigbee or Matter official logo.

Issue #553, opened 2022-06-14
Zap endpoint information


Mehradml:
In the endpoint section, the shown summary of all endpoints is linked into one and when you change the selected endpoint, the summary information of all endpoints changes.

brdandu:
This is linked to the removal of endpoint summary from all endpoint panes. Closing this as a duplicate of https://github.com/project-chip/zap/issues/725 now.

PR #581, opened 2022-07-02
Resolving UI Issues

Amirreza-Ghasemkhani:
https://github.com/project-chip/zap/issues/556
https://github.com/project-chip/zap/issues/570 
https://github.com/project-chip/zap/issues/557
https://github.com/project-chip/zap/issues/553
https://github.com/project-chip/zap/issues/532
https://github.com/project-chip/zap/issues/527

- Update check_summary tests based on the new UI

### DIFF
--- a/cypress/integration/check_summary/enabled-attributes.spec.js
+++ b/cypress/integration/check_summary/enabled-attributes.spec.js
@@ -29,12 +29,31 @@ describe('Testing enabled attributes amount', () => {
     'checks if number is updated',
     { retries: { runMode: 2, openMode: 2 } },
     () => {
-      cy.fixture('data').then((data) => {
-        cy.get(':nth-child(7) > .text-right').then(($div2) => {
-          const num2 = parseFloat($div2.text())
-          expect(num2).to.eq(Number(data.availableAttributes1))
-        })
+      cy.fixture('baseurl').then((data) => {
+        cy.visit(data.baseurl)
       })
+      cy.addEndpoint('Billing Unit (0x0203)', 'General')
+      cy.get('[data-test="endpoint-enabled-attributes-amount"]').then(
+        ($div) => {
+          const num1 = parseFloat($div.text())
+          cy.gotoAttributePage('', 'General')
+          cy.get(
+            '.table_body:eq(2) > :nth-child(2) > .q-mt-xs > .q-toggle__inner'
+          ).click()
+          cy.get('.router-link-active')
+            .contains('Back')
+            .click()
+            .then(() => {})
+
+          cy.get('[data-test="endpoint-body-toggler-show"]').click()
+          cy.get('[data-test="endpoint-enabled-attributes-amount"]').then(
+            ($div2) => {
+              const num2 = parseFloat($div2.text())
+              expect(num2).to.eq(17)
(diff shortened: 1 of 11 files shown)


brdandu:
test/ui.test.js is failing. Can we fix that before I start looking into this?

brdandu:
I am seeing the following in the UI which does not seem right:
Issue 1

brdandu:
How can I verify issue 570?

brdandu:
Still see this separation in the UI from 557
Issue 2

brdandu:
Issue 553 seems seems to have another issue now where the count is not correct
Issue 3
 Make sure to add a test for this once it is fixed

brdandu:
Issue 532 still seems to be the case. I go to cluster config or preferences and come back but see the preview pane still showing up.

brdandu (changes requested):
Take a look at all the comments

Amirreza-Ghasemkhani:
I will take a look at the issue that you mentioned.

Amirreza-Ghasemkhani:
> I am seeing the following in the UI which does not seem right: Issue 1
> 
> Just do Cluster configurator: Zigbee/Matter like @tecimovic suggested
> 
> Also I do not see a logo which was suggested by @jingteng25742

logo or Icon? because I added icon for smaller viewpoints if you resize the page you will see it will convert to just a icon ...
image

Amirreza-Ghasemkhani:
> Still see this separation in the UI from 557 Issue 2

which one do you perfer? 
1. 
image
2.
image
3.
image

brdandu:
> > Still see this separation in the UI from 557 Issue 2
> 
> which one do you perfer? 1. image 2. image 3. image

I am fine with 3 but why do we have a shadow from the top on the right side but not the left? not a big deal but if you can remove that to make it look the same on both sides then great. If not lets just go with 3.

Amirreza-Ghasemkhani:
> > > Still see this separation in the UI from 557 Issue 2
> > 
> > 
> > which one do you perfer? 1. image 2. image 3. image
> 
> I am fine with 3 but why do we have a shadow from the top on the right side but not the left? not a big deal but if you can remove that to make it look the same on both sides then great. If not lets just go with 3.

@brdanduIt’s because I took the screenshot when inspection elements were open, it’s okay without of any extra shadows if you don't open the inspect part of the site
.

brdandu:
> I am seeing the following in the UI which does not seem right: Issue 1
> 
> Just do Cluster configurator: Zigbee/Matter like @tecimovic suggested
> 
> Also I do not see a logo which was suggested by @jingteng25742

2 things here:
1) Get rid of the display button text
2) The "Cluster Configurator" text is getting cached which we do not want. When I switch between SDKs this does not get updated automatically

brdandu:
I switched between "npm run zap" and launching zap using the Zigbee stack and I still see the following:
Screen Shot 2022-08-10 at 9 27 07 AM

@Amirreza-Ghasemkhani as per our conversation, this should be updated based on the stack we are using. This way it just shows whatever got picked up first

brdandu (approved):
Feel free to merge and then fix the Cluster Configurator Issue

PR #589, opened 2022-07-08
Add the end-of-run cleanup action to set package.json back to 0.0.0

tecimovic:
This solves a problem, that after every zap action (generate, analyze, run UI, etc), the package.json ends up in a dirty state, because scripts insert a valid version instead of a 0.0.0.

After this change, the version is cleaned back to the committed state: 0.0.0, so anyone relying on cleanliness of submodule will be ok.

We can revisit this if we ever revisit zap versioning scheme.

### DIFF
--- a/src-script/zap-versionstamp.js
+++ b/src-script/zap-versionstamp.js
@@ -18,14 +18,9 @@
 
 const scriptUtil = require('./script-util.js')
 
-let startTime = process.hrtime.bigint()
-
 //workaround: executeCmd()/spawn() fails silently without complaining about missing path to electron
 process.env.PATH = process.env.PATH + ':/usr/local/bin/'
 
-scriptUtil
-  .stampVersion()
-  .then(() => scriptUtil.doneStamp(startTime))
-  .catch((err) => {
-    console.log(err)
-  })
+scriptUtil.stampVersion().catch((err) => {
+  console.log(err)
+})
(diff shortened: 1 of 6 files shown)


yunhanw-google (approved):

src-script/zap-versionstamp.js
```diff
@@ -18,14 +18,9 @@
...
 //workaround: executeCmd()/spawn() fails silently without complaining about missing path to electron
 process.env.PATH = process.env.PATH + ':/usr/local/bin/'
 
-scriptUtil
-  .stampVersion()
-  .then(() => scriptUtil.doneStamp(startTime))
-  .catch((err) => {
-    console.log(err)
-  })
+scriptUtil.stampVersion().catch((err) => {
```
brdandu, line 24: Do we not need this time stamp we had before?

brdandu (approved):

yunhanw-google (approved):

src-script/zap-versionstamp.js
tecimovic: We need the "stampVersion", but we don't need the "doneStamp".
DoneStamp simply printed out the "All done...." text earlier.
I've now change doneStamp() to also revert the changes to package.json.

In this case, that was unnecessary, so I removed the whole thing, including the "All done..." printout, which seemed to me, honestly, a bit stupid to print out just in case when you are creating the .version file.

PR #590, opened 2022-07-08
Handling Endianness correctly

brdandu:
Handling the negative default values and float values for attributes correctly in terms of endianness(little/big)

### DIFF
--- a/src-electron/generator/helper-zcl.js
+++ b/src-electron/generator/helper-zcl.js
@@ -2210,6 +2210,78 @@ async function if_mfg_specific_cluster(clusterId, options) {
 }
 
 /**
+ *
+ * @param {*} value
+ * @param {*} size
+ * @returns The negative hex value for a given positive hex value padded with
+ * the given size.
+ */
+async function convert_float_to_hex(value, size) {
+  let res = ''
+  const arrayBuffer = new ArrayBuffer(size)
+  const dataView = new DataView(arrayBuffer)
+  dataView.setFloat32(0, value, false)
+  for (let i = 0; i < size; i++) {
+    let t = dataView.getUint8(i).toString(16)
+    res += t == 0 ? '00' : t
+  }
+  return '0x' + res
+}
+
+/**
+ *
+ * @param {*} value
+ * @param {*} size
+ * @returns the negative hex value for a given positive hex value padded with
+ * the given size.
+ */
+async function get_negative_hex_value(value, size) {
+  let res = ''
+  // Performing negation using the binary form
+  let tempBinary = ''
+  let byteArray = value.replace('0x', '').match(/.{1,2}/g)
+  for (let b of byteArray) {
+    let binValue = parseInt(b, 16).toString(2)
+    tempBinary += '0'.repeat(8 - binValue.length) + binValue
+  }
+  let paddingSize = 8 * size - byteArray.length * 8
+  tempBinary = '0'.repeat(paddingSize) + tempBinary
+  tempBinary = tempBinary.replaceAll('1', 't')
+  tempBinary = tempBinary.replaceAll('0', '1')
+  tempBinary = tempBinary.replaceAll('t', '0')
+
+  // Adding 1 to the binary number
+  if (tempBinary.charAt(tempBinary.length - 1) == '0') {
+    tempBinary = tempBinary.replace(/0$/, '1')
+  } else {
+    let carry = '1'
+    let result = ''
+    for (let i = tempBinary.length - 1; i >= 0; i--) {
+      if (tempBinary.charAt(i) == '1') {
+        if (carry == '1') {
+          result = '0' + result
+        } else {
+          result = tempBinary.charAt(i) + result
+        }
+      } else {
+        if (carry == '1') {
+          result = '1' + result
+          carry = '0'
+        } else {
+          result = tempBinary.charAt(i) + result
+        }
+      }
+    }
+    tempBinary = result
+  }
+  for (let i = 0; i < tempBinary.length / 8; i++) {
+    let t = tempBinary.substring(i * 8, (i + 1) * 8)
+    res += parseInt(t, 2).toString(16)
+  }
+  return '0x' + res
+}
+
+/**
  * Given the value and size of an attribute along with endian as an option.
  * This helper returns the attribute value as big/little endian.
  * Example: {{as_generated_default_macro 0x00003840 4 endian="big"}}
@@ -2221,11 +2293,21 @@ async function if_mfg_specific_cluster(clusterId, options) {
  */
 async function as_generated_default_macro(value, attributeSize, options) {
   let default_macro_signature = ''
+  let temp = ''
   if (attributeSize > 2) {
-    let default_macro = helperC
-      .asHex(value, null, null)
-      .replace('0x', '')
-      .match(/.{1,2}/g)
+    // Positive or Negative Value
+    let tempValue = value > 0 ? value : Math.abs(value)
+    // Float value
+    if (!isNaN(value) && value.toString().indexOf('.') != -1) {
+      temp = await convert_float_to_hex(tempValue, attributeSize)
+    } else {
+      temp = helperC.asHex(tempValue, null, null)
+    }
+    // Negative value
+    if (value < 0) {
+      temp = await get_negative_hex_value(temp, attributeSize)
+    }
+    let default_macro = temp.replace('0x', '').match(/.{1,2}/g)
     let padding_length = attributeSize - default_macro.length
     for (let i = 0; i < padding_length; i++) {
       default_macro_signature += '0x00, '
@@ -2233,14 +2315,15 @@ async function as_generated_default_macro(value, attributeSize, options) {
     for (let m of default_macro) {
       default_macro_signature += ' 0x' + m + ','
     }
-    // Applying endianess to attributes with size less than equal to 8 bytes.
-    // Thus only swapping int64u or smaller
-    if (options.hash.endian != 'big' && attributeSize <= 8) {
-      default_macro_signature = default_macro_signature
-        .split(' ')
-        .reverse()
-        .join(' ')
-    }
+  }
+
+  // Applying endianess to attributes with size less than equal to 8 bytes.
+  // Thus only swapping int64u or smaller
+  if (options.hash.endian != 'big' && attributeSize <= 8) {
+    default_macro_signature = default_macro_signature
+      .split(' ')
+      .reverse()
+      .join(' ')
   }
   return default_macro_signature
 }
--- a/test/helpers.test.js
+++ b/test/helpers.test.js
@@ -245,6 +245,32 @@ test(
 )
 
 test(
+  'Generated Macro Negative Value',
+  () => {
+    let options = { hash: { endian: 'big' } }
+    return zclHelper
+      .as_generated_default_macro('-5', 8, options)
+      .then((res) =>
+        expect(res).toBe(' 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xfb,')
+      )
+  },
+  testUtil.timeout.short()
+)
+
+test(
+  'Generated Macro Float Value',
+  () => {
+    let options = { hash: { endian: 'little' } }
+    return zclHelper
+      .as_generated_default_macro('17.0', 8, options)
+      .then((res) =>
+        expect(res).toBe('0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x88, 0x41, ')
+      )
+  },
+  testUtil.timeout.short()
+)
+
+test(
   'Attribute Mask',
   () =>
     zclHelper


tecimovic:
Some unit test seems to be failing.

brdandu:
> Some unit test seems to be failing.
https://github.com/project-chip/zap/issues/593 Logged this for a UI warning when user enters an invalid hex value as an attribute default which was the reason for the test failure. Will be a better experience for a user to see this in the UI immediately than during compile time.

src-electron/generator/helper-zcl.js
```diff
@@ -2209,6 +2209,78 @@ async function if_mfg_specific_cluster(clusterId, options) {
...
+}
+
+/**
+ *
+ * @param {*} value
+ * @param {*} size
+ * @returns the negative hex value for a given positive hex value padded with
+ * the given size.
+ */
+async function get_negative_hex_value(value, size) {
```
tecimovic: I thought we generally don't use `get_` for our helpers. We mostly do `as_` or something?

src-electron/generator/helper-zcl.js
```diff
@@ -2209,6 +2209,78 @@ async function if_mfg_specific_cluster(clusterId, options) {
...
+      } else {
+        if (carry == '1') {
+          result = '1' + result
+          carry = '0'
+        } else {
+          result = tempBinary.charAt(i) + result
+        }
+      }
+    }
+    tempBinary = result
```
tecimovic: Wow. Is there seriously no package out there that does this? This seems very complicated...

src-electron/generator/helper-zcl.js
```diff
@@ -2299,13 +2227,13 @@ async function as_generated_default_macro(value, attributeSize, options) {
     let tempValue = value > 0 ? value : Math.abs(value)
     // Float value
     if (!isNaN(value) && value.toString().indexOf('.') != -1) {
-      temp = await convert_float_to_hex(tempValue, attributeSize)
+      temp = types.convertFloatToHex(tempValue, attributeSize)
```
tecimovic, line 2230: Have you purposefully lost `await` here?

tecimovic: Yeah... Having no await is correct, looking at the actual functions.

tecimovic (approved):
Minor comments, but can merge.

src-electron/generator/helper-zcl.js
brdandu: fixed it

src-electron/util/types.js
```diff
@@ -78,6 +78,78 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
...
+ * @param {*} value
+ * @param {*} size
+ * @returns The negative hex value for a given positive hex value padded with
+ * the given size.
+ */
+function convertFloatToHex(value, size) {
+  let res = ''
+  const arrayBuffer = new ArrayBuffer(size)
+  const dataView = new DataView(arrayBuffer)
+  dataView.setFloat32(0, value, false)
```
bzbarsky-apple, line 92: What if this is a 64-bit float?

Seems like this function should check for `size` being 4 or 8, throw if not one of those, then use the right accessors on the dataview.

src-electron/util/types.js
```diff
@@ -78,6 +78,78 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
   }
 }
 
+/**
+ *
+ * @param {*} value
+ * @param {*} size
+ * @returns The negative hex value for a given positive hex value padded with
+ * the given size.
```
bzbarsky-apple, line 86: No, that's not what this function does.

src-electron/util/types.js
```diff
@@ -78,6 +78,78 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
...
+ */
+function convertFloatToHex(value, size) {
+  let res = ''
+  const arrayBuffer = new ArrayBuffer(size)
+  const dataView = new DataView(arrayBuffer)
+  dataView.setFloat32(0, value, false)
+  for (let i = 0; i < size; i++) {
+    let t = dataView.getUint8(i).toString(16)
+    res += t == 0 ? '00' : t
```
bzbarsky-apple, line 95: Why is this `t == 0` check here?

In any case, I don't think this is right.  If we get a byte with value `1` this will append `"1"` to the string, whereas it should append `"01"`, no?

src-electron/util/types.js
```diff
@@ -78,6 +78,78 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
...
+}
+
+/**
+ *
+ * @param {*} value
+ * @param {*} size
+ * @returns the negative hex value for a given positive hex value padded with
+ * the given size.
+ */
+function asNegativeHexValue(value, size) {
```
bzbarsky-apple, line 107: This function is being used for floats as well, per above, and for floats this makes absolutely no sense....

I am also not sure why we are doing this transformation on the hex-encoded string for integers.  It seems like for integers (at least 1/2/4-byte ones; 8-byte is complicated) we could just detect that the value is negative and convert `2^{1/2/4} + N`, to hex as a positive value.  For 8-byte integers things are more annoying....  But in practice if the actual value is in the [-2^48, 0] range we can do `2^48 + value`, convert to a 6-byte integer, and pad out with 0xFF at the front.  And if it's outside that range, we should probably throw for now.  Not least because if it picks up another few bits we will stop being able to represent it exactly anyway and all this code will go awry, right?

src-electron/util/types.js
```diff
@@ -100,17 +172,34 @@ function longTypeDefaultValue(size, type, value) {
...
-    // First strip off the 0x.
-    if (value.startsWith('0x') || value.startsWith('0X'))
-      value = value.substring(2)
-
-    // Now pad the zeroes to the required size
-    if (size > 0) {
-      while (value.length / 2 < size) {
-        value = '0' + value
+    let temp = ''
+    // Positive or Negative Value
```
bzbarsky-apple, line 176: There's got to be a way to share more of this code with the code above....

test/gen-zigbee.test.js
```diff
@@ -389,17 +389,17 @@ test(
...
     // Test GENERATED_DEFAULTS for same attribute name but different side of cluster, compare it to above test
     expect(cfgVer2).toContain(
-      '0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0F, 0x0G, 0x0H, /* 28,DEFAULT value for cluster: Green Power, attribute: gp link key, side: client */'
+      '0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0F, 0x0F, 0x0F, /* 28,DEFAULT value for cluster: Green Power, attribute: gp link key, side: client */'
     )
 
     // Test GENERATED_DEFAULTS little endian for attribute of size > 8 is same as big endian. Bytes are not inverted
     expect(cfgVer2).not.toContain(
-      `0x0H, 0x0G, 0x0F, 0x0D, 0x0C, 0x0B, 0x0A, 0x09, 0x08, 0x07, 0x06, 0x05, 0x04, 0x03, 0x02, 0x01, /* 12,DEFAULT value for cluster: Green Power, attribute: gp link key, side: client*/`
+      `0x0F, 0x0F, 0x0F, 0x0D, 0x0C, 0x0B, 0x0A, 0x09, 0x08, 0x07, 0x06, 0x05, 0x04, 0x03, 0x02, 0x01, /* 12,DEFAULT value for cluster: Green Power, attribute: gp link key, side: client*/`
```
bzbarsky-apple, line 402: Why is this "12.DEFAULT" while the above toContain has "28,DEFAULT"?  Won't this cause the 12 vs 28 thing to decide what happens no matter what the bytes are?

test/helpers.test.js
```diff
@@ -244,6 +244,32 @@ test(
...
+  },
+  testUtil.timeout.short()
+)
+
+test(
+  'Generated Macro Float Value',
+  () => {
+    let options = { hash: { endian: 'little' } }
+    return zclHelper
+      .as_generated_default_macro('17.0', 8, options)
```
bzbarsky-apple, line 265: Needs tests for negative doubles.

bzbarsky-apple (changes requested):

src-electron/util/types.js
brdandu: Made changes for this.

test/gen-zigbee.test.js
brdandu: The tests are just searching for the content that has been generated. The order in the tests is not the order in which code is generated.

src-electron/generator/helper-zcl.js
```diff
@@ -2233,27 +2233,46 @@ async function if_mfg_specific_cluster(clusterId, options) {
...
   let default_macro_signature = ''
+  let temp = ''
   if (attributeSize > 2) {
-    let default_macro = helperC
-      .asHex(value, null, null)
-      .replace('0x', '')
-      .match(/.{1,2}/g)
+    // String value
+    if (isNaN(value)) {
+      return format_zcl_string_as_characters_for_generated_defaults(
```
bzbarsky-apple, line 2240: This is not used for Matter, I assume?  If it is, it's not handling long strings correctly.....

src-electron/util/types.js
```diff
@@ -78,6 +78,77 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
   }
 }
 
+/**
+ *
+ * @param {*} value
+ * @param {*} size
+ * @returns The big endian value for a given float value padded with
+ * the given size.
```
bzbarsky-apple, line 86: This should probably document the "0x" prefix, that the value will be hex, etc, etc.

src-electron/util/types.js
```diff
@@ -78,6 +78,77 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
...
+  }
+  return '0x' + res
+}
+
+/**
+ *
+ * @param {*} value
+ * @param {*} size
+ * @returns The big endian value for a given integer value padded with
+ * the given size.
```
bzbarsky-apple, line 116: Again, should document the hex and 0x bits.

src-electron/util/types.js
```diff
@@ -78,6 +78,77 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
...
+ */
+function convertIntToBigEndian(value, size) {
+  const arrayBuffer = new ArrayBuffer(size)
+  const dataView = new DataView(arrayBuffer)
+  if (size == 1) {
+    dataView.setInt8(0, value, false)
+  } else if (size == 2) {
+    dataView.setInt16(0, value, false)
+  } else if (size == 3) {
+    dataView.setInt32(0, value, false)
```
bzbarsky-apple, line 126: This is not going to do the right thing.  If you're setting as big-endian, this will leave the first byte 0 and set the 2nd through 4th bytes to the value, right?  But then you're going to look at the 1st to 3rd bytes below.

Same thing for sizes 5, 6, 7.

Either we need to do this little-endian, or we need to adjust where we start looking for the value or something.

src-electron/util/types.js
```diff
@@ -78,6 +78,77 @@ async function typeSizeAttribute(db, zclPackageId, at, defaultValue = null) {
...
+  } else if (size == 4) {
+    dataView.setInt32(0, value, false)
+  } else if (size == 5) {
+    dataView.setBigInt64(0, value, false)
```
bzbarsky-apple, line 130: Are you sure this will work right?  Why would `value` be a BigInt object here?  It almost certainly won't be.....  Is the expectation that `value` is a string and hence will auto-convert to BigInt?

src-electron/util/types.js
```diff
@@ -100,17 +172,34 @@ function longTypeDefaultValue(size, type, value) {
...
-    // Now pad the zeroes to the required size
-    if (size > 0) {
-      while (value.length / 2 < size) {
-        value = '0' + value
+    let temp = ''
+    // Float value
+    if (!isNaN(value) && value.toString().indexOf('.') != -1) {
+      temp = convertFloatToBigEndian(value, size)
+    } else {
+      // Postive value
```
bzbarsky-apple, line 180:
```suggestion
      // Positive value
```

test/helpers.test.js
```diff
@@ -244,6 +244,45 @@ test(
   testUtil.timeout.short()
 )
 
+test(
+  'Generated Macro Negative Value',
+  () => {
+    let options = { hash: { endian: 'big' } }
```
bzbarsky-apple, line 250: Should test all of these in both endian-nesses.

src-electron/generator/helper-zcl.js
brdandu: Yes this is not used by matter. Why do you say that?

bzbarsky-apple: `format_zcl_string_as_characters_for_generated_defaults` assumes the string length is a singly byte.  For long strings it's two bytes.

src-electron/util/types.js
brdandu: I believe this should work. I was mostly thinking in terms of the range. Switching it to setInt32 for a negative integer with attribute size 5 won't work unless I use dataView.setBigInt64

bzbarsky-apple: Sure.  The big question is whether this will throw an exception (which it will if `Value` is a Number: implicit conversion from Number to BigInt is not allowed in JS).  It would be safer to do `BigInt(value)`.

brdandu: Switched to BigInt

bzbarsky-apple (approved):

Issue #609, opened 2022-07-15
Need to support fabric-scoped commands


bzbarsky-apple:
We need to be able to mark commands as fabric-scoped in the XML.

The relevant pieces are:
1. Have `isFabricScoped="true"` in the XML on the relevant commands.
2. Add a field to the COMMAND table to store the value (similar to MUST_USE_TIMED_INVOKE).
3. Fix XML import to set the COMMAND field based on the XML.
4. Allow some way to do things with fabric-scoped commands via helpers.  Either expose the state in the data `zcl_commands` fetches, or have a custom helper like `zcl_commands_that_need_timed_invoke`; no really strong opinions here, though the `zcl_commands` approach seems a bit more flexible.

PR #611, opened 2022-07-18
Fabric scope

tecimovic:
Issue: https://github.com/project-chip/zap/issues/609

This PR adds fabric-scope flag to commands, events and attributes, adds them to all relevant queries, etc.
The schema is changed, the zcl.xsd has a new valid field `isFabriScoped="true"`.


src-electron/db/db-mapping.js
```diff
@@ -112,6 +112,7 @@ exports.map = {
         x.REPORTING_POLICY == dbEnums.reportingPolicy.suggested,
       reportingPolicy: x.REPORTING_POLICY,
       isSceneRequired: dbApi.fromDbBool(x.IS_SCENE_REQUIRED),
+      isFabricScoped: dbApi.fromDbBool(x.IS_FABRIC_SCOPED),
```
bzbarsky-apple, line 115: Attributes do not have a concept of fabric scoping in this sense, so I'm not sure we should add this here.

What has such a concept is _structs_, and adding this for struct definitions would be useful.  Then we could get rid of the `struct_is_fabric_scoped` bit in helper-zcl.js and just condition that whole "figure out what the fabric index field and the fabric sensitive fields bits are" on the fabric-scoped boolean from the DB.

src-electron/db/db-mapping.js
```diff
@@ -142,6 +143,7 @@ exports.map = {
       description: x.DESCRIPTION,
       side: x.SIDE,
       isOptional: dbApi.fromDbBool(x.IS_OPTIONAL),
+      isFabricScoped: dbApi.fromDbBool(x.IS_FABRIC_SCOPED),
```
bzbarsky-apple, line 146: Events do not have a concept of fabric-scoped.  They _do_ have a concept of fabric-sensitive.  And if we do this we can get rid of the `event_is_fabric_scoped` thing in helper-zcl.js, right?  Probably as a followup....

bzbarsky-apple (changes requested):

src-electron/db/db-mapping.js
tecimovic: Ok. I'll just rename this to IS_FABRIC_SENSITIVE.

src-electron/db/db-mapping.js
bzbarsky-apple, line 115: Again: attributes cannot be fabric scoped.

Issue #612, opened 2022-07-19
Add a way for first_unused_enum_value to produce hex output


bzbarsky-apple:
See https://github.com/project-chip/connectedhomeip/pull/20907#discussion_r924167962 -- Right now we can't easily get a hex value for the first unused enum value....

PR #613, opened 2022-07-19
Add the format="hex" attribute to the first_unused_enum_value helper.

tecimovic:
https://github.com/project-chip/zap/issues/612

### DIFF
--- a/src-electron/generator/helper-zcl.js
+++ b/src-electron/generator/helper-zcl.js
@@ -211,6 +211,7 @@ async function zcl_enum_items(options) {
  */
 async function first_unused_enum_value(options) {
   let mode = options.hash.mode
+  let format = options.hash.format // "hex" or "decimal"
   let items
   if (this.enum_items) {
     items = this.enum_items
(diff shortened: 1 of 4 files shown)


src-electron/generator/helper-zcl.js
```diff
@@ -241,7 +242,18 @@ async function first_unused_enum_value(options) {
       }
     } while (isPresent)
   }
-  return unusedValue
+
+  if (format == 'hex') {
+    let out = unusedValue.toString(16)
+    if (this.size && this.size > 0) {
+      while (out.length < this.size * 2) {
```
bzbarsky-apple, line 249: What is this.size going to be in practice?

bzbarsky-apple (approved):

src-electron/generator/helper-zcl.js
tecimovic: Number of bytes of the enum. This just pads the hex with zeros, so you get '01' instead of '1', so you can have nicely aligned values.

If you don't want this logic, let me know and I'll delete it.

PR #619, opened 2022-07-21
Switching the is helpers to return boolean

brdandu:
- Switched the isEnum/Bitmap/Struct helpers to return booleans. Changed the tests to reflect the same
- Updated the asUnderlyingZclTypeWithPackageId and zcl_command_argument_data_type helpers to account for tyhe boolean returns and cleaned up the helpers

Github: ZAP#413

### DIFF
--- a/src-electron/util/zcl-util.js
+++ b/src-electron/util/zcl-util.js
@@ -440,7 +440,7 @@ async function dataTypeCharacterFormatter(
 function isEnum(db, enum_name, packageId) {
   return queryZcl
     .selectEnumByName(db, enum_name, packageId)
-    .then((enums) => (enums ? dbEnum.zclType.enum : dbEnum.zclType.unknown))
+    .then((enums) => (enums ? true : false))
 }
 
 /**
@@ -454,7 +454,7 @@ function isEnum(db, enum_name, packageId) {
 function isStruct(db, struct_name, packageId) {
   return queryZcl
     .selectStructByName(db, struct_name, packageId)
-    .then((st) => (st ? dbEnum.zclType.struct : dbEnum.zclType.unknown))
+    .then((st) => (st ? true : false))
 }
 
 /**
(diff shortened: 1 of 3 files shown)


brdandu:
A couple of worries on this.

I have some concerns with backwards compatibility for this incase we were/are using these in our zigbee pro and matter templates before. The isEnum/Bitmap/Struct helpers go through promises so using them with an if can result in an if(unresolved promise) which can return true. I am wondering if I should not export these from the helper-zcl.js for templates and just use these internally.

bzbarsky-apple:
> can result in an if(unresolved promise) which will return true

This is a huge footgun in general.  Is there any way to make that fail?

src-electron/util/zcl-util.js
```diff
@@ -440,7 +440,7 @@ async function dataTypeCharacterFormatter(
 function isEnum(db, enum_name, packageId) {
   return queryZcl
     .selectEnumByName(db, enum_name, packageId)
-    .then((enums) => (enums ? dbEnum.zclType.enum : dbEnum.zclType.unknown))
+    .then((enums) => (enums ? true : false))
```
bzbarsky-apple, line 443: Or could do:
```suggestion
    .then((enums) => (Boolean(enums)))
```
but either way.

src-electron/util/zcl-util.js
```diff
@@ -468,7 +468,7 @@ function isStruct(db, struct_name, packageId) {
 function isBitmap(db, bitmap_name, packageId) {
   return queryZcl
     .selectBitmapByName(db, packageId, bitmap_name)
-    .then((st) => (st ? dbEnum.zclType.bitmap : dbEnum.zclType.unknown))
+    .then((st) => (st ? true : false))
```
bzbarsky-apple, line 471: Pre-existing but why "st" (apart from "we copied the struct code")?

bzbarsky-apple (approved):

PR #624, opened 2022-07-22
Fixing packageIds issue for custom xml

brdandu:
This needs to go side by side with a CHIP repo PR because zap code outside the zap repo is still using packageId instead of packageIds
Switching to use templateUtil.ensureZclPackageIds instead of templateUtil.ensureZclPackageId to handle the custom xml generation which is not being accounted for currently when loading custom xml from the UI

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -40,11 +40,11 @@
 <dt><a href="#module_DB API_ device type database access">DB API: device type database access</a></dt>
 <dd><p>This module provides queries for device types.</p>
 </dd>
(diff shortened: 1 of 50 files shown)


tecimovic:
@brdandu : I've alread merged all the Matter work into zap repo. The PR for the CHIP main repo is ongoing.

tecimovic (approved):
I ran a full matter regen with the branch that doesn't have the JS code on the stack any more, and it generates accurately.

PR #627, opened 2022-07-25
Helper categorization

tecimovic:


src-electron/generator/generation-engine.js
```diff
@@ -446,70 +446,72 @@ async function generateAllTemplates(
...
+  let packages = await queryPackage.getPackageByParent(
+    genResult.db,
+    genTemplateJsonPkg.id
+  )
+  let generationTemplates = []
+  let helperPromises = []
+  let partialPromises = []
+  let overridePath = null
+
+  let hb = templateEngine.hbInstance()
```
jingteng25742, line 458: `2. Desingletonize the handlebars instance.`


Doesn't this make the handlebar instances into a singleton?

PR #628, opened 2022-07-25
Fix FabricIndex handling in structs and events.

bzbarsky-apple:
1) Remove struct_is_fabric_scoped and event_is_fabric_scoped now that we have
   isFabricScoped/isFabricSensitive on those objects.  Add the relevant XML
   attributes instead of trying to guess based on fields.
2) Remove event_fabric_idx_field and struct_fabric_idx_field, since we are no
   longer guessing based on fields.
3) Synthesize the relevant struct item or event field with the right name, type,
   field id, per spec, if the struct is fabric-scoped or the event is
   fabric-sensitive.

This is a hard requirement for implementing Leave events properly for Matter:
those have a fabric_idx field but are _not_ fabric-scoped.

### DIFF
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -31,6 +31,10 @@ const dbEnum = require('../../src-shared/db-enum')
 const zclLoader = require('./zcl-loader')
 const _ = require('lodash')
 
+const fabricIndexFieldId = 0xfe
+const fabricIndexFieldName = 'FabricIndex'
+const fabricIndexType = 'fabric_idx'
+
 /**
  * Promises to read the JSON file and resolve all the data.
  * @param {*} ctx  Context containing information about the file
@@ -486,6 +490,22 @@ function prepareCluster(cluster, context, isExtension = false) {
           }
         })
       }
+      if (ev.isFabricSensitive) {
+        if (!ev.fields) {
+          ev.fields = []
+        }
+        ev.fields.push({
+          name: fabricIndexFieldName,
+          type: fabricIndexType,
+          isArray: false,
+          isNullable: false,
+          isOptional: false,
+          fieldIdentifier: fabricIndexFieldId,
+          introducedIn: null,
+          removedIn: null,
+        })
+      }
+
       // We only add event if it does not have removedIn
       if (ev.removedIn == null) ret.events.push(ev)
     })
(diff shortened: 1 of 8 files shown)


src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -31,6 +31,10 @@ const dbEnum = require('../../src-shared/db-enum')
 const zclLoader = require('./zcl-loader')
 const _ = require('lodash')
 
+const fabricIndexFieldId = 0xfe
+const fabricIndexFieldName = 'FabricIndex'
+const fabricIndexType = 'fabric_idx'
+
```
tecimovic, line 37: These need to be externalized somehow. We need to data-drive this via `zcl.json` somehow. There should be a field:
  "enableAutomaticFieldsLogic=true"
  which would turn this on, then these would be listed there as well, etc. You get the point.

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -486,6 +490,22 @@ function prepareCluster(cluster, context, isExtension = false) {
           }
         })
       }
+      if (ev.isFabricSensitive) {
+        if (!ev.fields) {
```
tecimovic, line 494: Same comment as above: i don't mind Zap being "smart" to automatically add stuff based on some logic, but it needs to be data driven from the SDK.

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1310,6 +1330,24 @@ async function processStructItems(db, filePath, packageId, data) {
         })
       })
     }
+
+    if (si.$.isFabricScoped == 'true') {
```
tecimovic, line 1334: Ditto.

tecimovic (commented):
@bzbarsky-apple :  for expediency sake, if you need this fast: you can merge this in, because it's not affecting Zigbee, but PLEASE log an issue to make this logic data driven and assign it to me. I don't mind doing it, but we have to do it in a way where zcl.json controls whether these "synthesized fields" do or don't happen.

src-electron/zcl/zcl-loader-silabs.js
bzbarsky-apple: Yes, I agree.  Filed https://github.com/project-chip/zap/issues/629 for that.

Issue #629, opened 2022-07-25
Make synthesis of FabricIndex event/struct fields data-driven


bzbarsky-apple:
Should be driven by `zcl.json` instead of hardcoded in zcl-loader-silabs.js.

@tecimovic

PR #634, opened 2022-07-27
Fabric index cleanup

tecimovic:
This addresses:
https://github.com/project-chip/zap/issues/629

It adds data-driven mechanism for fabric index field automatic creation.

** NOTE: this should be merged only after https://github.com/project-chip/connectedhomeip/pull/21273 is merged. **


bzbarsky-apple (approved):
Thank you!

PR #667, opened 2022-08-14
Generic async reporting + cypress fixes

tecimovic:
0. Remove dirty flag out of "Studio IDE" functionality.
1. Make it an example of a generic async reporting process.


paulr34 (approved):
Looks good. I cloned and everything is working like it was previously.

PR #688, opened 2022-08-19
Notification system (RENOP)

paulr34:
This is a Draft PR for the notification framework.

TODO:
1. **Decide how the querying of SESSION_NOTICE table will be implemented.**
_OPTIONS:_
A. Notifications should be able to be triggered at any moment. Therefore, it would make sense to have interval polling          which constantly checks for new notifications.  This is not in the PR but I have this working locally. Although this code is out of date, here is an example of how it might work https://github.com/project-chip/zap/commit/57d9486034af18b16d203a9e9c432212fdb6348e
B. If notifications are not able to be triggered at any moment and are specific for importing and exporting .zap files then there could be a function which queries the SESSION_NOTICE table after importing. This way there would be no need for polling. Basically, the notifications would only be queried when creating a session...
C. Both A and B

2. **Add notification page to the front end**
3. Test

### DIFF
--- a/src-electron/db/db-mapping.js
+++ b/src-electron/db/db-mapping.js
@@ -568,6 +568,17 @@ exports.map = {
       value: x.VALUE,
     }
   },
+  notifications: (x) => {
+    if (x == null) return undefined
+    return {
+      ref: x.SESSION_REF,
+      type: x.NOTICE_TYPE,
+      message: x.NOTICE_MESSAGE,
+      severity: x.NOTICE_SEVERITY,
+      order: x.NOTICE_ORDER,
+      display: x.DISPLAY,
+    }
+  },
 }
 
 exports.reverseMap = {
--- a/src-electron/db/query-notification.js
+++ b/src-electron/db/query-notification.js
@@ -1,0 +1,84 @@
+'use strict'
+/**
+ *
+ *    Copyright (c) 2020 Silicon Labs
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+/**
+ * This module provides notification related queries.
+ *
+ * @module DB API: session related queries.
+ */
+const dbApi = require('./db-api.js')
+const dbMapping = require('./db-mapping.js')
+/**
+ * Sets a notification in the SESSION_NOTICE table
+ *
+ * @export
+ * @param {*} db
+ * @param {*} type
+ * @param {*} status
+ * @param {*} sessionId
+ * @param {*} severity
+ */
+async function setNotification(db, type, status, sessionId, severity) {
+  return dbApi.dbUpdate(
+    db,
+    'INSERT INTO SESSION_NOTICE ( SESSION_REF, NOTICE_TYPE, NOTICE_MESSAGE, NOTICE_SEVERITY, DISPLAY) VALUES ( ?, ?, ?, ?, ?)',
+    [sessionId, type, status, severity, 1]
+  )
+}
+async function setAllNotificationDisplay(db, sessionId) {
+  return dbApi.dbUpdate(
+    db,
+    'UPDATE SESSION_NOTICE SET DISPLAY = ? WHERE SESSION_REF = ?',
+    [0, sessionId]
+  )
+}
+/**
+ * Deletes a notification from the SESSION_NOTICE table
+ *
+ * @export
+ * @param {*} db
+ * @param {*} upgrade
+ * @param {*} status
+ */
+async function deleteNotification(db, upgrade, status) {
+  return dbApi.dbUpdate(
+    db,
+    'DELETE FROM SESSION_NOTICE WHERE ( SESSION_REF, STATUS ) = ( ?, ? )',
+    [upgrade, status]
+  )
+}
+/**
+ * Retrieves a notification from the SESSION_NOTICE table
+ *
+ * @export
+ * @param {*} db
+ */
+async function getNotification(db, sessionId) {
+  let rows = []
+  rows = await dbApi.dbAll(
+    db,
+    'SELECT * FROM SESSION_NOTICE WHERE SESSION_REF = ?',
+    [sessionId]
+  )
+  return rows.map(dbMapping.map.notifications)
+}
+// exports
+exports.setNotification = setNotification
+exports.setAllNotificationDisplay = setAllNotificationDisplay
+exports.deleteNotification = deleteNotification
+exports.getNotification = getNotification
+//# sourceMappingURL=query-notification.js.map
--- a/src-shared/db-enum.js
+++ b/src-shared/db-enum.js
@@ -90,6 +90,7 @@ exports.pathRelativity = {
 exports.wsCategory = {
   generic: 'generic',
   dirtyFlag: 'dirtyFlag',
+  upgrade: 'upgrade',
   validation: 'validation',
   sessionCreationError: 'sessionCreationError',
   componentUpdateStatus: 'componentUpdateStatus',
--- a/src/layouts/ZclLayout.vue
+++ b/src/layouts/ZclLayout.vue
@@ -144,9 +144,9 @@ limitations under the License.
                     clickable
                     v-close-popup
                     @click="
-                    generationButtonText = file.category
-                    getGeneratedFile(file.category)
-                  "
+                      generationButtonText = file.category
+                      getGeneratedFile(file.category)
+                    "
                     :label="generationButtonText"
                   >
                     <q-item-section>
@@ -159,7 +159,7 @@ limitations under the License.
                 <template>
                   <div class="q-ma-md">
                     <q-scroll-area style="height: 70vh" ref="generationScroll">
-                    <pre class="q-ma-none container">{{
+                      <pre class="q-ma-none container">{{
                         generationData
                       }}</pre>
                       <q-scroll-observer @scroll="onScroll" />
@@ -222,7 +222,7 @@ export default {
         title: 'Select directory to generate into',
         mode: 'directory',
         defaultPath: currentPath,
-        buttonLabel: 'Generate'
+        buttonLabel: 'Generate',
       })
     },
     getGeneratedFiles() {
--- a/cypress/integration/devtools/devtools.spec.js
+++ b/cypress/integration/devtools/devtools.spec.js
@@ -16,9 +16,7 @@ describe('Testing visibility of devtools option', () => {
 
   it('check if devtools option gets visible', () => {
     cy.setZclProperties()
-    cy.get(
-      'a.q-btn > .q-btn__wrapper > .q-btn__content > .material-icons'
-    ).click()
+    cy.get('#preference > .q-btn__wrapper > .q-btn__content > .q-icon').click()
     cy.get('[aria-label="Enable development tools"] > .q-toggle__label').click()
     cy.get('.q-btn__wrapper').contains('Back').click()
     cy.get('.q-gutter-y-md > :nth-child(1)').should('contain', 'Dev Tools')
--- a/src-electron/db/zap-schema.sql
+++ b/src-electron/db/zap-schema.sql
@@ -1117,4 +1117,17 @@ CREATE TABLE IF NOT EXISTS "SETTING" (
   "VALUE" text,
   UNIQUE(CATEGORY, KEY)
 );
+/*
+  Session Notification table
+*/
+
+ CREATE TABLE IF NOT EXISTS "SESSION_NOTICE" (
+   "SESSION_REF" integer,
+   "NOTICE_TYPE" text,
+   "NOTICE_MESSAGE" text,
+   "NOTICE_SEVERITY" integer,
+   "NOTICE_ORDER"  integer primary key autoincrement,
+   "DISPLAY" integer,
+   foreign key (SESSION_REF) references SESSION(SESSION_ID) on delete cascade
+ );
 /* EO SCHEMA */
--- a/src-electron/importexport/import-isc.js
+++ b/src-electron/importexport/import-isc.js
@@ -28,6 +28,7 @@ const util = require('../util/util')
 const dbEnum = require('../../src-shared/db-enum')
 const restApi = require('../../src-shared/rest-api')
 const env = require('../util/env')
+const notification = require('../db/query-notification.js')
 
 /**
  * Locates or adds an attribute, and returns it.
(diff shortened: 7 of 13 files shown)


src-electron/db/db-mapping.js
```diff
@@ -142,7 +142,6 @@ exports.map = {
       description: x.DESCRIPTION,
       side: x.SIDE,
       isOptional: dbApi.fromDbBool(x.IS_OPTIONAL),
-      isFabricSensitive: dbApi.fromDbBool(x.IS_FABRIC_SENSITIVE),
```
brdandu: Is this related to your change?

src-shared/db-enum.js
```diff
@@ -41,9 +41,6 @@ exports.packageOptionCategory = {
   profileCodes: 'profileCodes',
   validationTimersFlags: 'validationTimersFlags',
   ui: 'ui',
-  helperCategories: 'helperCategories',
-  helperAliases: 'helperAliases',
-  resources: 'resources',
```
brdandu, line 46: Are the removals here related to your changes?

src-electron/db/db-mapping.js
paulr34: no. I just fixed it.

src-shared/db-enum.js
paulr34: no. it was an accident which I just fixed

brdandu:
> This is a Draft PR for the notification framework.
> 
> TODO:
> 
> 1. **Decide how the querying of SESSION_NOTICE table will be implemented.**
>    _OPTIONS:_
>    A. Notifications should be able to be triggered at any moment. Therefore, it would make sense to have interval polling          which constantly checks for new notifications.  This is not in the PR but I have this working locally. Although this code is out of date, here is an example of how it might work [57d9486](https://github.com/project-chip/zap/commit/57d9486034af18b16d203a9e9c432212fdb6348e)
>    B. If notifications are not able to be triggered at any moment and are specific for importing and exporting .zap files then there could be a function at the end of importing a file which queries the SESSION_NOTICE table. This way there would be no need for polling. Basically, the notifications would only be queried when creating a session...
>    C. Both A and B
> 2. **Add notification page to the front end**
> 3. Test

This whole triggering the notification table makes it complicated in my opinion. How about we just introduce an additional pane along with our menu bar
Screen Shot 2022-08-19 at 3 31 19 PM
In here we could have another icon for notifications. If a user clicks on this then we can show all the notifications for that session. This can simply show the results of "getNotification". To make it better if we want to alert the user regarding notifications then this notification icon can be turned to red until the user presses this once. Once they press it then it can go back to say yellow. If a new notification enters the session notice table again then turn this back to red until the user presses on it again.
For when do we turn the notification icon to red? => I would say every time there is a new entry to the session_notice table for that session.
My understanding is this should not require any kind of polling mechanism. Just the UI reflecting on the state of the backend.

paulr34:
> > This is a Draft PR for the notification framework.
> > TODO:
> > 
> > 1. **Decide how the querying of SESSION_NOTICE table will be implemented.**
> >    _OPTIONS:_
> >    A. Notifications should be able to be triggered at any moment. Therefore, it would make sense to have interval polling          which constantly checks for new notifications.  This is not in the PR but I have this working locally. Although this code is out of date, here is an example of how it might work [57d9486](https://github.com/project-chip/zap/commit/57d9486034af18b16d203a9e9c432212fdb6348e)
> >    B. If notifications are not able to be triggered at any moment and are specific for importing and exporting .zap files then there could be a function at the end of importing a file which queries the SESSION_NOTICE table. This way there would be no need for polling. Basically, the notifications would only be queried when creating a session...
> >    C. Both A and B
> > 2. **Add notification page to the front end**
> > 3. Test
> 
> This whole triggering the notification table makes it complicated in my opinion. How about we just introduce an additional pane along with our menu bar Screen Shot 2022-08-19 at 3 31 19 PM In here we could have another icon for notifications. If a user clicks on this then we can show all the notifications for that session. This can simply show the results of "getNotification". To make it better if we want to alert the user regarding notifications then this notification icon can be turned to red until the user presses this once. Once they press it then it can go back to say yellow. If a new notification enters the session notice table again then turn this back to red until the user presses on it again. For when do we turn the notification icon to red? => I would say every time there is a new entry to the session_notice table for that session. My understanding is this should not require any kind of polling mechanism. Just the UI reflecting on the state of the backend.

That is an interesting idea. It would keep the user from being bombarded with notifications especially if they just upgraded their extremely old .zap file.

Re: "For when do we turn the notification icon to red? => I would say every time there is a new entry to the session_notice table for that session. My understanding is this should not require any kind of polling mechanism. Just the UI reflecting on the state of the backend." How would the front end know if a new notification was added to the table without polling? Unless it only queries during the creation of a session or in other words at fixed moments, there would need to be an interval polling mechanism.

Case 1: POLLING (This is how the DIRTY FLAG works)
1.) User is playing around with their .zap file and messes up big time.
2.) The back end then creates a new entry in the table pointing at that sessionID
3.) Since the back end periodically polls that table, it will know to communicate to the front end that there is a new notification.

Case 2: NOT POLLING 
1.) User imports old .zap file with lots of upgrades which create notifications in the table
2.) After importing, a query is called to get all of the notifications that were created. 
3.) The front end displays them somehow someway

Basically, as far as I can tell, if the user can cause a change that requires an addition to the SESSION_NOTICE table at any random moment then we will need a polling mechanism. If not, then we can put a query in a line of code after all possible notifications have been created.

brdandu:
> > > This is a Draft PR for the notification framework.
> > > TODO:
> > > 
> > > 1. **Decide how the querying of SESSION_NOTICE table will be implemented.**
> > >    _OPTIONS:_
> > >    A. Notifications should be able to be triggered at any moment. Therefore, it would make sense to have interval polling          which constantly checks for new notifications.  This is not in the PR but I have this working locally. Although this code is out of date, here is an example of how it might work [57d9486](https://github.com/project-chip/zap/commit/57d9486034af18b16d203a9e9c432212fdb6348e)
> > >    B. If notifications are not able to be triggered at any moment and are specific for importing and exporting .zap files then there could be a function at the end of importing a file which queries the SESSION_NOTICE table. This way there would be no need for polling. Basically, the notifications would only be queried when creating a session...
> > >    C. Both A and B
> > > 2. **Add notification page to the front end**
> > > 3. Test
> > 
> > 
> > This whole triggering the notification table makes it complicated in my opinion. How about we just introduce an additional pane along with our menu bar Screen Shot 2022-08-19 at 3 31 19 PM In here we could have another icon for notifications. If a user clicks on this then we can show all the notifications for that session. This can simply show the results of "getNotification". To make it better if we want to alert the user regarding notifications then this notification icon can be turned to red until the user presses this once. Once they press it then it can go back to say yellow. If a new notification enters the session notice table again then turn this back to red until the user presses on it again. For when do we turn the notification icon to red? => I would say every time there is a new entry to the session_notice table for that session. My understanding is this should not require any kind of polling mechanism. Just the UI reflecting on the state of the backend.
> 
> That is an interesting idea. It would keep the user from being bombarded with notifications especially if they just upgraded their extremely old .zap file.
> 
> Re: "For when do we turn the notification icon to red? => I would say every time there is a new entry to the session_notice table for that session. My understanding is this should not require any kind of polling mechanism. Just the UI reflecting on the state of the backend." How would the front end know if a new notification was added to the table without polling? Unless it only queries during the creation of a session or in other words at fixed moments, there would need to be an interval polling mechanism.
> 
> Case 1: POLLING (This is how the DIRTY FLAG works) 1.) User is playing around with their .zap file and messes up big time. 2.) The back end then creates a new entry in the table pointing at that sessionID 3.) Since the back end periodically polls that table, it will know to communicate to the front end that there is a new notification.
> 
> Case 2: NOT POLLING 1.) User imports old .zap file with lots of upgrades which create notifications in the table 2.) After importing, a query is called to get all of the notifications that were created. 3.) The front end displays them somehow someway
> 
> Basically, as far as I can tell, if the user can cause a change that requires an addition to the SESSION_NOTICE table at any random moment then we will need a polling mechanism. If not, then we can put a query in a line of code after all possible notifications have been created.

For turning the colors of the notification icon, I was hoping there is some form of server push notification to the client which can just do this. Can we use the web socket to throw a push notification(from the backend/server) when there is a new entry in the session_notice table? The browser/ui can update when it gets this notification. I am not thinking of this as a polling mechanism but just an open connection for client and server to interact using the web socket. However, yes we do need a polling mechanism on the client or an open connection between client and server to update tell the user there are new notifications.

jingteng25742:
> For turning the colors of the notification icon, I was hoping there is some form of server push notification to the client which can just do this. Can we use the web socket to throw a push notification(from the backend/server) when there is a new entry in the session_notice table? The browser/ui can update when it gets this notification. I am not thinking of this as a polling mechanism but just an open connection for client and server to interact using the web socket. However, yes we do need a polling mechanism on the client or an open connection between client and server to update tell the user there are new notifications.

We already have the web socket notification scheme setup since it was necessary for Studio integration.
For the flow of the DIRTY FLAG, it was necessary due to how Studio/Eclipse's dirty flag worked internally.
For the notification, I like @brdandu 's idea of a notification icon / window for listing all notifications. e.g. similar to social media event history.
For init features like opening .zap, the notification can be pushed during some init phase and rest can be all async'ed based. We should use DB "TRIGGER" / API wrapper to watch / trigger operations to the DB if polling was only required for this reason.

paulr34:
> > For turning the colors of the notification icon, I was hoping there is some form of server push notification to the client which can just do this. Can we use the web socket to throw a push notification(from the backend/server) when there is a new entry in the session_notice table? The browser/ui can update when it gets this notification. I am not thinking of this as a polling mechanism but just an open connection for client and server to interact using the web socket. However, yes we do need a polling mechanism on the client or an open connection between client and server to update tell the user there are new notifications.
> 
> We already have the web socket notification scheme setup since it was necessary for Studio integration. For the flow of the DIRTY FLAG, it was necessary due to how Studio/Eclipse's dirty flag worked internally. For the notification, I like @brdandu 's idea of a notification icon / window for listing all notifications. e.g. similar to social media event history. For init features like opening .zap, the notification can be pushed during some init phase and rest can be all async'ed based. We should use DB "TRIGGER" / API wrapper to watch / trigger operations to the DB if polling was only required for this reason.

Triggering a change to the database works for changing the SESSION_NOTICE table but how would the front end know that the SESSION_NOTICE table was changed without constantly checking it?

We could have a DB trigger execute a function which communicated with the front end but this has not been implemented before while using an interval for polling the DB is already implemented.

As far as I can tell, the difference between the DIRTY FLAG and other notifications you mention is that the DIRTY FLAG can be changed at any moment rather than at a fixed moment.  This is why I have used it as a model for this project.  My assumption is we want a system where a notification can be created and used at any moment.

src-electron/server/ws-server.js
```diff
@@ -145,7 +156,9 @@ function sendWebSocketData(socket, category, payload) {
  * @param {*} msg
  */
 function sendWebSocketMessage(socket, msg) {
-  doSend(socket, msg)
+  return new Promise((resolve, reject) => {
```
jingteng25742, line 159: why is doSend wrapped within a Promise block? doSend() does not intentionally return any value to be consumed later.

src-electron/server/ws-server.js
```diff
@@ -134,9 +144,10 @@ function sendWebSocketData(socket, category, payload) {
     category: category,
     payload: payload,
   }
-  doSend(socket, obj)
+  return new Promise((resolve, reject) => {
+    resolve(doSend(socket, obj))
```
jingteng25742, line 148: why is doSend wrapped within a Promise block? doSend() does not intentionally return any value to be consumed later.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "type": "commonjs",
   "name": "zap",
-  "version": "0.0.0",
+  "version": "2022.10.27",
```
jingteng25742, line 4: discard this change since the version is auto-generated during build / release time.

src-electron/util/async-reporting.js
```diff
@@ -38,7 +39,6 @@ let asyncReports = {
     sessionBased: true,
   },
 }
-
 /**
```
jingteng25742, line 42: should leave this for spacing

jingteng25742 (changes requested):

package.json
tecimovic: @paulr34 , you do not have husky hooks properly initialized. If they are, then they do not let you commit package.json, unless the `version` field is set to "0.0.0".
Make sure you run `husky install`.

tecimovic: LIke this:
```
[~/git/zap (master)]$ git commit -a
🚀   > Pretty quick reformat...
🔍  Finding changed files since git revision aa0c1150.
🎯  Found 1 changed file.
✅  Everything is awesome!
🚀   > JSdoc regen...
🚀   > Eslint...
🚀   > Reset version in package.json to 0.0.0...
⛔ Version in package.json was not set correctly. It was automatically updated to 0.0.0. Review and commit again, please.
husky - pre-commit hook exited with code 1 (error)
```

paulr34: okay thanks

paulr34: @tecimovic I think I do have husky installed because I have that output, but after I committed, I pulled, and was forced to checkout package.json for some reason. Something is wrong..

src-electron/server/ws-server.js
paulr34: It needs to return something because the function that calls it is expecting a resolution. I can probably just take this out for now and just go with the page and have the pop up as a later issue.

src-electron/server/ws-server.js
paulr34: duplicate comment

paulr34: We discussed adding this to make sure something was returned, ill just take it out for now because it is not being used

package.json
paulr34: fixed

src-electron/util/async-reporting.js
paulr34: moving this code to a different PR

src-electron/db/query-notification.js
```diff
@@ -0,0 +1,84 @@
...
+ * @param {*} severity
+ */
+async function setNotification(db, type, status, sessionId, severity) {
+  return dbApi.dbUpdate(
+    db,
+    'INSERT INTO SESSION_NOTICE ( SESSION_REF, NOTICE_TYPE, NOTICE_MESSAGE, NOTICE_SEVERITY, DISPLAY) VALUES ( ?, ?, ?, ?, ?)',
+    [sessionId, type, status, severity, 1]
+  )
+}
+async function setAllNotificationDisplay(db, sessionId) {
```
brdandu, line 42: Add function description for this function as well.

src/layouts/ZclLayout.vue
```diff
@@ -144,9 +144,9 @@ limitations under the License.
                     clickable
                     v-close-popup
                     @click="
-                    generationButtonText = file.category
-                    getGeneratedFile(file.category)
-                  "
+                      generationButtonText = file.category
+                      getGeneratedFile(file.category)
+                    "
```
brdandu, line 149: See if you can get rid of these additional diffs so as to maintain the git history with a actual changes.

brdandu (approved):

src/layouts/ZclLayout.vue
paulr34: I tried getting rid of those but when I committed and pushed there were "no changes to show". I assume it has to do with our formatter?

jingteng25742 (approved):

PR #625, opened 2022-07-25
Vue tour

tecimovic:
Ongoing PR from `vue-tour` branch, so we have CI and we're ready to merge this in.

PR #690, opened 2022-08-20
fix vue tour problems

Amirreza-Ghasemkhani:
https://github.com/project-chip/zap/pull/625


src/components/ZclEndpointCard.vue
```diff
@@ -317,7 +293,7 @@ export default {
               }
             }
           })
-          this.selectedservers = [...enabledServers, ...enabledServers]
+          this.selectedservers = [...enabledServers, ...enabledClients]
```
jingteng25742, line 296: why did it change to enabledClients?

src-electron/db/query-package.js
```diff
@@ -71,6 +71,25 @@ async function getPackageByParent(db, parentId) {
     .then((rows) => rows.map(dbMapping.map.package))
 }
 
+/**
```
jingteng25742, line 74: @Amirreza-Ghasemkhani , 
I don't think this is part of our vue tour code. 
Did this get accidentally introduced as part of a rebase/merge?

src-electron/generator/matter/app/zap-templates/common/ChipTypesHelper.js
```diff
@@ -54,5 +54,5 @@ function asBasicType(type) {
 exports.asBasicType = asBasicType;
 exports.meta = {
   category: dbEnum.helperCategory.matter,
-  alias: ['common/ChipTypesHelper.js'],
+  alias: ['common/ChipTypesHelper.js', 'matter-chip-types-helper'],
```
jingteng25742, line 57: accident w/ merge?

jingteng25742 (changes requested):
A lot of matter code seem to be accidentally introduced as part of the merge.

src/components/ZclEndpointCard.vue
Amirreza-Ghasemkhani: Here we have to enableCluster but we should use enableClient for one of them I change that to fix issue

src-electron/db/query-package.js
Amirreza-Ghasemkhani: I tried to fix vuetour on matter and for that I need allpackages or uioption I tried to fix that but we need work on some of apies so after discussion we decide to work on this problem in another branch

src-electron/generator/matter/app/zap-templates/common/ChipTypesHelper.js
Amirreza-Ghasemkhani: After rebasing that with master this was appeared I think

Issue #528, opened 2022-05-25
Feature Request: Duplicate endpoint configuration


ldoering:
When working with a device containing multiple endpoints with the same functionality on multiple endpoints, it would be much easier if the configuration can be done once and duplicated for all desired endpoints. The current approach is to configure all endpoints manually, which is quite time consuming.

Amirreza-Ghasemkhani:
https://github.com/project-chip/zap/pull/695

ldoering:
When testing this feature, I noticed that the enabled commands seem to be ignored and had to enabled manually. 
Is this a bug or a feature?

PR #695, opened 2022-08-26
Duplicate Endpoint

pedramk:
#528


src-electron/db/query-config.js
```diff
@@ -20,1160 +20,1202 @@
  *
  * @module DB API: user configuration queries against the database.
  */
-const dbApi = require('./db-api.js')
```
jingteng25742, line 23: The whole file seem to have changed and makes it very difficult to actually review the changes.
Can you fix this?

src-electron/rest/user-data.js
```diff
@@ -20,744 +20,789 @@
  *
  * @module REST API: user data
  */
-const fs = require('fs')
```
jingteng25742, line 23: The whole file seem to have changed and makes it very difficult to actually review the changes.
Can you fix this?

src/store/zap/actions.js
```diff
@@ -14,803 +14,820 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-import Vue from 'vue'
```
jingteng25742, line 17: The whole file seem to have changed and makes it very difficult to actually review the changes.
Can you fix this?

src/util/common-mixin.js
```diff
@@ -15,198 +15,215 @@
  *    limitations under the License.
  */
 
-import * as Util from './util'
```
jingteng25742, line 18: The whole file seem to have changed and makes it very difficult to actually review the changes.
Can you fix this?

jingteng25742 (changes requested):
Screen Shot 2022-08-26 at 11 26 58 AM

During the call, I believe the intention is to remove all the text and only leave icons.
The tooltips can include the texts.
For the expand / un-expand, let's use the default up / down arrow

e.g.
Screen Shot 2022-08-26 at 11 28 43 AM

src/util/common-mixin.js
Amirreza-Ghasemkhani: Fixed

src/store/zap/actions.js
Amirreza-Ghasemkhani: Fixed

src-electron/rest/user-data.js
Amirreza-Ghasemkhani: Fixed

src-electron/db/query-config.js
Amirreza-Ghasemkhani: Fixed

jingteng25742: > Screen Shot 2022-08-26 at 11 26 58 AM
> 
> During the call, I believe the agreed intention is to remove all the text (Edit/Copy/etc) and only leave icons. The tooltips can include the texts. For the expand / un-expand, let's use the default up / down arrow
> 
> e.g. Screen Shot 2022-08-26 at 11 28 43 AM

I am still the same issue for the UI.

src-electron/rest/user-data.js
```diff
@@ -697,6 +697,81 @@ function httpDeleteSessionPackage(db) {
...
+        let newEndpointTypeClusterId = await queryConfig.insertOrReplaceClusterState(db,newEndpointTypeId,endpointTypeCluster.clusterRef,endpointTypeCluster.side, endpointTypeCluster.enabled)
+        let oldAttributes = await queryAttribute.selectEndpointTypeAttributesByEndpointTypeRefAndClusterRef(db,oldEndpointTypeId,endpointTypeCluster.endpointTypeClusterId)
+        oldAttributes.forEach(async (attrubute) => {
+          await queryAttribute.duplicateEndpointTypeAttribute(db,
+            newEndpointTypeId,
+            newEndpointTypeClusterId,
+            attrubute)
+        })
+      })
+  } catch (err) {
```
jingteng25742, line 770: Instead of catching the error here, the exception should just bubble to httpPostDuplicateEndpointType() and the error should be printed  / responded there. otherwise httpPostDuplicateEndpointType() would just fail silently.

src/components/ZclEndpointCard.vue
```diff
@@ -394,8 +427,9 @@ export default {
   },
   created() {
     if (this.$serverGet != null) {
-      this.selectedAttributes= []
-      this.selectedReporting =[]
+      this.selectedservers = []
```
jingteng25742, line 430: typo selectedservers?

jingteng25742 (changes requested):

src/components/ZclEndpointCard.vue
Amirreza-Ghasemkhani: No, sometime it's break a part of code. after adding this line we don't have any issue

Amirreza-Ghasemkhani: Fixed

jingteng25742 (approved):

PR #703, opened 2022-09-01
Fixing the out of bounds dataview error when attribute is of size 3 b…

brdandu:
Fixing the out of bounds dataview error when attribute is of size 3 bytes

### DIFF
--- a/src-electron/util/types.js
+++ b/src-electron/util/types.js
@@ -116,7 +116,7 @@ function convertFloatToBigEndian(value, size) {
  * the given size. The value is returned in hex format and prefixed with '0x'.
  */
 function convertIntToBigEndian(value, size) {
-  const arrayBuffer = size <= 4 ? new ArrayBuffer(size) : new ArrayBuffer(8)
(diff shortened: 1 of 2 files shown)


src-electron/util/types.js
```diff
@@ -116,7 +116,12 @@ function convertFloatToBigEndian(value, size) {
  * the given size. The value is returned in hex format and prefixed with '0x'.
  */
 function convertIntToBigEndian(value, size) {
-  const arrayBuffer = size <= 4 ? new ArrayBuffer(size) : new ArrayBuffer(8)
+  const arrayBuffer =
+    size <= 4
+      ? size != 3
```
bzbarsky-apple, line 121: At some point something like:
```
  const arrayBuffer = new ArrayBuffer(Math.pow(2, Math.ceil(Math.log2(size))));
```
starts being more readable....

bzbarsky-apple (approved):

tecimovic (approved):

Issue #711, opened 2022-09-08
Keep Endpoint Open When Changing Page


Amirreza-Ghasemkhani:
When you going to enable attribute section or changing page to any other page of the application after back to the main page you will see all the endpoints that you extend are closed.

Amirreza-Ghasemkhani:
https://github.com/project-chip/zap/pull/712

PR #712, opened 2022-09-08
Keep Endpoint Open

Amirreza-Ghasemkhani:
https://github.com/project-chip/zap/issues/711

### DIFF
--- a/src/components/ZclEndpointCard.vue
+++ b/src/components/ZclEndpointCard.vue
@@ -69,8 +69,8 @@ limitations under the License.
             </q-tooltip>
           </q-btn>
           <q-btn
-            v-if="showAllInformationOfEndpoint"
-            @click.stop="toggleShowAllInformationOfEndpoint"
+            v-if="getEndpointInformation"
+            @click.stop="toggleShowAllInformationOfEndpoint(false)"
             flat
             dense
             icon="mdi-chevron-up"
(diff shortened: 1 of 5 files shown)


src/components/ZclEndpointCard.vue
```diff
@@ -48,8 +48,8 @@ limitations under the License.
             @click="modifyEndpointDialog = !modifyEndpointDialog"
           />
           <q-btn
-            v-if="showAllInformationOfEndpoint"
-            @click.stop="toggleShowAllInformationOfEndpoint"
+            v-if="getShowEndpointDataStatus"
```
brdandu, line 51: Change this to something like showEndpointInformation

brdandu (changes requested):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "type": "commonjs",
   "name": "zap",
-  "version": "0.0.0",
+  "version": "2022.9.13",
```
brdandu, line 4: I don't think this should change. @tecimovic

src/store/zap/mutations.js
```diff
@@ -543,4 +543,9 @@ export function updateIsProfileIdShown (state, value) {
 
 export function updateIsClusterOptionChanged(state, value) {
   state.isClusterOptionChanged = value
+}
+
```
brdandu, line 547: Start adding function description for all new functions that are added

brdandu (changes requested):

src/store/zap/mutations.js
```diff
@@ -541,6 +541,12 @@ export function updateIsProfileIdShown (state, value) {
...
 }
 
+// This function will update the cluster stage if cluster changed it will update the endpoint data
 export function updateIsClusterOptionChanged(state, value) {
   state.isClusterOptionChanged = value
+}
+
+export function toggleShowEndpoint(state, item) {
+  Vue.set(state.showEndpointData, item.id, item.value)
+  console.log(state.showEndpointData);
```
brdandu, line 551: Do we need the console.log statement here?

src/store/zap/mutations.js
brdandu: I meant toggleShowEndpoint which is the new function you added.

brdandu (changes requested):

brdandu (approved):

PR #717, opened 2022-09-09
Updating attribute and command queries to use packageIds such that custom extensions for standard clusters are handled appropriately

brdandu:
- When a cluster extension is added to standard clusters in the form of attributes or commands through custom xml then there is a need to account for packageIds
- For eg if a user adds a custom xml for the above use case and removes it the generation continues to generate for the added custom xml since it is not checking for removed packages.
- Github: ZAP#698


tecimovic (approved):
Generally, looks ok. If you ran regeneration, and you see nothing weird happens after you regen, I am ok with this.

PR #737, opened 2022-09-20
Matter Cleanup for Chip Tool Repo

brdandu:
Adding generic helpers with less hardcoding such that they can be used across different kinds of templates

- Adding selectStructsWithClusterAssociation which takes the logic src-electron/generator/matter/chip-tool/templates/helper.js#structs_with_cluster_name to the backend(sql query). structs_with_clusters helper uses this instead of exports.structs_with_cluster_name
- Updating cleanseLabelAsKebabCase so that it can be used instead of asDelimitedCommand
- Added as_underlying_zcl_type_size_bits which can be used to calculate the size of data types such as bitmaps, enums and numbers in bits or bytes. HOwever the implementation currently returns 0 for struct, stringg or anything else. May be this can be updated over time based on usage
- Added if_is_underlying_zcl_type_signed which can be used to find whether a data type is signed or not
- Similarly if_is_float_or_double_data_type is added to find out if a data type is float or double.
- endpoint type was being treated as signed integer due to parsing logic in zcl-silabs-loader.js. Added a separate condition for this
- Added generation tests for all the new helpers with an additional matter generation template
- Github: ZAP#679

This PR will be associated with a CHIP repo PR https://github.com/project-chip/connectedhomeip/pull/22762


src-electron/generator/helper-zcl.js
```diff
@@ -2456,6 +2456,160 @@ async function format_zcl_string_as_characters_for_generated_defaults(
...
+    )
+  } else {
+    number = await queryZcl.selectNumberById(this.global.db, type)
+  }
+
+  if (
+    number &&
+    (number.name.includes('float') ||
+      number.name.includes('double') ||
+      number.name.includes('single'))
```
bzbarsky-apple, line 2591: This can't end up being a struct/enum/bitmap type with a name that happens to include those strings?

src-electron/generator/matter/chip-tool/templates/helper.js
```diff
@@ -163,13 +163,27 @@ async function assertSameTestType(current, expected) {
...
+  'asDelimitedCommand has been deprecated. Use cleanse_label_as_kebab_case instead'
+);
+exports.asTypeMinValue = dep(
+  asTypeMinValue,
+  'asTypeMinValue has been deprecated. Use as_underlying_zcl_type_size_bits, if_is_underlying_zcl_type_signed and if_is_float_or_double_data_type instead'
+);
+exports.asTypeMaxValue = dep(
+  asTypeMaxValue,
+  'asTypeMaxValue has been deprecated. Use as_underlying_zcl_type_size_bits, if_is_underlying_zcl_type_signed and if_is_float_or_double_data_type instead'
+);
```
bzbarsky-apple, line 182: This forces a lot of complicated logic on consumers, as it turned out in the Matter-side PR....  We need a better solution here.

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1057,7 +1057,12 @@ async function processDataType(
 function prepareNumber(a, dataType) {
   return {
     size: a.$.size,
-    is_signed: a.$.name.endsWith('u') || !a.$.name.includes('int') ? 0 : 1,
+    is_signed:
+      a.$.name.endsWith('u') ||
+      !a.$.name.includes('int') ||
+      a.$.name.includes('endpoint')
```
bzbarsky-apple, line 1063: Why this hardcoding for "endpoint"?  What about other numeric types that have interesting names?  Why is "endpoint" special?  at bare minimum, this needs to be documented.

bzbarsky-apple (changes requested):
Er, meant to request changes.

vivien-apple:
@brdandu I see 4 different changes here and each of them could have been its own dedicated PR. That would make the process easier and not *all* changes will be blocked on *one* particular changes:
 1. Changes to the structs helper
 2. Changes to the asTypeMinValue/asTypeMaxValue helpers (for those I wonder if a partial on the Matter repo PR side would not make it easier to accept)
 3. Changes to asDelimitedCommand which I believe would easily be accepted if it was a single dedicated PR
 4. endpoint type handling

src-electron/zcl/zcl-loader-silabs.js
brdandu: got rid of the hardcoding

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1060,7 +1060,8 @@ async function processDataType(
 function prepareNumber(a, dataType) {
   return {
     size: a.$.size,
-    is_signed: a.$.name.endsWith('u') || !a.$.name.includes('int') ? 0 : 1,
+    is_signed:
+      a.$.name.endsWith('u') || !/int[0-9]{1,2}/g.test(a.$.name) ? 0 : 1,
```
bzbarsky-apple, line 1064: How about:
```suggestion
      /^int[0-9]{1,2}$/.test(a.$.name) ? 1 : 0,
```

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1060,7 +1060,8 @@ async function processDataType(
 function prepareNumber(a, dataType) {
   return {
     size: a.$.size,
-    is_signed: a.$.name.endsWith('u') || !a.$.name.includes('int') ? 0 : 1,
+    is_signed:
```
bzbarsky-apple, line 1063: How does this handle signed types like `temp-diff` or `temperature`, which are named integer types derived from signed types?

src-electron/generator/helper-zcl.js
```diff
@@ -2456,6 +2456,172 @@ async function format_zcl_string_as_characters_for_generated_defaults(
...
+      } else if (dataType && dataType.name == 'boolean') {
+        return `1`
+      } else {
+        return 'UINT' + dataTypesize * 8 + '_MAX'
+      }
+    } else {
+      return 'INT' + dataTypesize * 8 + '_MAX'
+    }
+  } else {
+    return dataTypesize * 8
```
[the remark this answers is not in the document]
brdandu: It felt useful to just give the size away as a default when no language is specified. They could use this size in their template in the future. May be I should mention size is in bits or even give an option if a user wants the size in bytes or bits

src-electron/generator/helper-zcl.js
```diff
@@ -2456,6 +2456,172 @@ async function format_zcl_string_as_characters_for_generated_defaults(
...
+      } else if (dataType && dataType.name == 'double') {
+        return `-std::numeric_limits<double>::infinity()`
+      } else {
+        return 0
+      }
+    } else {
+      return 'INT' + dataTypesize * 8 + '_MIN'
+    }
+  } else {
+    return dataTypesize * 8
```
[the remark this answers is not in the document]
brdandu: For min type value yes it makes sense to throw an error in this case.

src-electron/generator/matter/chip-tool/templates/helper.js
```diff
@@ -163,13 +163,27 @@ async function assertSameTestType(current, expected) {
...
 //
-exports.asDelimitedCommand = asDelimitedCommand;
-exports.asTypeMinValue = asTypeMinValue;
-exports.asTypeMaxValue = asTypeMaxValue;
-exports.structs_with_cluster_name = structs_with_cluster_name;
+exports.asDelimitedCommand = dep(
+  asDelimitedCommand,
+  'asDelimitedCommand has been deprecated. Use cleanse_label_as_kebab_case instead'
+);
+exports.asTypeMinValue = dep(
```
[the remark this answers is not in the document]
brdandu: We have gotten into a habit of deprecating already created helpers and not removing them because that maintains backwards compatibility. For eg: A customer has an old release of matter(with old helpers) and updates zap then things continue to work for them instead of breaking. However they will get these warnings which will let them know that they need to update their templates/helpers.

However I am starting to see a large accumulation of deprecated helpers so we may need a plan on how to deprecate old helpers.  May be remove all deprecated helpers after 2-3 releases of the stack but this can annoy customers. +@tecimovic

src-electron/zcl/zcl-loader-silabs.js
brdandu: > How about:

I have seen xml go with either `int8s` or `int8 for signed types

src-electron/generator/helper-zcl.js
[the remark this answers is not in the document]
bzbarsky-apple: I don't see how it can be useful.  I ask for the min value for int64u, I get back "64".  I ask for the max value, I also get "64".  How is that useful?

src-electron/zcl/zcl-loader-silabs.js
bzbarsky-apple: In that case `/^int[0-9]{1,2}s?$/.test(a.$.name) ? 1 : 0`.

src-electron/zcl/zcl-loader-silabs.js
brdandu: The loader first loads atomic types and then derived ones but we may not be doing that for numbers. Could you point me to an existing derived type for a number? I can test that. Any reason we have to define a type for temperature and not just use one of the existing number type for it?

bzbarsky-apple: @brdandu I'm not sure `temperature` and `temp-diff` are used in the Matter XML right now, but the Matter spec does define them in the Thermostat cluster.  So ideally the XML would use them, to match the spec.

The reason to use a named type is that the named type can define what values actually mean.

Ideally, it seems to me, `<type>` would have a signed/unsigned marker thing when `analog="true"`....  

Anyway derived types for number would be all of these, in Matter:
```
    <type id="0xE0" description="Time of day"              name="tod"               size="4"  analog="true"   />
    <type id="0xE1" description="Date"                     name="date"              size="4"  analog="true"   />
    <type id="0xE2" description="UTC Time"                 name="utc"               size="4"  analog="true"   />
    <type id="0xE3" description="Epoch Microseconds"       name="epoch_us"          size="8"  analog="true"   />
    <type id="0xE4" description="Epoch Seconds"            name="epoch_s"           size="4"  analog="true"   />
    <type id="0xE5" description="System Time Microseconds" name="systime_us"        size="8"  analog="true"   />
    <type id="0xE6" description="Percentage units 1%"      name="percent"           size="1"  analog="true"   />
    <type id="0xE7" description="Percentage units 0.01%"   name="percent100ths"     size="2"  analog="true"   />
    <type id="0xE8" description="Cluster ID"               name="cluster_id"        size="4"  discrete="true" />
    <type id="0xE9" description="Attribute ID"             name="attrib_id"         size="4"  discrete="true" />
    <type id="0xEA" description="Field ID"                 name="field_id"          size="4"  discrete="true" />
    <type id="0xEB" description="Event ID"                 name="event_id"          size="4"  discrete="true" />
    <type id="0xEC" description="Command ID"               name="command_id"        size="4"  discrete="true" />
    <type id="0xED" description="Action ID"                name="action_id"         size="1"  discrete="true" />
    <type id="0xEF" description="Transaction ID"           name="trans_id"          size="4"  discrete="true" />
    <type id="0xF0" description="Node ID"                  name="node_id"           size="8"  discrete="true" />
    <type id="0xF1" description="Vendor ID"                name="vendor_id"         size="2"  discrete="true" />
    <type id="0xF2" description="Device Type ID"           name="devtype_id"        size="4"  discrete="true" />
    <type id="0xF3" description="Fabric ID"                name="fabric_id"         size="8"  discrete="true" />
    <type id="0xF4" description="Group ID"                 name="group_id"          size="2"  discrete="true" />
    <type id="0xF5" description="Status Code"              name="status"            size="2"  discrete="true" />
    <type id="0xF6" description="Data Version"             name="data_ver"          size="4"  discrete="true" />
    <type id="0xF7" description="Event Number"             name="event_no"          size="8"  discrete="true" />
    <type id="0xF8" description="Endpoint Number"          name="endpoint_no"       size="2"  discrete="true" />
    <type id="0xF9" description="Fabric Index"             name="fabric_idx"        size="1"  discrete="true" />
```
plus as I said at least `temp-diff` and `temperature`, which as I said are not using named types right now, probably because those don't work for the signed case...

src-electron/generator/helper-zcl.js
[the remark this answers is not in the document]
brdandu: That is what I was trying to say. for max you get back 64 which can be used in future templates but min can throw an error.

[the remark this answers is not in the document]
bzbarsky-apple: Ah, I see.  But for an `int64s`, what does "max is 64" tell you that is actionable?  How would you use it?

src-electron/zcl/zcl-loader-silabs.js
brdandu: Hopefully that does not break something in the Zigbee pro stack. I can try it.

src-electron/generator/helper-zcl.js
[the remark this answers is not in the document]
brdandu: So in the template a user could do something like `int{{as_type_max_value type}}s` or `int{{as_type_max_value type}}` instead of `{{as_type_max_value type language='c++'}}` despite no specific language mentioned. The template can use that max size in different use cases that may arise.

src-electron/generator/helper-zcl.js
```diff
@@ -2456,6 +2456,172 @@ async function format_zcl_string_as_characters_for_generated_defaults(
   return formatted_string
 }
 
+/**
+ * Given a zcl device type returns its sign, size and zcl data type info stored
+ * in the database table.
+ * @param {*} type
+ * @param {*} context
+ * @returns returns sign, size and info of zcl device type
```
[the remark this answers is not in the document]
brdandu: I have changed the loader to account for float and single as exceptions. I also added logic such that for future signed types we can mentioned signed="true" in the xml and that will take precedence over the existing logic.
Added more documentation to clear things up.

src-electron/zcl/zcl-loader-silabs.js
brdandu: Added more logic to handle any future exceptions that can arise from types such as temperature, etc.
Just mentioned signed="true" in the xml and that will load them correctly.

src-electron/zcl/zcl-loader-silabs.js
brdandu: I have added the current exceptions into the loader. For future data types which make it into the types xml, I have updated the loader to account for `signed=true` in the xml. So when defining the signed variables just add that. That will take precedence and load the sign right for these data types

bzbarsky-apple (commented):
This generally looks pretty good, but I'm still not convinced on the max behavior.  See https://github.com/project-chip/zap/pull/737#discussion_r989213866

src-electron/generator/helper-zcl.js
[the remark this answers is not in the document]
brdandu: Well, it does not have to return the value. I can throw an error here too. We can have a separate helper for size if need be to avoid this confusion. This was more of a way to get size than max.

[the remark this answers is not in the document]
bzbarsky-apple: Separate helper for size would be _much_ clearer.

bzbarsky-apple (approved):

tecimovic (approved):

Issue #647, opened 2022-08-03
Cannot add custom xml nor Generate from the browser


brdandu:
Steps to reproduce

- Launch ZAP using rpm run zap
- In your browser go to http://localhost:9070/#/
- Click on zcl extensions and try adding a file. Clicking on '+' does nothing
- Also clicking on the Generate button does nothing

PR #741, opened 2022-09-23
Upload new ZCL extension on browser

pedramk:
- Add custom ZCL file extension from browser  (#647)


tecimovic:
Dealing with this differently.

Issue #723, opened 2022-09-14
count of enabled attribute has a problem


Amirreza-Ghasemkhani:
when you create an endpoint attribute count is okay but after refreshing the page it will change and it's not correct.

Amirreza-Ghasemkhani:
after disabling all the clusters attribute count show wrong number also after enabling cluster and disabling that attribute count show wrong number.

brdandu:
Need to get rid of counts. Moving this to do.

PR #742, opened 2022-09-23
Enabled Attributes Count Issue 

Amirreza-Ghasemkhani:
https://github.com/project-chip/zap/issues/723


brdandu (changes requested):
Add tests with multiple use cases of adding and removing endpoints

PR #757, opened 2022-10-17
Rename CLI executable to zap-cli and point apack.json to it

tecimovic:


jingteng25742:
Have we tested the CLI exe through all the CI already?

tecimovic:
I tested it locally and it worked ok. Our internal build engineers, are ok with merging this in, and then taking a new release and testing it. It's simpler than wrangling with changing the CI to point to the PR and so on.

And Matter, at this point, is unaffected by this at all, as it doesn't use apack.json.

jingteng25742 (dismissed):
zap-cli broken.

tecimovic:
The /snapshot should be fixed now, so we should be good to go.

jingteng25742:
Tested via Studio rel b1748. Sample app generation with Z3Light / Z3Switch are working.

PR #771, opened 2022-10-26
Fix the issues with the package selection page

tecimovic:
0. Properly shortcut the config page if there is only one of each package type.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -46,11 +46,15 @@ Cypress.Commands.add('rendererApi', (...args) => {
 })
 
 Cypress.Commands.add('setZclProperties', () => {
-  // These should happen if the selection page is shown
-  //cy.get('[data-test="gen-template"]').should('not.exist')
-  //cy.get('[data-test="login-submit"]').should('not.exist')
-
-  // These happen if selection page is not shown
-  cy.get('[data-test="gen-template"]').should('not.exist')
-  cy.get('[data-test="login-submit"]').should('not.exist')
+  cy.get('body').then(($body) => {
+    if ($body.find('[data-test="gen-template"]').length) {
+      // These should happen if the selection page is shown
+      cy.get('[data-test="gen-template"]').click()
+      cy.get('[data-test="login-submit"]').click()
+    } else {
+      // These happen if selection page is not shown
+      cy.get('[data-test="gen-template"]').should('not.exist')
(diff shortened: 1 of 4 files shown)


cypress/support/commands.js
```diff
@@ -48,6 +46,17 @@ Cypress.Commands.add('rendererApi', (...args) => {
...
 
 Cypress.Commands.add('setZclProperties', () => {
-  cy.get('[data-test="gen-template"]').click()
-  cy.get('[data-test="login-submit"]').click()
-})
\ No newline at end of file
+  cy.get('body').then(($body) => {
+    // The above doesn't work. It does find the buttons, even
+    // if they are hidden, obviously....
+    /*if ($body.find('[data-test="gen-template"]').length) {
```
brdandu, line 52: Do we need to update the tests to account for this behavior. With single package and without single package?

src/App.vue
```diff
@@ -113,8 +112,11 @@ export default {
     getAppData() {
       if (this.$serverGet != null) {
         this.$serverGet(restApi.uri.uiOptions).then((res) => {
-          this.$store.commit('zap/updateIsProfileIdShown', res.data.showProfileId)
-          console.log(res.data);
+          this.$store.commit(
+            'zap/updateIsProfileIdShown',
+            res.data.showProfileId
+          )
+          console.log(res.data)
```
brdandu, line 119: Get rid of console.log here?

src/App.vue
```diff
@@ -187,13 +189,13 @@ export default {
...
-        this.$router.push({path: '/login'})
+      if (val != true) {
+        this.$router.push({ path: '/login' })
       } else {
         this.$router.push({ path: '/' })
         this.getAppData()
       }
-    }
-  }
+    },
```
brdandu, line 198: Not sure why the additional , here comes from.

paulr34 (approved):

cypress/support/commands.js
tecimovic: We need to do something. tried the above and failed. Not sure what else to try. I'll have to study up cypress a bit more.

src/App.vue
tecimovic: yeah.

src/App.vue
tecimovic: Auto-formatter. I'll let it do its thing.

PR #786, opened 2022-11-07
Notification alert

paulr34:
When the user has notifications to check the "Notifications" button changes to red as previously discussed.


paulr34:
Remaking with a new branch

PR #787, opened 2022-11-07
adding code which changes the notification color when there are notif…

paulr34:
When the user has notifications to check the "Notifications" button changes to red as previously discussed. Also, removing an unused query.


paulr34:
accidentally rebased the wrong branch. going to make a new PR

Issue #793, opened 2022-11-09
Invent an "appendedSubdirectory=" field in templates.json and a --appendGenerationSubdirectory argument


tecimovic:
Problem at hand:
  - within the same SDK, zap is executed with a same command line and same `-o` argument.
  - some templates need to generate into a subdirectory.
  - other templates need to generate into a directory itself.

Solution:
   1.) Introduce a key "appendedSubdirectory: XYZ" localized per each genTemplates.json file.
   2.) Introduce a command-line argument `--appendGenerationSubdirectory`, which will make zap honor the data described under 1.
   3.) If the arg is present zap will not generate into DIR passed by -o, but will generate in a subdirectory of that DIR, specified in the specific genTemplates.json
  4.) In the CLI that executes this (apack.json), update the CLI to include the `--appendGenerationSubdirectory` arg.

The command-line argument is required to not break the existing APIs that simply generate into -o, and would like to simply ignore this new token of data inside the genTemplates.json.

PR #797, opened 2022-11-10
Add ability for generator options to carry an appended subdirectory.

tecimovic:
https://github.com/project-chip/zap/issues/793


selissia:
Example of what should be added to the templates.json to enable this:
"options": {
        "generator": {
            "appendDirectory": "zap-generated"
        }
    },

PR #806, opened 2022-11-13
Automatic loading of packages from zap files.

tecimovic:


src-electron/generator/generation-engine.js
```diff
@@ -527,16 +538,16 @@ async function loadSingleTemplate(db, genTemplatesJson) {
     return context
   }
   context.path = file
-  await dbApi.dbBeginTransaction(db)
+  if (!isTransactionAlreadyExisting) await dbApi.dbBeginTransaction(db)
```
brdandu, line 541: What happens if transaction is already active? Don't we have to queue a new transaction that could have come?

src-electron/importexport/import-isc.js
```diff
@@ -638,7 +638,7 @@ async function loadSessionKeyValues(db, sessionId, keyValues) {
  * @param {*} state
  * @param {*} sessionId
  */
-async function iscDataLoader(db, state, sessionId) {
+async function iscDataLoader(db, state, sessionId, packageMatch) {
```
brdandu, line 641: Is packageMatch being used here or can it be removed?

src-electron/importexport/import-json.js
```diff
@@ -57,10 +58,39 @@ function getPkgPath(pkg, zapFilePath) {
   }
 }
 
+// AutoLoad package. If succesful it returns an object:
+//  {
+//      packageId: <pkgId>
+//      packageType: <type>
+// }
```
brdandu, line 65: We may want to be consistent with our comments under
/* */

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -2257,11 +2204,14 @@ async function loadSilabsZcl(db, metafile, isJson = false) {
...
     db: db,
   }
+  let isTransactionAlreadyExisting = dbApi.isTransactionActive()
   env.logDebug(`Loading Silabs zcl file: ${metafile}`)
   if (!fs.existsSync(metafile)) {
     throw new Error(`Can't locate: ${metafile}`)
   }
-  await dbApi.dbBeginTransaction(db)
+
+  if (!isTransactionAlreadyExisting) await dbApi.dbBeginTransaction(db)
```
brdandu, line 2213: I guess we somehow keep pushing the transactions by 10s if one is going on. I am imagining a bunch of transactions just being pushed by 10s all the time and if that can also disrupt the order of things. for eg:
- transaction 1 is on going
- transaction 2 tries to execute but transaction 1 is ongoing so gets pushed by 10s
- transaction 3 tries to execute but transaction 1 is ongoing so gets pushed by 10s
- transaction 2 tries to execute again but transaction 1 is ongoing so gets pushed by another 10s
- transaction 1 finishes execution
- transaction 3 executes
- transaction 2 executes

In the above example now T3 happened before T2 even though the order of things was different. I am wondering if we should queue them to maintain the order of things instead of bumping them 10s every time.

brdandu: This is an issue if one transaction depends on another transaction to finish before it.

test/server-bare.test.js
```diff
@@ -209,21 +209,17 @@ describe('Session specific tests', () => {
   // We save and then load, which creates a new session.
```
jingteng25742, line 209: Can we add unit tests that covers the following scenario?
It'll ensure all this logic from silently breaking for customers.

1) Zigbee + Matter templates are pre-loaded.
    * verify Zigbee config loads correctly
    * verify Matter config loads correctly
    * if no matching packages are found, what should happen here? 
2) Only Zigbee template is pre-loaded
    * verify Matter config loads correctly
3) No templates are pre-loaded
    * verify Zigbee config loads correctly
    * verify Matter config loads correctly

src-electron/importexport/import-isc.js
tecimovic: Can be removed. I added it, because the import.js will pass it to this function, as it's using a common interface to both import-isc.js and import-json.js. It wasn't clear to me which is better:
   1.) Declare an argument that IS passed and then not use it.
   2.) Pretend that there is no argument, even though the caller WILL add the argument.

src-electron/zcl/zcl-loader-silabs.js
tecimovic: Our transaction handling leaves a lot to be desired. The fact that SQLITE can run only one transaction at the time is a problem. What helps us, is that we need transactions ONLY when we're writing large bulk data into the database, which happens only in 3 occasions:
   1.) When loading a .zap file.
   2.) When loading a ZCL properties.
   3.) When loading a gen-templates.json

These used to be separated, but with this work, the case 1, can also automatically "include" cases 2 and 3.

The 10 second logic is a hopeful hack, essentially built as:
   1.) You load a.zap.
   2.) At the same time you load b.zap.
   
Loading of a file should be done in 10 seconds, so b.zap can wait until a.zap completes and closes transaction.

I can see tons of issues in this whole scenario, but honestly, I don't know what to do exactly about it. Any ideas appreciated.

test/server-bare.test.js
tecimovic: We should. I will log an issue.

tecimovic: Created: https://github.com/project-chip/zap/issues/817

src-electron/importexport/import-json.js
tecimovic: Good point. I pushed another commit that cleans up this comment.

src-electron/importexport/import-isc.js
brdandu: Can the interface not have this?

src-electron/zcl/zcl-loader-silabs.js
brdandu: What if we change `isTransactionAlreadyExisting` to an array such as `AllIncomingTransactions = []` which can be an array of promises. We take them based on a queue. Execute them in the order they came in and remove them from the queue. I feel having an order for these transactions will make it easier for us to debug things if we ever run into issues and will probably help us avoid complex issues.

tecimovic: Valid point. You should log an issue and we can implement that in a separate PR.

src-electron/importexport/import-isc.js
tecimovic: I'll just remove it from here...

src-electron/zcl/zcl-loader-silabs.js
tecimovic: Logged: https://github.com/project-chip/zap/issues/819
Assigned it to you.

PR #821, opened 2022-11-16
Providing a mechanism in Overridable to prevent additional comments

brdandu:
- adding no_warning option in overridable.atomicType so that /* TYPE WARNING: */ message does not get generated because if this helper is used within a comments section of the code then compiler will throw an error
- Bumping the feature level
- JIRA: ZAPP-991


tecimovic:
makes sense.....

PR #835, opened 2022-11-28
Add a CI step that runs codegen on matter SDK

andy31415:
This validates on every pull request if the changes will have sideffects in codegen in matter.

Currently this step takes 8 minutes on my machine with act, expect similar timing in github CI, may drop once we enable parallel codegen in the matter repo.

There are a few things hardcoded:
  - checkout master from chip sdk. Seems reasonable since we want to see if we are affecting matter as of today
  - fixes the chip-build docker image. I needed this because of clang-format compatibility. It is likely that we do not need to update this frequently.


jingteng25742:
@andy31415 
Can you move the "matter-sdk-codegen" job into "release.yml" and have it depend on (needs:) "build-zap"? 
That way, you can reduce the runtime by reusing the prebuilt zap via "download-artifact" instead of rebuilding ZAP all over again.

andy31415:
> @andy31415 Can you move the "matter-sdk-codegen" job into "release.yml" and have it depend on (needs:) "build-zap"? That way, you can reduce the runtime by reusing the prebuilt zap via "download-artifact" instead of rebuilding ZAP all over again. It can also be leveraged to gatekeep from making a release in case regen fails.

Updated code. Not sure how to test this one, so made changes that look ok, but expect I will have to iterate a bit (even if it works, I need to remove the debug `ls -R`

andy31415:
> Updated code. Not sure how to test this one, so made changes that look ok, but expect I will have to iterate a bit (even if it works, I need to remove the debug `ls -R`

Update complete, it seems to finish in around 9 minutes currently (matter did not yet enable parallelism)

tecimovic (approved):
Looks good to me. The more testing the better.

PR #840, opened 2022-11-29
Add logic that deals with unmatched package version.

tecimovic:


apack.json
```diff
@@ -4,7 +4,7 @@
   "description": "Graphical configuration tool for application and libraries based on Zigbee Cluster Library.",
   "path": [".", "node_modules/.bin/", "ZAP.app/Contents/MacOS"],
   "requiredFeatureLevel": "apack.core:9",
-  "featureLevel": 87,
+  "featureLevel": 88,
```
brdandu, line 7: Does our sdk need a feature level bump too?

tecimovic: No.

tecimovic:
Merging this even the "matter regen" is broken because of an ongoing conversation how to deal with that. It's not a blocker.

PR #857, opened 2022-12-14
Add backend for Darwin availability annotations.

bzbarsky-apple:


src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -273,6 +306,340 @@ function hasArguments() {
...
+    if (st && st.isFabricScoped) {
+      options.hash.deprecationMessage =
+        options.hash.fabricScopedDeprecationMessage;
+    } else {
+      options.hash.deprecationMessage =
+        options.hash.nonFabricScopedDeprecationMessage;
+    }
+  }
+
+  // This assumes that an item is introduced in some single release (which
```
ksperling-apple, line 489: Is there a straightforward way to catch if this is violated and throw an error?

bzbarsky-apple: `findReleaseForPath` could keep looking instead of returning once if finds something, and throw if it finds a second thing.

bzbarsky-apple: `findReleaseForPath` could keep looking instead of returning once if finds something, and throw if it finds a second thing.

bzbarsky-apple: Made that change.

tecimovic:
@bzbarsky-apple : what does this do? Can you describe in a short paragraph what is the purpose of this "availability" thing?

bzbarsky-apple:
@tecimovic Darwin APIs need to be annotated with macros that indicate which iOS, macOS, etc releases the APIs shipped in.  These take two forms `API_AVAILABLE(ios(version), macos(version), ....)` and `API_DEPRECATED("message", ios(introducedVersion, deprecatedVersion), ...)`.  That way, if you are compiling against headers from version X but want your application to target an earlier version Y you can get your compiler to tell you when you use APIs not available in version Y.

This is putting in place a mechanism whereby:

1. There is a YAML file that lists releases and for each release lists which things were introduced and deprecated in that release.
2. In that YAML file we can associate version numbers (for ios, etc) to a release.
3. We can then auto-generate these annotations for the APIs that we code-generate.

See https://github.com/project-chip/connectedhomeip/pull/24071 for the YAML file and template changes to make use of these new helpers.

bzbarsky-apple:
@tecimovic See also https://developer.apple.com/documentation/swift/marking-api-availability-in-objective-c

woody-apple (approved):

tecimovic (approved):

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -273,6 +306,344 @@ function hasArguments() {
   return !!this.arguments.length;
 }
 
+let availabilityData;
+function fetchAvailabilityData(global) {
+  if (!availabilityData) {
+    let f = path.join(path.dirname(global.templatePath), 'availability.yaml');
```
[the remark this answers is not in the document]
tecimovic: We can deal with this later though. There is a lot of general cleanup in the queue anyway...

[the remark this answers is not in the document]
bzbarsky-apple: Added it to "resources" in the template.json.

tecimovic (approved):

PR #863, opened 2022-12-16
Update the main page

tbrkollar:


src/components/InitialContent.vue
```diff
@@ -16,57 +16,128 @@ limitations under the License.
...
-<li><a href="docs/coding-standard.md">Coding standard</a></li>
-<li><a href="docs/api.md">API</a></li>
-<li><a href="docs/releasenotes.md">Release notes</a></li>
-<li><a href="docs/faq.md">FAQ</a></li>
-</ul>
-</div>
+    <div v-if="uiTheme === 'matter'">
+      <h3 id="zcl-advanced-platform">Welcome to Matter</h3>
+    </div>
+    <div v-else-if="uiTheme === 'zigbee'">
```
tecimovic, line 22: I'm assuming this is temporary, right? We can merge this for now, but we will have to weed this out, we can't hard-code "zigbee" or "matter", they need to be data driven.

The idea is:
   1.) When loading a zcl.json file, there will be a possible key there: "content=something.html" or something like that, and then that HTML is loaded from a file that comes from SDK.
   2.) If the loaded zcl.json does not have a 'content' key, then there will be a placeholder content right in this file.

src/components/InitialContent.vue
```diff
@@ -16,57 +16,128 @@ limitations under the License.
...
+      <h3 id="zcl-advanced-platform">Welcome to Zigbee</h3>
+    </div>
+    <div v-else>
```
tecimovic, line 25: Ok, cool. This is placeholder content, I guess.

I'll update this myself once we merge this in....

tecimovic (approved):
I'm ok with merging this in. Merge it, then I'll open a follow-up issue to get the content properly from the SDK files....

PR #868, opened 2022-12-19
Add support for removing APIs in Darwin codegen.

bzbarsky-apple:
* Makes the compat*NameRemapping functions just rely entirely on the availability rename data instead of doing non-acronym-preserving upcasing.
* Adds compatCommandNameRemapping to handle renamed commands.
* Adds a wasRemoved helper.
* Adds helpers to be able to do boolean logic in "if/unless".
* Adds helpers to be able to do boolean conditions on async values.

### DIFF
--- a/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
+++ b/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
@@ -137,7 +137,7 @@ function asObjectiveCNumberType(label, type, asLowerCased) {
 
 function compatClusterNameRemapping(cluster, options) {
   cluster = appHelper.asUpperCamelCase(cluster, {
-    hash: { preserveAcronyms: false },
+    hash: { preserveAcronyms: true },
   });
 
   const old = oldName.call(this, cluster, options);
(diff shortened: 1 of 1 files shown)


vivien-apple (approved):
Thanks for adding "and/or/not and async_and/async_or/async_not" !

bzbarsky-apple:
> Thanks for adding "and/or/not and async_and/async_or/async_not" !

You're welcome!  It made _my_ life much simpler when writing the templates!

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -711,6 +734,90 @@ function wasIntroducedBeforeRelease(releaseName, clusterName, options) {
...
+      );
+    }
+    return running || current;
+  }, false);
+}
+
+function not(value) {
+  if (value instanceof Promise) {
+    throw new Error(
+      "Promise passed as argument to 'not'.  You probably want to use async_if/async_not"
```
tecimovic, line 778: Could you not simply have one `not` and then it detects if it's a promise and does what `async_not` does?

Is there a value of explicitely stating whether you need a promise or not, compared to dynamically reacting on either a promise or not a promise being passed in?

tecimovic (approved):
Left a comment. Not a merge-blocker....

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
bzbarsky-apple: I could have done that, but:

1. Then I could not use `async function`, because that will _always_ return a promise.  I would have had to do manual promise chaining and either return a promise or a boolean.
2. Use of `not` with a promise in an `if` would silently do the wrong thing.

So I went with something being a little more explicit: if you see `if` being mixed with `async_not` you know you are doing it wrong, and if you see `if` with `not` then any failures due to promises sneaking in will be not be silent.  And the code is cleaner due to being able to use `async function` for the async version.

PR #888, opened 2023-01-14
Fix handling of manufacturerCode in various places.

bzbarsky-apple:
There were two bugs:

1) asMEI produced incorrect output if the manufacturer code was high enough that
   the high bit in it was set, due to ending up with negative numbers.
2) chip_server_clusters for some reason used mfgCode, not manufacturerCode
   (which almost everything else uses).

### DIFF
--- a/src-electron/generator/helper-endpointconfig.js
+++ b/src-electron/generator/helper-endpointconfig.js
@@ -465,7 +465,10 @@ function endpoint_attribute_long_defaults(options) {
 }
 
 function asMEI(manufacturerCode, code) {
-  return '0x' + bin.int32ToHex((manufacturerCode << 16) + code)
+  // Left-shift (and for that matter bitwise or) produces a _signed_ 32-bit
+  // number, which will probably be negative.  Force it to unsigned 32-bit using
+  // >>> 0.
+  return '0x' + bin.int32ToHex(((manufacturerCode << 16) | code) >>> 0);
 }
 
 // The representation of null depends on the type, so we can't use a single
--- a/src-electron/generator/matter/app/zap-templates/templates/app/helper.js
+++ b/src-electron/generator/matter/app/zap-templates/templates/app/helper.js
@@ -469,7 +469,10 @@ function chip_friendly_endpoint_type_name(options) {
 }
 
 function asMEI(prefix, suffix) {
-  return cHelper.asHex((prefix << 16) + suffix, 8);
+  // Left-shift (and for that matter bitwise or) produces a _signed_ 32-bit
+  // number, which will probably be negative.  Force it to unsigned 32-bit using
+  // >>> 0.
+  return cHelper.asHex(((prefix << 16) | suffix) >>> 0, 8);
 }
 
 // Not to be exported.
--- a/src-electron/db/query-endpoint-type.js
+++ b/src-electron/db/query-endpoint-type.js
@@ -172,6 +172,7 @@ async function selectAllClustersDetailsFromEndpointTypes(db, endpointTypes) {
       description: x.DESCRIPTION,
       define: x.DEFINE,
       mfgCode: x.MANUFACTURER_CODE,
+      manufacturerCode: x.MANUFACTURER_CODE,
       side: x.SIDE,
       enabled: x.ENABLED,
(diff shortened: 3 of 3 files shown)


src-electron/generator/helper-endpointconfig.js
```diff
@@ -465,7 +465,10 @@ function endpoint_attribute_long_defaults(options) {
 }
 
 function asMEI(manufacturerCode, code) {
-  return '0x' + bin.int32ToHex((manufacturerCode << 16) + code)
+  // Left-shift (and for that matter bitwise or) produces a _signed_ 32-bit
+  // number, which will probably be negative.  Force it to unsigned 32-bit using
+  // >>> 0.
+  return '0x' + bin.int32ToHex(((manufacturerCode << 16) | code) >>> 0);
```
tcarmelveilleux, line 471: Shouldn't the >>>0 be applied to manufacturerCode (or the << 16 expression), or is it OK to do after the full expression (i.e. start interpreting 2's complement value as unsigned)? Not super familiar with that corner of Javascript, but this is a bit odd.

Perhaps a less tricky alternative is:

```
  return '0x' + bin.int32ToHex(manufacturerCode) + bin.int32ToHex(code).padStart(4, '0');
```

Functions like asHex(val, numBitsWidth) and asZeroPaddedHex(val, numBitsWidth) would probably hex with composing more of these if needed.

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
```diff
@@ -469,7 +469,10 @@ function chip_friendly_endpoint_type_name(options) {
 }
 
 function asMEI(prefix, suffix) {
-  return cHelper.asHex((prefix << 16) + suffix, 8);
+  // Left-shift (and for that matter bitwise or) produces a _signed_ 32-bit
+  // number, which will probably be negative.  Force it to unsigned 32-bit using
+  // >>> 0.
+  return cHelper.asHex(((prefix << 16) | suffix) >>> 0, 8);
```
tcarmelveilleux, line 475: Is this basically a version of the solution I suggested above re: padding? If so, then it may be a possible to use concatenation of those asHex rather than int32toHex.

tcarmelveilleux (approved):

src-electron/generator/helper-endpointconfig.js
bzbarsky-apple: Both the `<<` operator and the `|` operator produce signed output in JavaScript.  See https://262.ecma-international.org/6.0/#sec-left-shift-operator-runtime-semantics-evaluation step 12 and https://262.ecma-international.org/6.0/#sec-binary-bitwise-operators-runtime-semantics-evaluation steps 7 and 11.

So `(a << b)>>>0 | c` would shift by `b`, then return an unsigned integer, which would get converted to a signed integer, and the bitwise or would happen on signed integers....

So the `>>>` really does need to be on the final result.  I totally agree the bitwise ops in JS are _very_ odd.

> Perhaps a less tricky alternative is:

This works only if `code` does not already included the `manufacturerCode` in its high bits.  Which is a property the old approach had as well, and might be OK, since including the manufacturer code in the high bits of `code` if you also set `manufacturerCode` is arguably pilot error...

So maybe I should just not worry about handling that case "gracefully" and just fix the actual correctness bug around signed values.

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
bzbarsky-apple: This is the same code as above, but using `cHelper.asHex` instead of `bin.int32ToHex`, mostly because that's what the existing code at the two callsites was doing.  I don't know why we have those two separate conversion utilities....

jingteng25742 (approved):

src-electron/generator/helper-endpointconfig.js
bzbarsky-apple: >   `return '0x' + bin.int32ToHex(manufacturerCode) + bin.int32ToHex(code).padStart(4, '0');`

No, this does not do the right thing at all: it will create a string of length 2 + 8 + 8 = 18, whereas we want a string of length at most 10 here.

Since code might in fact be a full 32-bit value, we _have_ to combine `manufacturerCode` with `code` as integers, not as strings.

PR #899, opened 2023-01-19
Adding to zcl_attributes_server and enabled_attributes_for_cluster_and_side helpers to replace chip_server_cluster_attributes

brdandu:
- chip_server_cluster_attributes is returning server attributes which are not actually enabled because the server side cluster is disabled.
- The reason chip_server_cluster_attributes is picking the server side attributes which are not enabled is because those are enabled in the .zap file(Which is the case for saving user selections and ease of use). However the helper is not actually checking if the cluster is enabled as well.
- all_user_cluster_attributes_irrespective_of_manufacturing_specification is solving this issue correctly and actually only showing attributes which are truly enabled.
- Github: ZAP#898

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -40,11 +40,11 @@
 <dt><a href="#module_DB API_ device type database access">DB API: device type database access</a></dt>
 <dd><p>This module provides queries for device types.</p>
 </dd>
-<dt><a href="#module_DB API_ endpoint configuration queries against the database.">DB API: endpoint configuration queries against the database.</a></dt>
-<dd><p>This module provides queries for endpoint configuration.</p>
-</dd>
 <dt><a href="#module_DB API_ endpoint type queries against the database.">DB API: endpoint type queries against the database.</a></dt>
 <dd><p>This module provides queries for endpoint type.</p>
+</dd>
+<dt><a href="#module_DB API_ endpoint configuration queries against the database.">DB API: endpoint configuration queries against the database.</a></dt>
+<dd><p>This module provides queries for endpoint configuration.</p>
 </dd>
 <dt><a href="#module_DB API_ zcl database access">DB API: zcl database access</a></dt>
 <dd><p>This module provides queries for enums.</p>
@@ -64,13 +64,13 @@
 <dt><a href="#module_DB API_ package-based queries.">DB API: package-based queries.</a></dt>
 <dd><p>This module provides queries related to packages.</p>
 </dd>
-<dt><a href="#module_DB API_ session related queries.">DB API: session related queries.</a></dt>
-<dd><p>This module provides session related queries.</p>
-</dd>
 <dt><a href="#module_DB API_ zcl database access">DB API: zcl database access</a></dt>
 <dd><p>This module provides queries for ZCL static entities
 inside a single session. Things like:
    all visible clusters, etc.</p>
+</dd>
+<dt><a href="#module_DB API_ session related queries.">DB API: session related queries.</a></dt>
+<dd><p>This module provides session related queries.</p>
 </dd>
 <dt><a href="#module_DB API_ zcl database access">DB API: zcl database access</a></dt>
 <dd><p>This module provides queries for enums.</p>
@@ -122,6 +122,9 @@ inside a single session. Things like:
 </dd>
 <dt><a href="#module_REST API_ generation functions">REST API: generation functions</a></dt>
 <dd><p>This module provides the REST API to the IDE component handling.</p>
+</dd>
+<dt><a href="#module_REST API_ initialization functions">REST API: initialization functions</a></dt>
+<dd><p>This module provides the REST API to the session initialization</p>
 </dd>
 <dt><a href="#module_REST API_ static zcl functions">REST API: static zcl functions</a></dt>
 <dd><p>This module provides the REST API to the static zcl queries.</p>
@@ -530,6 +533,10 @@ with its actual type from disciminator table.</p>
 <dd><p>Gathers the data type information of an entry based on data type name along
 with its actual type from disciminator table.</p>
 </dd>
+<dt><a href="#selectDataTypeByNameAndClusterId">selectDataTypeByNameAndClusterId(db, name, clusterId, packageIds)</a> ⇒</dt>
+<dd><p>Gathers the data type information based on data type name and
+clusterId along with its actual type from disciminator table.</p>
+</dd>
 <dt><a href="#selectAllDataTypes">selectAllDataTypes(db, packageId)</a> ⇒</dt>
 <dd><p>Gathers All the data types</p>
 </dd>
@@ -540,6 +547,9 @@ if string.</p>
 </dd>
 <dt><a href="#selectNumberByName">selectNumberByName(db, name, packageIds)</a> ⇒</dt>
 <dd><p>Select an number matched by name.</p>
+</dd>
+<dt><a href="#selectNumberByNameAndClusterId">selectNumberByNameAndClusterId(db, name, packageIds)</a> ⇒</dt>
+<dd><p>Select a number matched by name and clusterId</p>
 </dd>
 <dt><a href="#selectNumberById">selectNumberById(db, name)</a> ⇒</dt>
 <dd><p>Select an number matched by id.</p>
@@ -660,6 +670,11 @@ endpoint type array, for the appropriate endpoint.</p>
 <dd><p>Generates array of { index , mfgCode } pairs, matching
 the indexes in attribute table.</p>
 </dd>
+<dt><a href="#endpoint_reporting_config_defaults">endpoint_reporting_config_defaults(options)</a></dt>
+<dd><p>This helper supports an &quot;order&quot; CSV string, such as:
+  &quot;direction,endpoint,clusterId,attributeId,mask,mfgCode,minmax&quot;
+The string above is a default value, and it determines in what order are the fields generated.</p>
+</dd>
 <dt><a href="#collectAttributes">collectAttributes()</a></dt>
 <dd><p>Attribute collection works like this:
    1.) Go over all the clusters that exist.
@@ -713,6 +728,33 @@ Use it as: {{future name=&quot;NAME&quot;}}</p>
 <dd><p>The token template assigns an unique ID to each unique attribute. These IDs
   span all attributes from all clusters from all endpointTypes. This helper
   function allows the template to increment the token ID within the tokens context.</p>
+</dd>
+<dt><a href="#token_attribute_util">token_attribute_util(context, options)</a> ⇒</dt>
+<dd><p>Util function that extracts all the token attribute information.</p>
+</dd>
+<dt><a href="#token_attributes">token_attributes(endpointTypeRef, options)</a> ⇒</dt>
+<dd><p>Get information about all the token attributes in the configuration or this
+helper can be used within an endpoint block helper to fetch the
+corresponding token attributes based on endpoint type given.
+Available Options:
+isSingleton: 0/1, option can be used to filter attributes based on singleton
+or non-singleton(Available with endpointTypeRef only)</p>
+</dd>
+<dt><a href="#token_attribute_clusters">token_attribute_clusters(endpointTypeRef, options)</a> ⇒</dt>
+<dd><p>This helper can return all token associated clusters across endpoints or
+this helper can be used within an endpoint block helper to fetch the
+corresponding token associated clusters.
+Available Options:
+isSingleton: 0/1, option can be used to filter clusters based on singleton
+or non-singleton attributes.</p>
+</dd>
+<dt><a href="#token_attribute_endpoints">token_attribute_endpoints(options)</a> ⇒</dt>
+<dd><p>Get all endpoints which have token attributes in the configuration.
+AvailableOptions:</p>
+<ul>
+<li>isSingleton: 0/1, option can be used to filter endpoints based on singleton
+or non-singleton.</li>
+</ul>
 </dd>
 <dt><a href="#get_cli_size">get_cli_size(size, type, allowZclTypes)</a> ⇒</dt>
 <dd></dd>
@@ -781,13 +823,6 @@ queries that are needed to load the attribute state</p>
 <dd><p>Function that actually loads the data out of a state object.
 Session at this point is blank, and has no packages.</p>
 </dd>
-<dt><a href="#readDataFromFile">readDataFromFile(filePath)</a> ⇒</dt>
-<dd><p>Reads the data from the file and resolves with the state object if all is good.</p>
-</dd>
-<dt><a href="#importDataFromFile">importDataFromFile(db, filePath)</a> ⇒</dt>
-<dd><p>Writes the data from the file into a new session.
-NOTE: This function does NOT initialize session packages.</p>
-</dd>
 <dt><a href="#importSessionKeyValues">importSessionKeyValues(db, sessionId, keyValuePairs)</a></dt>
 <dd><p>Resolves with a promise that imports session key values.</p>
 </dd>
@@ -805,6 +840,13 @@ with the succesfull writing into the database.</p>
 </dd>
 <dt><a href="#readJsonData">readJsonData(filePath, data)</a> ⇒</dt>
 <dd><p>Parses JSON file and creates a state object out of it, which is passed further down the chain.</p>
+</dd>
+<dt><a href="#readDataFromFile">readDataFromFile(filePath)</a> ⇒</dt>
+<dd><p>Reads the data from the file and resolves with the state object if all is good.</p>
+</dd>
+<dt><a href="#importDataFromFile">importDataFromFile(db, filePath)</a> ⇒</dt>
+<dd><p>Writes the data from the file into a new session.
+NOTE: This function does NOT initialize session packages.</p>
 </dd>
 <dt><a href="#initSessionTimers">initSessionTimers()</a></dt>
 <dd><p>Start session specific validation.</p>
@@ -850,15 +892,6 @@ with the succesfull writing into the database.</p>
 </dd>
 <dt><a href="#startUpMainInstance">startUpMainInstance(quitFunction, argv)</a></dt>
 <dd><p>Default startup method.</p>
-</dd>
-<dt><a href="#packagesAndSessions">packagesAndSessions(db)</a> ⇒</dt>
-<dd><p>This function returns Properties, Templates and Dirty-Sessions</p>
-</dd>
-<dt><a href="#initializeSession">initializeSession(db, options:)</a> ⇒</dt>
-<dd><p>This function creates a new session with its packages according to selected Properties and Templates</p>
-</dd>
-<dt><a href="#loadPreviousSessions">loadPreviousSessions(db)</a> ⇒</dt>
-<dd><p>This function reloads previous session by user selected session&#39;s id</p>
 </dd>
 <dt><a href="#doOpen">doOpen(menuItem, browserWindow, event)</a></dt>
 <dd><p>Perform a file-&gt;open operation.</p>
@@ -1005,31 +1038,6 @@ that will be resolved when all the XML files are done, or rejected if at least o
 <dd><p>Toplevel function that loads the xml library file
 and orchestrates the promise chain.</p>
 </dd>
-<dt><a href="#recordToplevelPackage">recordToplevelPackage(db, metadataFile, crc)</a> ⇒</dt>
-<dd><p>Records the toplevel package information and resolves into packageId</p>
-</dd>
-<dt><a href="#recordVersion">recordVersion(db, ctx)</a></dt>
-<dd><p>Records the version into the database.</p>
-</dd>
-<dt><a href="#loadZclMetafiles">loadZclMetafiles(db, metadataFile)</a> ⇒</dt>
-<dd><p>Toplevel function that loads the zcl file and passes it off to the correct zcl loader.</p>
-</dd>
-<dt><a href="#loadZcl">loadZcl(db, metadataFile)</a> ⇒</dt>
-<dd><p>Loads individual zcl.json metafile.</p>
-</dd>
-<dt><a href="#loadIndividualFile">loadIndividualFile(db, filePath, sessionId)</a></dt>
-<dd><p>Load individual custom XML files.</p>
-</dd>
-<dt><a href="#qualifyZclFile">qualifyZclFile(db, info, parentPackageId)</a> ⇒</dt>
-<dd><p>Promises to qualify whether zcl file needs to be reloaded.
-If yes, the it will resolve with {filePath, data, packageId}
-If not, then it will resolve with {error}</p>
-</dd>
-<dt><a href="#processZclPostLoading">processZclPostLoading(db)</a> ⇒</dt>
-<dd><p>Promises to perform a post loading step.</p>
-</dd>
-<dt><a href="#getDiscriminatorMap">getDiscriminatorMap(db, packageIds)</a> ⇒</dt>
-<dd></dd>
 <dt><a href="#collectDataFromJsonFile">collectDataFromJsonFile(ctx)</a> ⇒</dt>
 <dd><p>Promises to read the JSON file and resolve all the data.</p>
 </dd>
@@ -1200,6 +1208,31 @@ e.g. for ClusterExtensions.</p>
 <dd><p>Toplevel function that loads the toplevel metafile
 and orchestrates the promise chain.</p>
 </dd>
+<dt><a href="#recordToplevelPackage">recordToplevelPackage(db, metadataFile, crc)</a> ⇒</dt>
+<dd><p>Records the toplevel package information and resolves into packageId</p>
+</dd>
+<dt><a href="#recordVersion">recordVersion(db, ctx)</a></dt>
+<dd><p>Records the version into the database.</p>
+</dd>
+<dt><a href="#loadZclMetafiles">loadZclMetafiles(db, metadataFile)</a> ⇒</dt>
+<dd><p>Toplevel function that loads the zcl file and passes it off to the correct zcl loader.</p>
+</dd>
+<dt><a href="#loadZcl">loadZcl(db, metadataFile)</a> ⇒</dt>
+<dd><p>Loads individual zcl.json metafile.</p>
+</dd>
+<dt><a href="#loadIndividualFile">loadIndividualFile(db, filePath, sessionId)</a></dt>
+<dd><p>Load individual custom XML files.</p>
+</dd>
+<dt><a href="#qualifyZclFile">qualifyZclFile(db, info, parentPackageId)</a> ⇒</dt>
+<dd><p>Promises to qualify whether zcl file needs to be reloaded.
+If yes, the it will resolve with {filePath, data, packageId}
+If not, then it will resolve with {error}</p>
+</dd>
+<dt><a href="#processZclPostLoading">processZclPostLoading(db)</a> ⇒</dt>
+<dd><p>Promises to perform a post loading step.</p>
+</dd>
+<dt><a href="#getDiscriminatorMap">getDiscriminatorMap(db, packageIds)</a> ⇒</dt>
+<dd></dd>
 </dl>
 
 <a name="module_DB API_ External URLs."></a>
@@ -1571,22 +1604,29 @@ This module provides cache for commonly used static database queries.
 
 - [DB API: zcl database access](#module*DB API* zcl database access)
   - [~clear()](#module*DB API* zcl database access..clear)
-  - [~put(key, packageId, data)](#module*DB API* zcl database access..put)
+  - [~put(key, packageId, data)](#module*DB API* zcl database access..put) ⇒
   - [~get(key, packageId)](#module*DB API* zcl database access..get) ⇒
   - [~isCached(key, packageId)](#module*DB API* zcl database access..isCached) ⇒
+  - [~cacheQuery(key, packageId)](#module*DB API* zcl database access..cacheQuery) ⇒
+  - [~cacheStats()](#module*DB API* zcl database access..cacheStats)
+  - [~enable()](#module*DB API* zcl database access..enable)
+  - [~disable()](#module*DB API* zcl database access..disable)
   - [~selectAtomicType(db, packageId, typeName)](#module*DB API* zcl database access..selectAtomicType)
   - [~selectAtomicById(db, packageId)](#module*DB API* zcl database access..selectAtomicById)
   - [~selectAllBitmaps(db)](#module*DB API* zcl database access..selectAllBitmaps) ⇒
+  - [~selectBitmapByNameAndClusterId(db, name, clusterId, packageIds)](#module*DB API* zcl database access..selectBitmapByNameAndClusterId) ⇒
   - [~selectAllEnums(db, packageId)](#module*DB API* zcl database access..selectAllEnums) ⇒
   - [~selectClusterEnums(db, packageId, clusterId)](#module*DB API* zcl database access..selectClusterEnums) ⇒
   - [~selectAllEnumItemsById(db, id)](#module*DB API* zcl database access..selectAllEnumItemsById) ⇒
   - [~selectAllEnumItems(db, packageId)](#module*DB API* zcl database access..selectAllEnumItems) ⇒
   - [~selectEnumById(db, id)](#module*DB API* zcl database access..selectEnumById) ⇒
   - [~selectEnumByName(db, name, packageIds)](#module*DB API* zcl database access..selectEnumByName) ⇒
+  - [~selectEnumByNameAndClusterId(db, name, clusterId, packageIds)](#module*DB API* zcl database access..selectEnumByNameAndClusterId) ⇒
   - [~selectSessionClusterByCode(db, sessionId)](#module*DB API* zcl database access..selectSessionClusterByCode) ⇒
   - [~selectAllSessionClusters(db, sessionId)](#module*DB API* zcl database access..selectAllSessionClusters) ⇒
   - [~selectSessionAttributeByCode(db, sessionId)](#module*DB API* zcl database access..selectSessionAttributeByCode) ⇒
   - [~selectSessionCommandByCode(db, sessionId)](#module*DB API* zcl database access..selectSessionCommandByCode) ⇒
+  - [~selectStructByNameAndClusterId(db, name, clusterId, packageIds)](#module*DB API* zcl database access..selectStructByNameAndClusterId) ⇒
   - [~selectStructsWithClusterAssociation(db, packageIds, groupByStructName)](#module*DB API* zcl database access..selectStructsWithClusterAssociation) ⇒
   - [~selectClusterBitmaps(db, packageId, clusterId)](#module*DB API* zcl database access..selectClusterBitmaps) ⇒
   - [~selectAllDomains(db)](#module*DB API* zcl database access..selectAllDomains) ⇒
@@ -1614,11 +1654,12 @@ Clears the entire cache.
 **Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
 <a name="module_DB API_ zcl database access..put"></a>
 
-### DB API: zcl database access~put(key, packageId, data)
+### DB API: zcl database access~put(key, packageId, data) ⇒
 
 Puts a data object into the cache under a given key/packageId
 
-**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
+**Returns**: Returns true on success.
 
 | Param     | Type            |
 | --------- | --------------- |
@@ -1633,7 +1674,7 @@ Puts a data object into the cache under a given key/packageId
 Returns a data object under a given key/packageId.
 
 **Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
-**Returns**: cached object or null if none is present
+**Returns**: cached object or undefined if none is present or expired.
 
 | Param     | Type            |
 | --------- | --------------- |
@@ -1654,6 +1695,46 @@ Returns true if a given key/packageId cache exists.
 | key       | <code>\*</code> |
 | packageId | <code>\*</code> |
 
+<a name="module_DB API_ zcl database access..cacheQuery"></a>
+
+### DB API: zcl database access~cacheQuery(key, packageId) ⇒
+
+Cache input / output of provided queryFunction
+The queryFunction is assumed to have the following signature:
+
+async function queryFunction(db, ...) {...}
+
+The DB handle is ignored and the remaining arguments are used as the cache key.
+
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
+**Returns**: true or false, depending on whether the cache is present.
+
+| Param     | Type            |
+| --------- | --------------- |
+| key       | <code>\*</code> |
+| packageId | <code>\*</code> |
+
+<a name="module_DB API_ zcl database access..cacheStats"></a>
+
+### DB API: zcl database access~cacheStats()
+
+Returns the cache statistics.
+
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
+<a name="module_DB API_ zcl database access..enable"></a>
+
+### DB API: zcl database access~enable()
+
+Enable the Database Query cache
+
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
+<a name="module_DB API_ zcl database access..disable"></a>
+
+### DB API: zcl database access~disable()
+
+Disable the database cache
+
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
 <a name="module_DB API_ zcl database access..selectAtomicType"></a>
 
 ### DB API: zcl database access~selectAtomicType(db, packageId, typeName)
@@ -1693,6 +1774,22 @@ Retrieves all the bitmaps in the database.
 | Param | Type            |
 | ----- | --------------- |
 | db    | <code>\*</code> |
+
+<a name="module_DB API_ zcl database access..selectBitmapByNameAndClusterId"></a>
+
+### DB API: zcl database access~selectBitmapByNameAndClusterId(db, name, clusterId, packageIds) ⇒
+
+Select a bitmap matched by name and clusterId.
+
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
+**Returns**: bitmap information or undefined
+
+| Param      | Type            |
+| ---------- | --------------- |
+| db         | <code>\*</code> |
+| name       | <code>\*</code> |
+| clusterId  | <code>\*</code> |
+| packageIds | <code>\*</code> |
 
 <a name="module_DB API_ zcl database access..selectAllEnums"></a>
 
@@ -1780,6 +1877,22 @@ Select an enum matched by name.
 | name       | <code>\*</code> |
 | packageIds | <code>\*</code> |
 
+<a name="module_DB API_ zcl database access..selectEnumByNameAndClusterId"></a>
+
+### DB API: zcl database access~selectEnumByNameAndClusterId(db, name, clusterId, packageIds) ⇒
+
+Select an enum matched by name and clusterId.
+
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
+**Returns**: enum information or undefined
+
+| Param      | Type            |
+| ---------- | --------------- |
+| db         | <code>\*</code> |
+| name       | <code>\*</code> |
+| clusterId  | <code>\*</code> |
+| packageIds | <code>\*</code> |
+
 <a name="module_DB API_ zcl database access..selectSessionClusterByCode"></a>
 
 ### DB API: zcl database access~selectSessionClusterByCode(db, sessionId) ⇒
@@ -1835,6 +1948,22 @@ Returns the command available to this session by the code.
 | --------- | --------------- |
 | db        | <code>\*</code> |
 | sessionId | <code>\*</code> |
+
+<a name="module_DB API_ zcl database access..selectStructByNameAndClusterId"></a>
+
+### DB API: zcl database access~selectStructByNameAndClusterId(db, name, clusterId, packageIds) ⇒
+
+Select a struct matched by name and clusterId
+
+**Kind**: inner method of [<code>DB API: zcl database access</code>](#module*DB API* zcl database access)  
+**Returns**: struct information or undefined
+
+| Param      | Type            |
+| ---------- | --------------- |
+| db         | <code>\*</code> |
+| name       | <code>\*</code> |
+| clusterId  | <code>\*</code> |
+| packageIds | <code>\*</code> |
 
 <a name="module_DB API_ zcl database access..selectStructsWithClusterAssociation"></a>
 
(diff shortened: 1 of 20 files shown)


src-electron/db/query-attribute.js
```diff
@@ -32,9 +33,35 @@ function attributeExportMapping(x) {
...
+    isNullable: dbApi.fromDbBool(x.IS_NULLABLE),
+    maxLength: x.MAX_LENGTH,
+    minLength: x.MIN_LENGTH,
+    arrayType: x.ARRAY_TYPE,
+    entryType: x.ARRAY_TYPE,
+    min: x.MIN,
+    max: x.MAX,
+    mustUseTimedWrite: dbApi.fromDbBool(x.MUST_USE_TIMED_WRITE),
+    isSceneRequired: dbApi.fromDbBool(x.IS_SCENE_REQUIRED),
+    isOptionalAttribute: dbApi.fromDbBool(x.IS_OPTIONAL),
```
bzbarsky-apple, line 55: This is not the name IS_OPTIONAL is exposed with for other helpers....  Every single other mapping uses `isOptional` for this.  In general, why are we not just using or existing dbmapping bits here instead of coming up with a new mapping?

src-electron/db/query-attribute.js
```diff
@@ -164,23 +198,68 @@ async function selectAllAttributeDetailsFromEnabledClusters(
...
     ATTRIBUTE.SIDE,
-    ATTRIBUTE.TYPE,
+    CASE
+       WHEN
+         ATTRIBUTE.ARRAY_TYPE IS NULL
+       THEN
+         ATTRIBUTE.TYPE
+       ELSE
+         ATTRIBUTE.ARRAY_TYPE
+     END AS TYPE,
```
bzbarsky-apple, line 208: This seems like a pretty significant behavior change.  Are there no existing consumers of this?

src-electron/generator/helper-session.js
```diff
@@ -612,9 +613,12 @@ function all_user_cluster_commands_irrespective_of_manufaturing_specification(
 
 /**
  * Creates endpoint type cluster attribute iterator. This fetches all
- * manufacturing and non-manufaturing specific attributes which have been enabled
- * on added endpoints
+ * manufacturing and non-manufaturing specific attributes which have been
```
bzbarsky-apple, line 616: What is "non-manufacturing" (which is not spelled right here)?

bzbarsky-apple (commented):
I'm a little confused by what this is trying to address.

There are two ways to use `chip_server_cluster_attributes`.

If includeAll is true, then it just returns all the attributes that are defined in the relevant cluster.  Whether that cluster is enabled or not might affect whether `chip_server_cluster_attributes` is _reached_ at all.

If includeAll is false, then it returns all the attributes that are enabled in the relevant cluster.  Again, whether that cluster is enabled or not affects whether it's reached, but in practice in this mode `chip_server_cluster_attributes` should not be getting invoked for disabled clusters at all.

How is any of this related to the helpers being changed here?

src-electron/db/query-attribute.js
brdandu: The query was using this map so I just extended this one

src-electron/db/query-attribute.js
brdandu: This should be fine because entryType was introduced in matter xml and didn't exist in the old Zigbee 3.0 xml.

src-electron/generator/helper-session.js
brdandu: The standard cluster attributes

src-electron/db/query-attribute.js
bzbarsky-apple: I understand yes. My point is that it led to this map not looking like other maps, which raises the question of why we are not using the other map to start with.

src-electron/db/query-attribute.js
bzbarsky-apple: And there are no consumers who are using this with Matter XML right now?

src-electron/generator/helper-session.js
bzbarsky-apple: I think that should be rephrased to be clear, then.

brdandu:
> I'm a little confused by what this is trying to address.
> 
> There are two ways to use `chip_server_cluster_attributes`.
> 
> If includeAll is true, then it just returns all the attributes that are defined in the relevant cluster. Whether that cluster is enabled or not might affect whether `chip_server_cluster_attributes` is _reached_ at all.
> 
> If includeAll is false, then it returns all the attributes that are enabled in the relevant cluster. Again, whether that cluster is enabled or not affects whether it's reached, but in practice in this mode `chip_server_cluster_attributes` should not be getting invoked for disabled clusters at all.
> 
> How is any of this related to the helpers being changed here?

I switched {{#chip_server_cluster_attributes}} in MatterIDL.zapt to {{#chip_server_cluster_attributes includeAll=false}} but that is not removing all the server attributes which should not be generated because the corresponding server cluster is disabled.

src-electron/db/query-attribute.js
brdandu: Yeah this was not being used by Matter before.

src-electron/db/query-attribute.js
brdandu: I can look into converging this with the other map. Probably some divergence from old code.

bzbarsky-apple:
> I switched {{#chip_server_cluster_attributes}} in MatterIDL.zapt to {{#chip_server_cluster_attributes includeAll=false}}

The includeAll=false needs to be on the `chip_server_clusters`, no?  And that's the default behavior, by the way....

The real problem, it seems to me, is that `chip_server_cluster_attributes` is supposed to be used inside `chip_server_clusters`, which only iterates the _enabled ones_ unless `includeAll=true`.  The documentation for the function even says that.

But MatterIDL.zapt uses it inside `all_user_clusters` for some reason, which iterates the disabled ones?

src-electron/db/query-attribute.js
bzbarsky-apple: Thank you.  Followup is OK for that, in general, but for now we should just make sure they match exactly for the new things being added.

brdandu:
> chip_server_clusters



> > I switched {{#chip_server_cluster_attributes}} in MatterIDL.zapt to {{#chip_server_cluster_attributes includeAll=false}}
> 
> The includeAll=false needs to be on the `chip_server_clusters`, no? And that's the default behavior, by the way....
> 
> The real problem, it seems to me, is that `chip_server_cluster_attributes` is supposed to be used inside `chip_server_clusters`, which only iterates the _enabled ones_ unless `includeAll=true`. The documentation for the function even says that.
> 
> But MatterIDL.zapt uses it inside `all_user_clusters` for some reason, which iterates the disabled ones?

I am looking at ChipCallbackTypes.zapt which uses {{#chip_client_clusters}} and {{#chip_server_cluster_attributes}}. I see Identify server cluster attributes generated in ChipCallbackTypes.h even though the Identify server cluster is disabled. I do not see includeAll=true here(Assuming includeAll is false by default like you mentioned).

bzbarsky-apple:
> I am looking at ChipCallbackTypes.zapt which uses {{#chip_client_clusters}} and {{#chip_server_cluster_attributes}}.

That one is a hack which explicitly relies on "we enable the server attributes, but then actually enable the client clusters".

That is, it _wants_ the observed behavior.  It should really be using server clusters and includeAll=true, but the owners of that code have sort of abandoned it....

That is, this is codegen for the java "control everything" client.  It should not depend on what server clusters are enabled (none of them, in practice) for controllers.

brdandu:
> > I am looking at ChipCallbackTypes.zapt which uses {{#chip_client_clusters}} and {{#chip_server_cluster_attributes}}.
> 
> That one is a hack which explicitly relies on "we enable the server attributes, but then actually enable the client clusters".
> 
> That is, it _wants_ the observed behavior. It should really be using server clusters and includeAll=true, but the owners of that code have sort of abandoned it....
> 
> That is, this is codegen for the java "control everything" client. It should not depend on what server clusters are enabled (none of them, in practice) for controllers.

To do further analysis here. Which of the following files are not hacks and not abandoned?
       modified:   src/controller/java/zap-generated/CHIPAttributeTLVValueDecoder.cpp
	modified:   src/controller/java/zap-generated/CHIPCallbackTypes.h
	modified:   src/controller/java/zap-generated/CHIPClustersWrite-JNI.cpp
	modified:   src/controller/java/zap-generated/CHIPReadCallbacks.cpp
	modified:   src/controller/java/zap-generated/CHIPReadCallbacks.h
	modified:   src/controller/java/zap-generated/chip/devicecontroller/ChipClusters.java
	modified:   src/controller/java/zap-generated/chip/devicecontroller/ChipIdLookup.java
	modified:   src/controller/java/zap-generated/chip/devicecontroller/ClusterInfoMapping.java
	modified:   src/controller/java/zap-generated/chip/devicecontroller/ClusterReadMapping.java
	modified:   src/controller/java/zap-generated/chip/devicecontroller/ClusterWriteMapping.java
	modified:   src/controller/python/chip/clusters/CHIPClusters.py
        modified:   src/darwin/Framework/CHIP/zap-generated/MTRAttributeTLVValueDecoder.mm
	modified:   src/darwin/Framework/CHIP/zap-generated/MTRBaseClusters.h
	modified:   src/darwin/Framework/CHIP/zap-generated/MTRBaseClusters.mm
	modified:   src/darwin/Framework/CHIP/zap-generated/MTRCallbackBridge.h
	modified:   src/darwin/Framework/CHIP/zap-generated/MTRCallbackBridge.mm
	modified:   src/darwin/Framework/CHIP/zap-generated/MTRClusters.h
	modified:   src/darwin/Framework/CHIP/zap-generated/MTRClusters.mm
        modified:   zzz_generated/chip-tool/zap-generated/test/Commands.h
	modified:   zzz_generated/darwin-framework-tool/zap-generated/cluster/Commands.h
	modified:   zzz_generated/darwin/controller-clusters/zap-generated/CHIPClientCallbacks.h

bzbarsky-apple:
> Which of the following files are not hacks and not abandoned?

_Files_ are not hacks.  Specific things in a file might be hacks.

The java stuff is all more or less abandoned, at the moment.  Maybe new maintainers can be found for it; please check with the Google folks.

The darwin and chip-tool bits are not abandoned, but they also have the behavior they want to have right now.  So any PR that changes their output needs to be looked at very carefully to see what changed about the output and why.

brdandu:
> > Which of the following files are not hacks and not abandoned?
> 
> _Files_ are not hacks. Specific things in a file might be hacks.
> 
> The src/controller/java stuff is all more or less abandoned, at the moment, as far as I understand. Maybe new maintainers can be found for it; please check with the Google folks.
> 
> The python, darwin and chip-tool bits are not abandoned, but they also have the behavior they want to have right now. So any PR that changes their output needs to be looked at very carefully to see what changed about the output and why.

@andy31415 would you know anything about the java code. Is that still being used or can it be removed?

andy31415:
> abandoned

I do not believe the java stuff was abandoned. Historically we had to move some of the files to be code generated by the python code generator because a single large file would crash the gcc NDK compiler with internal error.

However converting handlebars to python jinja2 templates is a non-trivial effort, so we converted only what was crashing and not the rest. Which puts us in the awkward state of having 2 code generators even for java itself.

bzbarsky-apple:
@brdandu To be clear, the Java stuff is _used_.  It's just not particularly actively _maintained_ and hasn't been in a while as far as I can tell.

brdandu:
> @brdandu To be clear, the Java stuff is _used_. It's just not particularly actively _maintained_ and hasn't been in a while as far as I can tell.

ok I will make some changes to the templates here and run them by you.

src-electron/db/db-mapping.js
```diff
@@ -475,36 +481,61 @@ exports.map = {
...
+      mustUseTimedWrite: dbApi.fromDbBool(x.MUST_USE_TIMED_WRITE),
+      name: x.NAME, // Attribute Name
+      reportableChange: x.REPORTABLE_CHANGE,
+      side: x.SIDE, // Attribute Side
+      singleton: dbApi.fromDbBool(x.SINGLETON),
       smallestEndpointIdentifier: x.SMALLEST_ENDPOINT_IDENTIFIER, // Smallest endpoint Id in which the attribute is present
+      storage: x.STORAGE_OPTION,
+      storageOption: x.STORAGE_OPTION,
+      tokenId: x.TOKEN_ID, // Endpoint type attribute's token id
+      type: x.TYPE, // Attribute type
```
bzbarsky-apple, line 538: Why does this not match the "attribute" map?  Is that because the underlying query it's used for does this "stick the ARRAY_TYPE in TYPE" bit?  And if so, why the difference in implementation?

src-electron/db/query-attribute.js
```diff
@@ -247,7 +283,7 @@ async function selectAttributeDetailsFromAllEndpointTypesAndClustersUtil(
   GROUP BY ATTRIBUTE.NAME
         `
     )
-    .then((rows) => rows.map(attributeExportMapping))
+    .then((rows) => rows.map(dbMapping.map.endpointTypeAttributeExtended))
```
bzbarsky-apple, line 286: But this does not use the same thing for TYPE that the other consumer of endpointTypeAttributeExtended does.... Why not?

src-electron/generator/helper-attribute.js
```diff
@@ -65,6 +66,77 @@ async function attributeDefault(options) {
   return templateUtil.templatePromise(this.global, p)
 }
 
+/**
+ * If helper that checks if an attribute is not supported.
+ * An attribute is not supported when it is a struct. However it is supported
```
bzbarsky-apple, line 71: "supported" in what sense?  Matter absolutely supports struct-typed attributes....  This needs a better description of what it's really doing and why.

src-electron/generator/helper-attribute.js
```diff
@@ -65,6 +66,77 @@ async function attributeDefault(options) {
...
+ * @returns Promise of content
+ */
+async function if_unsupported_attribute_type(type, isArray, options) {
+  let struct = null
+  if (isArray) {
+    return options.inverse(this)
+  } else {
+    let packageIds = await templateUtil.ensureZclPackageIds(this)
+    if (type && typeof type == 'string') {
+      struct = await queryZcl.selectStructByName(
```
bzbarsky-apple, line 85: Won't this fail if structs with the same name are defined in multiple clusters?

src-electron/generator/helper-attribute.js
```diff
@@ -65,6 +66,77 @@ async function attributeDefault(options) {
...
+ * @param {*} options
+ * @returns Promise of content
+ */
+async function if_unsupported_attribute_type(type, isArray, options) {
+  let struct = null
+  if (isArray) {
+    return options.inverse(this)
+  } else {
+    let packageIds = await templateUtil.ensureZclPackageIds(this)
+    if (type && typeof type == 'string') {
```
bzbarsky-apple, line 84: So this allow passing a type name or a type id for "type"?  That is ... extremely confusing.  Seems like the public API should be clearer about what it expects.

src-electron/generator/helper-attribute.js
```diff
@@ -65,6 +66,77 @@ async function attributeDefault(options) {
...
+    } else {
+      return options.inverse(this)
+    }
+  }
+}
+
+/**
+ * If helper that checks if an attribute is complex or not.
+ * An attribute is complex if it is either nullable, optional, array type or
+ * struct.
```
bzbarsky-apple, line 106: Why is this the right definition of "complex"?  In general, "optional" doesn't mean "complex".....

src-electron/generator/helper-attribute.js
```diff
@@ -65,6 +66,77 @@ async function attributeDefault(options) {
...
+ * {{else}}
+ * type is not complex
+ * {{/if_attribute_complex}}
+ * @param {*} type
+ * @param {*} options
+ * @returns Promise of content
+ */
+async function if_attribute_complex(type, options) {
+  let struct = null
+  if (this.isNullable || this.isOptional || this.isArray) {
```
bzbarsky-apple, line 121: If we're assuming that `this` has useful state, why are we requiring the type to be passed in instead of just getting it of `this` as well?

src-electron/generator/helper-attribute.js
```diff
@@ -65,6 +66,77 @@ async function attributeDefault(options) {
...
+    return options.fn(this)
+  } else {
+    let packageIds = await templateUtil.ensureZclPackageIds(this)
+    if (type && typeof type == 'string') {
+      struct = await queryZcl.selectStructByName(
```
bzbarsky-apple, line 126: Again, how does this handle identically-named structs in different clusters?

src-electron/generator/helper-session.js
```diff
@@ -294,7 +294,8 @@ async function all_user_cluster_attribute_util(
       await queryAttribute.selectAllAttributeDetailsFromEnabledClusters(
         currentContext.global.db,
         endpointsAndClusters,
-        packageIds
+        packageIds,
+        side
```
bzbarsky-apple, line 298: This used to not restrict to a particular side, right?  Why is this change OK?

src-electron/generator/helper-session.js
```diff
@@ -612,18 +616,22 @@ function all_user_cluster_commands_irrespective_of_manufaturing_specification(
 
 /**
  * Creates endpoint type cluster attribute iterator. This fetches all
- * manufacturing and non-manufaturing specific attributes which have been enabled
- * on added endpoints
+ * manufacturing and standard attributes which have been enabled on
```
bzbarsky-apple, line 619:
```suggestion
 * manufacturer-specific and standard attributes which have been enabled on
```

src-electron/generator/helper-session.js
```diff
@@ -612,18 +616,22 @@ function all_user_cluster_commands_irrespective_of_manufaturing_specification(
...
+ * added endpoints based on the name and side of the cluster. When side
+ * is not mentioned then client and server attributes are returned.
+ * Available Options:
+ * - removeKey: Removes one of the keys from the map(for eg keys in db-mapping.js)
```
bzbarsky-apple, line 623: Why is this useful?  And why is it useful to only allow removing one key?

src-electron/generator/helper-zcl.js
```diff
@@ -730,33 +730,38 @@ function zcl_attributes_client(options) {
  * Iterator over the server attributes. If it is used at toplevel, if iterates over all the server attributes
  * in the database. If used within zcl_cluster context, it iterates over all the server attributes
  * that belong to that cluster.
- *
+ * Available Options:
+ * - removeKey: Removes one of the keys from the map(for eg keys in db-mapping.js)
```
bzbarsky-apple, line 734: Again, why only one key, if this is useful at all?

src-electron/generator/helper-zcl.js
```diff
@@ -2693,6 +2698,261 @@ function if_compare(leftValue, rightValue, options) {
...
+  let isSigned = false
+  if (dataType.discriminatorName.toLowerCase() == dbEnum.zclType.bitmap) {
+    let bitmap = await queryZcl.selectBitmapByName(
+      context.global.db,
+      packageIds,
+      dataType.name
+    )
+    result = bitmap.size
+  } else if (dataType.discriminatorName.toLowerCase() == dbEnum.zclType.enum) {
+    let en = await queryZcl.selectEnumByName(
```
bzbarsky-apple, line 2725: How does this work if the same enum name is used in multiple clusters?

src-electron/generator/helper-zcl.js
```diff
@@ -2693,6 +2698,261 @@ function if_compare(leftValue, rightValue, options) {
...
+ * @param {*} context
+ * @returns The corresponding cpp data type for a zcl data type.
+ */
+async function as_underlying_cpp_zcl_type(type, dataType, hash, context) {
+  let res = await new Promise((resolve, reject) => {
+    let asUpperCamelCase =
+      hash && hash.asUpperCamelCase ? hash.asUpperCamelCase : false
+    if (dataType.discriminatorName.toLowerCase() == dbEnum.zclType.string) {
+      if (octetStringTypes.includes(type.toUpperCase())) {
+        resolve('OctetString')
```
bzbarsky-apple, line 2763: I'm not sure I follow this.  There's no OctetString C++ type in the Matter sdk.  Or CharString.

src-electron/generator/helper-zcl.js
```diff
@@ -2864,3 +3124,5 @@ exports.as_type_max_value = as_type_max_value
 exports.as_type_min_value = as_type_min_value
 exports.as_zcl_type_size = as_zcl_type_size
 exports.if_compare = if_compare
+exports.as_underlying_language_specific_zcl_type =
```
bzbarsky-apple, line 3127: This feels like it should be handled via overridables like the existing basic_types stuff...  Having it all hardcoded here is odd, since different things may want different type mappings.

src-electron/db/db-mapping.js
brdandu: All the maps in db-mapping are actually an exact representation of the database tables. I wanted to keep that map the same way for now. I created an extended one because this does not just have entries from the endpointTypeAttribute table but other tables as well. You are also right about the array_type in type bit but that is not the reason I made a separate map.

This way we can bring all the maps floating around in the code into db-mapping.

src-electron/generator/helper-session.js
brdandu: This was added as an option to handle https://github.com/project-chip/zap/blob/master/src-electron/generator/matter/app/zap-templates/common/ClustersHelper.js#L287 when replacing `chip_server_cluster_attributes` and generate the same code using `enabled_attributes_for_cluster_and_side` and `zcl_attributes_server`

brdandu: I see your point though. I can update it to removeMultiple keys.

src-electron/generator/helper-zcl.js
brdandu: Explained above

src-electron/db/db-mapping.js
bzbarsky-apple: My point is this: all maps for attributes should have "type" meaning the same thing.  And right now we don't seem to be doing that.

src-electron/db/query-attribute.js
bzbarsky-apple: How are consumers supposed to work with "type" if it means different things for the same objects depending on how you got them?  Consumers have no idea what these things look like under the hood, and its easy to end up with bad mistakes.

Maybe the answer needs to be that we deprecate "type" on attributes and define new field names that are sane and consistent?  Not necessary to block this PR on that, I guess, but we should document the discrepancies in the code and get issues filed.  But that's likely to make it impossible to work in a consistent way with attributes and things like command fields too, though I think they may already mean different things by "type" right now....

This is not a theoretical problem: the "type" weirdness on attributes led to a _lot_ of bugs in Matter until we caught (we think) all the issues.

bzbarsky-apple: And just to be clear: from my point of view, Zigbee Pro constraints are not a reason for _Matter_ to have problems.  We need to make sure that all Matter helpers are consistent, and the ones that do different things that Zigbee Pro needs should not be exposed to Matter at all.

src-electron/generator/helper-attribute.js
bzbarsky-apple: If that's what we are trying to represent, I suggest we name it "if_attribute_type_unsupported_in_java_codegen" or something, so it's clear what it _really_ means.  And then we can put it in some helper js file that is _only_ pulled in by the java codegen, for clarity.

src-electron/generator/helper-session.js
bzbarsky-apple: I see, ok.  Yeah, having some sort of more flexible setup here, if it's not too hard to implement and not too hard to then use, would be good.

src-electron/generator/helper-session.js
brdandu: It defaults to null when not specified and the query is handled accordingly in that case.

src-electron/generator/helper-zcl.js
brdandu: They seem pretty hard coded in https://github.com/project-chip/zap/blob/master/src-electron/generator/matter/controller/java/templates/helper.js . This should help get all of the dataType and zclType logic to one spot and deprecate the others.

src-electron/generator/helper-attribute.js
brdandu: Fixed this to include clusterIds as well

src-electron/generator/helper-zcl.js
brdandu: Resolved this

src-electron/generator/helper-zcl.js
brdandu: Sorry this was objective C. Fixed this

src-electron/generator/helper-zcl.js
bzbarsky-apple: If these are Matter-Java-codegen-specific, why are we using a generic name for the helper?

src-electron/generator/helper-zcl.js
```diff
@@ -2693,6 +2698,293 @@ function if_compare(leftValue, rightValue, options) {
...
+  dataType,
+  hash,
+  context
+) {
+  let res = await new Promise((resolve, reject) => {
+    let asUpperCamelCase =
+      hash && hash.asUpperCamelCase ? hash.asUpperCamelCase : false
+    if (dataType.discriminatorName.toLowerCase() == dbEnum.zclType.string) {
+      if (octetStringTypes.includes(type.toUpperCase())) {
+        resolve('OctetString')
```
bzbarsky-apple, line 2773: There is _definitely_ no objective C OctetString/CharString thing.

src-electron/db/db-mapping.js
brdandu: I forgot type was already set to `type: x.TYPE != 'array' ? x.TYPE : x.ARRAY_TYPE,` for attribute. Fixed this

src-electron/db/query-attribute.js
brdandu: I forgot what we were doing in attribute map. the endpoint type extended map now does the same thing as attribute map. Type will be based off of array_type and type now. See my change in db-mapping.

src-electron/generator/helper-attribute.js
brdandu: Fixed this. Now it just takes names

src-electron/generator/helper-zcl.js
brdandu: This comes from attribute_data_callback_name which is referenced by MTRBaseClusters-src.zapt. I named it objectiveC because it was used to generate a .mm file. Do you want me to name this something else?

bzbarsky-apple: I don't see any OctetString bits in https://github.com/project-chip/connectedhomeip/blob/3c6bed3bb28f66cc2e1ebbe3561fa0838f6b7c51/src/darwin/Framework/CHIP/templates/partials/attribute_data_callback_name.zapt (as of today).  What am I missing?

src-electron/generator/helper-session.js
brdandu: Fixed this

src-electron/generator/helper-zcl.js
brdandu: https://github.com/project-chip/connectedhomeip/blob/3c6bed3bb28f66cc2e1ebbe3561fa0838f6b7c51/src/darwin/Framework/CHIP/templates/partials/attribute_data_callback_name.zapt#L18 that part?

bzbarsky-apple: I guess it's doing `chipCallback.name`?

If that's the thing we are trying to deal with this has nothing to do with "as_underlying_objective_c_zcl_type".  What's really going on there is how the type name is being mapped into function type names (for the callback function types).  The actual setup there is somewhat complicated, since it needs to encode nullability, lists of things (just bare "List" is definitely not ok there, for example).

src-electron/generator/helper-zcl.js
brdandu: To give more context. I am handling things like `{{asJavaBoxedType type chipType}}`, `{{convertAttributeCallbackTypeToJavaName chipCallback.type}}`, `{{asPythonType chipType}}` and `{{chipCallback.name}}` from within stateless helpers such as `zcl_server_attributes` and `enabled_attributes_for_cluster_and_side` instead of `chip_server_cluster_attributes`. That is the reason I created these language specific zcl type helpers and felt having all of these in one place is a good cleanup. I could move these out of this file and put it in the respective files. Is that what you want?

bzbarsky-apple: >  I could move these out of this file and put it in the respective files.

I would prefer this, yes.

And again, if we can do {{chipCallback.name}} without extra helpers just in the one template that uses it, that would be ideal.....

src-electron/generator/helper-zcl.js
brdandu: I see that there are more references to `chipCallback.name` than I had pulled in locally but I was able to handle that using the following in attribute_data_callback_name.zapt:
`{{as_underlying_language_specific_zcl_type type ../id signedPrefix='Int' unSignedPrefix='Int' unSignedPostfix='u' signedPostfix='s' sizeIn='bits' roundUpToPowerOfTwo=1 language='objectiveC' boolean='Boolean' asUpperCamelCase=1}}`

I now understand that I should not be using `language='objectiveC'` here. I can move all of that into a helper with less options and call it `{{as_underlying_callback_for_zcl_type type ../id}}` . I could also continue to have those options if that needs to be different at times based on language being generated for(Make it more scalable).

bzbarsky-apple: The thing is, as_underlying_callback_for_zcl_type doesn't really make sense either, since it's only (1) part of the callback name (2) only for some types.

What does that function end up returning in the end?  Maybe we can just replace it with something simpler....

src-electron/generator/helper-attribute.js
brdandu: It is now only unsupported if language is java. Else it is supported.

src-electron/generator/helper-attribute.js
brdandu: added language option to this.

src-electron/generator/helper-attribute.js
brdandu: It now passes type and clusterId which makes the helper more readable to have those 2 specified but unresolved this if you do not like this.

src-electron/generator/helper-attribute.js
brdandu: added clusterId

src-electron/generator/helper-zcl.js
brdandu: The function as_underlying_language_specific_zcl_type/as_underlying_callback_for_zcl_type just returns a string based on the zcl type given

brdandu: Right now it is doing ```async function as_underlying_objective_c_zcl_type(
  type,
  dataType,
  hash,
  context
) {
  let res = await new Promise((resolve, reject) => {
    let asUpperCamelCase =
      hash && hash.asUpperCamelCase ? hash.asUpperCamelCase : false
    if (dataType.discriminatorName.toLowerCase() == dbEnum.zclType.string) {
      if (octetStringTypes.includes(type.toUpperCase())) {
        resolve('OctetString')
      } else if (characterStringTypes.includes(type.toUpperCase())) {
        resolve('CharString')
      } else if (context.isArray) {
        resolve('List')
      } else {
        resolve(null)
      }
    } else if (dataType.name.includes('float')) {
      return asUpperCamelCase ? resolve('Float') : resolve('float')
    } else if (dataType.name.includes('double')) {
      return asUpperCamelCase ? resolve('Double') : resolve('double')
    } else if (dataType.name.includes('single')) {
      return asUpperCamelCase ? resolve('Float') : resolve('float')
    }
    resolve(null)
  })
  return res
}```

src-electron/generator/helper-zcl.js
brdandu: I have moved the code for python and java helpers to the respective files. Need to just figure out on what needs to be done for something like `{{chipCallback.name}}`

brdandu: I think this has too many exceptions to create the right output in the template and can lead to comparing a lot of that in the template. For now I have clubbed this into a helper.

src-electron/generator/helper-zcl.js
brdandu: For now I renamed the helper to avoid any language specific activity. Instead it looks for a modifier such as `chipCallbackName` to determine the right output for exceptions.

src-electron/generator/helper-attribute.js
```diff
@@ -73,3 +155,5 @@ async function attributeDefault(options) {
 
 exports.global_attribute_default = attributeDefault
 exports.feature_bits = featureBits
+exports.if_unsupported_attribute_callback = if_unsupported_attribute_callback
```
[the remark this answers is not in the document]
brdandu: This is currently only unsupported if user specifies the java option for this helper.

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
```diff
@@ -772,7 +779,9 @@ exports.chip_endpoints = chip_endpoints;
 exports.chip_endpoint_clusters = chip_endpoint_clusters;
 exports.if_chip_enum = if_chip_enum;
 exports.if_chip_complex = if_chip_complex;
-exports.if_basic_global_response = if_basic_global_response;
+exports.if_basic_global_response = dep(if_basic_global_response, {
+  to: 'if_attribute_complex',
```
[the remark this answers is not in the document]
brdandu: Fixed this to if_basic_attribute in helper-attribute.js

src-electron/generator/helper-zcl.js
brdandu: `as_underlying_zcl_type_modifier` helper handles `chipCallbackName`. The above image was not the complete implementation. `as_underlying_zcl_type_modifier` calls the above image along the way to handle certain exceptions. If the above image returns null then it handles integers like you asked.

bzbarsky-apple: > Regarding the context.isArray. I believe I got that fromClusterHelper#asChipCallback

Yes, but we never use `chipCallback.name` if `isArray` is true, at least in the Darwin code.

So just so I can understand: all this machinery is to replace the `name` computation we have right now in `asChipCallback`?  It looks like that's used in one place in the Darwin code and in a bunch of places in Java code.

Speaking about the Darwin side for now, and looking at asChipCallback:

1. We can detect octet and char strings directly in the template and do the right thing.
2. We already detect arrays in the template.
3. For the numeric types, what we really need are:
    a. Ways to detect/canonicalize the type for bool/float/double.  Do those exist already?
    b. A way to get the integer width, in bytes, for an integer type.
    c. A way to check whether an integer type is signed, for an integer type.

If we had those pieces for numeric types, this is all doable in the template without building all this machinery.

I can't speak to the java pieces, though even there I expect a lot of it can be done similarly.

Basically, I don't think we should be building a 1-1 replacement for `chipCallback.name`; that's a very leaky abstraction that sort of made sense when the helpers were in-tree but does not make sense as an actual ZAP API.  Instead we should be exposing information about the types and letting people write templates that do what they want.

brdandu: I am a proponent of doing things in the template but have had resistance with adding too much to the template in the past. There will be a lot of if else logic in the template. Are we ok with that? If so, then I can remove the logic surrounding `chipCallback.name`
The java code has a lot more references to `chipCallback.name`

brdandu: Are you ok with doing the following in the template instead
`{{#if_is_octet_string type}}
      OctetString
      {{else}}
        {{#if_is_char_string type}}
        CharString
        {{else}}
          {{#if isArray}}
          List
          {{else if (is_str_equal type "float")}}
          Float
          {{else if (is_str_equal type "double")}}
          Double
          {{else if (is_str_equal type "single")}}
          Float
          {{else if (is_str_equal type "boolean")}}
          Boolean
          {{else}}
            {{#if_is_data_type_signed type ../id}}
            Int{{as_zcl_data_type_size type ../id sizeIn='bits' roundUpToPowerOfTwo=1}}s
            {{else}}
            Int{{as_zcl_data_type_size type ../id sizeIn='bits' roundUpToPowerOfTwo=1}}u
            {{/if_is_data_type_signed}}
          {{/if}}
        {{/if_is_char_string}}
      {{/if_is_octet_string}}`

bzbarsky-apple: > if_is_octet_string type

Note that we have `isOctetString` already.

> if_is_char_string

Similarly, we have `isCharString`

> `{{#if isArray}} List`

That's not what the template would have, but sure.

> `is_str_equal type "float"`

An `isFloat` that works on types and does the float/single thing would be useful, perhaps, but in practice do we actually have both "float" and "single" in the XML?   Alternately, the template could just do `if (or ....)` as needed.

> if_is_data_type_signed

That looks great, but why does it need `../id`?

> as_zcl_data_type_size

That also looks great.  We could simplify things a bit by doing the "Int{size}" bits once and just have the signed conditional around the "s/u" decision.

But yes, in general that sort of if/else cascade is just fine and we do that in other places (e.g. `src/darwin/Framework/CHIP/templates/partials/init_struct_member.zapt`, `src/darwin/Framework/CHIP/templates/partials/decode_value.zapt`, etc.

src-electron/generator/helper-attribute.js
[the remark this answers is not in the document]
bzbarsky-apple: Let me ask this a different way:  What does this let us do that just using `{{if_is_struct type}}` in the java codegen would not let us do?  Possibly after checking for "array of structs".

src-electron/generator/helper-zcl.js
brdandu: ok I will start adding the requirements for the above.
`../id` is the clusterId in this case.

brdandu: Fixed this. Got rid of it.

src-electron/generator/helper-attribute.js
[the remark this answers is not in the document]
brdandu: Yes that can be done.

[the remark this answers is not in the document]
brdandu: I have removed these from the list of generic helpers and put them along with the java helper.js

docs/api.md
bzbarsky-apple, line 1: I did not review the changes to this file, since a bunch of them seem unrelated to this PR...  It might be better to split out the unrelated bits to a separate PR.

src-electron/db/query-attribute.js
```diff
@@ -165,26 +156,71 @@ async function selectAllAttributeDetailsFromEnabledClusters(
...
     ATTRIBUTE.SIDE,
-    ATTRIBUTE.TYPE,
+    CASE
+       WHEN
+         ATTRIBUTE.ARRAY_TYPE IS NULL
+       THEN
+         ATTRIBUTE.TYPE
+       ELSE
+         ATTRIBUTE.ARRAY_TYPE
+     END AS TYPE,
```
bzbarsky-apple, line 166: But doesn't the mapping handle this now?  Why do we need to handle this in the query _and_ the mapping?

tecimovic (approved):
This seems like a reasonable step towards replacing stateful helpers with stateless helpers. Work in progress...

src-electron/db/query-bitmap.js
```diff
@@ -68,6 +68,62 @@ WHERE (DATA_TYPE.NAME = ? OR DATA_TYPE.NAME = ?) AND DATA_TYPE.PACKAGE_REF IN ($
...
+    .then((rows) => rows.map(dbMapping.map.bitmap))
+
+  if (res && res.length == 1) {
+    return res[0]
+  } else {
+    return dbApi
+      .dbGet(
+        db,
+        `
+SELECT
```
bzbarsky-apple, line 104: The query duplication here seems a bit sub-optimal, in that we have to remember to change both queries in sync.

If there is a good reason (you mentioned performance optimization) to not just do the one query that includes the cluster id, then maybe we could have a function that returns the query string and takes an argument for whether to include the cluster id check, and does string concatenation or some other mechanism to include it or not while keeping the rest of the query as a single string?  Similar for other places where we are adding such queries.

src-electron/db/query-bitmap.js
```diff
@@ -68,6 +68,62 @@ WHERE (DATA_TYPE.NAME = ? OR DATA_TYPE.NAME = ?) AND DATA_TYPE.PACKAGE_REF IN ($
...
+  DATA_TYPE
+ON
+  BITMAP.BITMAP_ID = DATA_TYPE.DATA_TYPE_ID
+INNER JOIN
+  DATA_TYPE_CLUSTER
+ON
+  DATA_TYPE_CLUSTER.DATA_TYPE_REF = BITMAP.BITMAP_ID
+WHERE
+  (DATA_TYPE.NAME = ? OR DATA_TYPE.NAME = ?)
+  AND DATA_TYPE_CLUSTER.CLUSTER_REF = ?
```
bzbarsky-apple, line 119: Would it perhaps make more sense to match on cluster code, not the temporary-for-the-database id?  That would allow templates to do checks with known cluster codes as needed....  Again, same for the other places where we are adding these sorts of queries.

src-electron/generator/helper-session.js
```diff
@@ -612,18 +617,22 @@ function all_user_cluster_commands_irrespective_of_manufaturing_specification(
 
 /**
  * Creates endpoint type cluster attribute iterator. This fetches all
- * manufacturing and non-manufaturing specific attributes which have been enabled
- * on added endpoints
+ * manufacturing-specific and standard attributes which have been enabled on
```
bzbarsky-apple, line 620:
```suggestion
 * manufacturer-specific and standard attributes which have been enabled on
```

src-electron/generator/helper-zcl.js
```diff
@@ -2693,6 +2699,83 @@ function if_compare(leftValue, rightValue, options) {
...
 
+/**
+ * Check if the given type is signed or not based on the type name and cluster
+ * id.
+ * @param {*} type
+ * @param {*} clusterId
+ * @param {*} options
+ * @returns Promise of content
+ */
+async function if_is_data_type_signed(type, clusterId, options) {
```
bzbarsky-apple, line 2710: Again, is it better to use id or cluster code here?  For purposes of doing this in a loop over clusters it doesn't matter much, but should this be usable outside such a loop?

src-electron/generator/helper-zcl.js
```diff
@@ -2693,6 +2699,83 @@ function if_compare(leftValue, rightValue, options) {
...
+ * Available Options:
+ * - roundUpToPowerOfTwo: Rounds the size up to the nearest power of 2
+ * - sizeIn: By default size is returned in bytes but it can be returned in bits
+ * by mentioning sizeIn="bits"
+ * @param {*} type
+ * @param {*} clusterId
+ * @param {*} options
+ * @returns size of the data type
+ */
+async function as_zcl_data_type_size(type, clusterId, options) {
```
bzbarsky-apple, line 2746: Again, cluster id vs cluster code?

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
```diff
@@ -745,6 +745,97 @@ async function chip_access_elements(options) {
...
+ * type is basic
+ * {{else}}
+ * type is not basic
+ * {{/if_basic_attribute}}
+ * @param {*} type
+ * @param {*} clusterId
+ * @param {*} options
+ * @returns Promise of content
+ */
+async function if_basic_attribute(type, clusterId, options) {
```
bzbarsky-apple, line 767: This is Matter-java-codegen-specific, right?  Including some Matter-java-codegen specific definition of "basic"?

Please put this in the java helper file, not here.

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
```diff
@@ -745,6 +745,97 @@ async function chip_access_elements(options) {
...
+  let hash = options.hash;
+  let language = hash && hash.language ? hash.language : null;
+  if (language == 'java') {
```
bzbarsky-apple, line 770: And then we don't need the language check.

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
```diff
@@ -745,6 +745,97 @@ async function chip_access_elements(options) {
...
+ * However it is supported if it is an array of structs.
+ * Available Options:
+ * - language: attribute callback support based on language
+ * @param {*} type
+ * @param {*} isArray
+ * @param {*} clusterId
+ * @param {*} options
+ * @returns Promise of content
+ */
+async function if_unsupported_attribute_callback(
```
bzbarsky-apple, line 806: Again, this is Matter-java-specific, right?  Please place this in the relevant helper file, remove the language bit.

It also feels like we should be able to build this out of existing "is this an array?" and "is this a struct type?" pieces, but since I am hoping this will just go away sometime I don't care too much about that part...

src-electron/generator/matter/controller/python/templates/helper.js
```diff
@@ -43,6 +49,58 @@ function asPythonType(zclType) {
...
+    type,
+    clusterId,
+    packageIds
+  );
+  if (
+    dataType.discriminatorName.toLowerCase() == dbEnum.zclType.bitmap ||
+    dataType.discriminatorName.toLowerCase() == dbEnum.zclType.enum ||
+    dataType.discriminatorName.toLowerCase() == dbEnum.zclType.number
+  ) {
+    // returning nothing for floats
```
bzbarsky-apple, line 86: Why?  The comment should explain the "why", not the "what"....

src-electron/generator/matter/controller/python/templates/helper.js
```diff
@@ -43,6 +49,58 @@ function asPythonType(zclType) {
...
+ * for eg: (as_underlying_python_zcl_type type clusterId boolean='bool')
+ * will return "bool" for "boolean" type
+ * @param {*} type
+ * @param {*} clusterId
+ * @param {*} options
+ * @returns The corresponding python data type for a zcl data type.
+ */
+async function as_underlying_python_zcl_type(type, clusterId, options) {
+  let hash = options.hash;
+  // Overwrite any type with the one coming from the template options
```
bzbarsky-apple, line 66: This is only used for "boolean" and "BOOLEAN", right?  Why wouldn't we just handle those directly in this helper like we handle other types?

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
brdandu: This is where the corresponding deprecated helper is(if_basic_global_response). Do you still want this in a newly created java helper file?

src-electron/generator/matter/controller/python/templates/helper.js
brdandu: I don't know, the deprecated helper `asPythonCType` does not seem to say why

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
bzbarsky-apple: I think so, yes.  The if_basic_global_response thing was mis-placed to start with...

src-electron/generator/matter/controller/python/templates/helper.js
brdandu: 2 things:
- It becomes a one line corresponding change in the template. So less questions asked in a PR:)
- I sometimes am not sure if this was sufficient. In case there are more use cases which I have not covered. Makes it easy to add/refer to additional exceptions.

docs/api.md
brdandu: This just has not been updated for a while. It is done through running `rpm run apidoc`

src-electron/generator/matter/controller/python/templates/helper.js
bzbarsky-apple: So ... I don't know why `asPythonCType` does what it does, but it's not used anywhere, right?  

I'm assuming you meant `asPythonType`.  If so: I suspect that thing is just broken.  What we should probably have here is a comment that (1) says this is being doen for compat with `asPythonType` and (2) pointing to the issue tracking this brokenness...

src-electron/generator/matter/controller/python/templates/helper.js
bzbarsky-apple: > It becomes a one line corresponding change in the template. So less questions asked in a PR:)

I'm not sure I follow.  The template change would be one line no matter what; we would just remove the boolean= and BOOLEAN= bits from the template and handle them here.

I think I am fine leaving the functionality, but it's weird to handle string types here, but not boolean types...

src-electron/db/query-bitmap.js
brdandu: I prefer cluster_ref since it is unique. Cluster code can sometimes be mixed up with mfg codes which then becomes a separate check for Zigbee pro. This handles both matter and zigbee pro. I would not change this unless really necessary.

src-electron/db/query-attribute.js
brdandu: removed

src-electron/db/query-bitmap.js
brdandu: Removed duplication

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
brdandu: Moved to java helper.js

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
brdandu: Moved to java helper.js and language option has been removed

src-electron/generator/matter/controller/python/templates/helper.js
brdandu: done

src-electron/generator/matter/controller/python/templates/helper.js
brdandu: Leaving this as is for now

src-electron/generator/matter/app/zap-templates/templates/chip/helper.js
brdandu: removed

src-electron/generator/helper-zcl.js
brdandu: id is unique so prefer using id. code is unique in Matter but not unique in zigbee pro. This way it is used in a generic way.

src-electron/generator/helper-zcl.js
brdandu: id is unique

src-electron/generator/matter/controller/python/templates/helper.js
```diff
@@ -43,6 +49,60 @@ function asPythonType(zclType) {
...
+  // Eg: {{as_underlying_python_zcl_type type [clusterId] boolean='bool'}}
```
bzbarsky-apple, line 67: Maybe I was unclear:

I am ok leaving the override functionality if we think it might be useful in the future, but we should doing the right thing for booleans without any overrides.

src-electron/generator/matter/controller/python/templates/helper.js
bzbarsky-apple: Again, we should handle booleans correctly without overrides, even if we leave the overrides in.

src-electron/generator/matter/controller/python/templates/helper.js
brdandu: ok I can change the templates for this accordingly.

src-electron/db/query-bitmap.js
brdandu: I left a note in all the helpers which are using clusterId

bzbarsky-apple (approved):

src-electron/generator/matter/controller/python/templates/helper.js
bzbarsky-apple: That's fine, but I'm not sure why we're not just making this helper do the right thing, as long as we have this helper....

PR #914, opened 2023-02-03
Added code to enable EventList support for Matter

emargolis:
- Added new method to extract events from database: `selectEndpointClusterEvents()`
 - Added new helper functions to construct list (array) of supported event IDs:
     `chip_endpoint_generated_event_count()`
     `chip_endpoint_generated_event_list()`
 - Added `eventList` and `eventCount` entries to the EmberAfCluster struct in the generated `endpoint_config.h` file

The EventList data is treated similar to CommandList data.

The assumption is that events are always `server --> client`.


src-electron/util/zcl-util.js
```diff
@@ -75,6 +75,23 @@ function commandComparator(a, b) {
...
+ * @param {*} a
+ * @param {*} b
+ * @returns -1, 0 or 1
+ */
+function eventComparator(a, b) {
+  if (a.manufacturerCode < b.manufacturerCode) return -1
+  if (a.manufacturerCode > b.manufacturerCode) return 1
+
+  if (a.hexCode < b.hexCode) return -1
+  if (a.hexCode > b.hexCode) return 1
```
bzbarsky-apple, line 90: So here, sort by code, not hexcode, once we switch to the existing shared event mapping?

test/gen-template/matter/endpoint-config.zapt
```diff
@@ -37,6 +37,13 @@
 
 #define ZAP_GENERATED_COMMANDS_INDEX(index) ((chip::CommandId *) (&generatedCommands[index]))
 
+// clang-format off
+#define GENERATED_EVENT_COUNT {{ chip_endpoint_generated_event_count }}
+#define GENERATED_EVENTS {{ chip_endpoint_generated_event_list }}
+// clang-format on
+
+#define ZAP_GENERATED_EVENTS_INDEX(index) ((chip::EventId *) (&generatedEvents[index]))
```
bzbarsky-apple, line 45: Why do we need a cast here?  I don't think we should have a cast here....  We should just have our types be actually right.  Why don't we?

bzbarsky-apple: I know we have it for commands, but we should not have it there either, either here or on the Matter side.

emargolis: Yes, it should work without casting. Just copied if from the line above for `generatedCommands`

src-electron/db/query-endpoint.js
```diff
@@ -265,6 +265,51 @@ ORDER BY C.CODE
...
+  E.EVENT_ID = ETE.EVENT_REF
+WHERE
+  E.CLUSTER_REF = ?
+  AND ETE.ENDPOINT_TYPE_REF = ?
+ORDER BY E.MANUFACTURER_CODE, E.CODE
+  `,
+    [clusterId, endpointTypeId]
+  )
+
+  return rows.map((row) => {
```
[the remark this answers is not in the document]
emargolis: updated

src-electron/util/zcl-util.js
emargolis: done

emargolis:
@bzbarsky-apple Thanks for a quick review. Your comments are addressed.

src-electron/db/query-endpoint.js
```diff
@@ -265,6 +265,40 @@ ORDER BY C.CODE
...
+async function selectEndpointClusterEvents(db, clusterId, endpointTypeId) {
+  return dbApi.dbAll(
+    db,
+    `
+SELECT
+  E.EVENT_ID,
+  E.NAME,
+  E.CODE,
+  E.MANUFACTURER_CODE,
+  E.IS_OPTIONAL
```
bzbarsky-apple, line 285: You might want to select more columns here; see the set of things that `dbMapping.map.event` maps.

bzbarsky-apple (approved):

src-electron/db/query-endpoint.js
emargolis: updated

robszewczyk (approved):

PR #918, opened 2023-02-06
[matter] Ignore values: null in readEvent response for YAML tests

vivien-apple:
##### Problem

The javascript parser annotate `readEvent` test step with `expectMultipleResponses`. I would like to do differently onto the python parser and as such I need to update `TestEvents` with:
```
response:
    - values: null
```

But the current parser will just generate codes for that that won't work. So the simplest migration path is just to ask it to ignore this type of response which has no meaning for it.


vivien-apple:
We found out an other way. Closing for now.

PR #919, opened 2023-02-08
Create token helpers for generating the token files using ZAP

brdandu:
- Updated helper-tokens with token_attributes, token_attribute_clusters and token_attribute_endpoints helpers which can be used by the generation templates.
- Deprecated tokens_context and token_next helpers which were giving incomplete and wrong generation results
- Updated some existing block helpers in helper-zcl with token information to help with generation
- Added the queries for token helpers in query-attribute and query-cluster files
- Updated dbMapping with endpointTypeClusterExtended and endpointTypeAttributeExtended such that map functions can be used in a generic way
- JIRA: ZAPP-1036

### DIFF
--- a/src-electron/db/db-mapping.js
+++ b/src-electron/db/db-mapping.js
@@ -78,6 +78,7 @@ exports.map = {
       domainName: x.DOMAIN_NAME,
       isSingleton: dbApi.fromDbBool(x.IS_SINGLETON),
       revision: x.REVISION,
+      isManufacturingSpecific: dbApi.toDbBool(x.CLUSTER_MANUFACTURER_CODE),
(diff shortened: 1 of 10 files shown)


src-electron/db/db-mapping.js
```diff
@@ -457,6 +472,41 @@ exports.map = {
     }
   },
 
+  endpointTypeAttributeExtended: (x) => {
```
tecimovic, line 475: Same here...

tecimovic (changes requested):
Some minor comments, but otherwise ok.

jperreault20 (approved):

PR #924, opened 2023-02-10
Fix a problem where custom XML in a .zap file would not load properly.

tecimovic:
- Add custom cluster config.
- Properly handle the custom package ID regardless of whether it has been loaded or not before.


tecimovic:
Note: I've added a new test file that I used. The unit test to use it will be added later.

brdandu (approved):
The changes seem good to me. The tests will be useful for the future

PR #932, opened 2023-02-15
Notification documentation

paulr34:
Adding documentation for adding a Notification


docs/notifications.md
```diff
@@ -0,0 +1,41 @@
+# Notification Development Instructions
+
+The purpose of this document is to provide ZAP developers with simple instructions on how to add a notification.
```
brdandu, line 3: Adding an image on where notifications show up will help user understand where these notifications show up in the zap ui.

docs/notifications.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Back End
+
+### Schema
+
+Within zap-schema.sql there is a table SESSION_NOTICE which is queried from the front end. SESSION_NOTICE has the following table entries:
+
+![session_notice](schema_notification.png)
+
+**SESSION_REF:** This entry is a key identifier for the current session. It is important because any warning or error that prompts a notification is only relevant to one session.
```
brdandu, line 13: This entry is a key identifier for the current session => This entry is a key identifier for a session ?

docs/notifications.md
```diff
@@ -0,0 +1,41 @@
...
+
+**NOTICE_TYPE:** This entry is used as a label for the notification. For example, for a file upgrade the type would be "UPGRADE".
```
brdandu, line 15: Are notice types pre-defined or can be anything?

docs/notifications.md
```diff
@@ -0,0 +1,41 @@
...
+
+**NOTICE_MESSAGE:** This entry is for the notification message.
+
+**NOTICE_SEVERITY:** This entry is for determining the level of severity of the notification. For example, an error that causes your project to completely fail would be higher severity while an issue that will not break your project but could cause confusion would be lower severity etc.
+
+**NOTICE_ORDER:** This entry is for organizing the entries in the "Notifications.Vue" page. **TODO:** higher priority items will have a higher order and therefore presented first or maybe with a more serious color representation like red.
```
brdandu, line 21: Remove NOTICE_ORDER from here since it is not implemented. Create an issue for this. If one exists then update that issue to include updating documentation in that ticket as well.

docs/notifications.md
```diff
@@ -0,0 +1,41 @@
...
+
+**DISPLAY:** This entry is for identifying a notification which would require a "pop up" rather than simply being shown on the Notifications page.
```
brdandu, line 23: Is this a boolean or integer?
If display is enabled for a notification then it will show as pop up on top of the zap UI apart from being shown in the notifications page.

docs/notifications.md
```diff
@@ -0,0 +1,41 @@
...
+
+### ADDING a Notification:
+
+**query-notification.js** - zap/src-electron/query-notification.js
+
+- Add the function "setNotification" located in query-notification.js to your project with the table entries described above as the parameters.
+- Within "helper functions" you can leverage "this.global.sessionID" for finding the session ID.
+- If youre adding a notification outside of the "helper functions" be careful not to add "setNotification" within a function that is used for multiple different operations. It is important to add "setNotification" where the notification is relevant and NOT for any other scenario.
```
brdandu, line 31: No need for all of the above. Just say something like, In order to add a notification, call the setNotification function in zap/src-electron/query-notification.js. setNotification takes the following arguments:
--- explain each argument

brdandu (changes requested):

docs/notifications.md
paulr34: okay

docs/notifications.md
paulr34: okay thanks

docs/notifications.md
```diff
@@ -0,0 +1,35 @@
+# Notification Development Instructions
+
+The purpose of this document is to provide ZAP developers with simple instructions on how to add a notification.
+
+## ADDING a Notification:
+
+In order to add a notification, call the setNotification function in zap/src-electron/query-notification.js. setNotification takes the following arguments:
+
+- db => database object (object)
+- type => the type of notification you wish to define (string)
```
brdandu, line 10: is there a list of existing types?

docs/notifications.md
```diff
@@ -0,0 +1,35 @@
...
+  await notification.setNotification(
+    db,
+    'UPGRADE',
+    'ISC FILE UPGRADED TO ZAP FILE. PLEASE SAVE AS TO SAVE OFF NEWLY CREATED ZAP FILE.',
+    sessionId,
+    1
+  )
+```
+
+- The front end code does not need to be changed when adding a notification.
```
brdandu, line 29: You can get rid of this comment on the front end.

brdandu (changes requested):
Get rid of add_notification file since it is no longer needed

docs/notifications.md
paulr34: okay

docs/notifications.md
paulr34: Not yet but I could add a couple to this page to make it work or delete it. I would prefer to keep it as it is a nice identifier for the user. For example, there could be a complicated error message that a user can quickly identify as a specific type of problem.

paulr34: It would sort of be like the subject to an email

brdandu (approved):

Issue #937, opened 2023-02-16
SQLITE_ERROR: cannot commit - no transaction is active


rcasallas-silabs:
Found in RaspberryPi4 Ubuntu 22.04 aarch64:

```
🤖 ZAP generation started:
    🔍 input files: /home/ricasall/git/matter/core/examples/chip-tool/third_party/connectedhomeip/src/controller/data_model/controller-clusters.zap
    🔍 output pattern: /home/ricasall/git/matter/core/out/chip-tool/gen/third_party/connectedhomeip/src/controller/data_model/zap_pregen/zap-generated
    🔍 using templates: /home/ricasall/git/matter/core/examples/chip-tool/third_party/connectedhomeip/src/app/zap-templates/app-templates.json
    🔍 using zcl data: /home/ricasall/git/matter/core/examples/chip-tool/third_party/connectedhomeip/src/app/zap-templates/zcl/zcl.json
    🔍 zap version: ver. 2023.1.23, featureLevel 89, commit: 7f47fc19f39a4e1af038bbfd9c059b69bb93345b from 2023-01-23T18:17:00.000Z, mode: source, exe: node
🕐 Setup time: 6s 334ms
👉 using input file: /home/ricasall/git/matter/core/examples/chip-tool/third_party/connectedhomeip/src/controller/data_model/controller-clusters.zap
[Error: SQLITE_ERROR: cannot commit - no transaction is active] {
  errno: 1,
  code: 'SQLITE_ERROR'
}
⇝ Zap generation error: Error: SQLITE_ERROR: cannot commit - no transaction is active
👎 Program npx exited with error code: 1
```

paulr34:
@rcasallas-silabs I was able to reproduce this with an older commit but I cannot reproduce it with the latest commit. Based on that, would you mind if I closed this?

PR #938, opened 2023-02-16
Fixes ##937: SQLITE_ERROR: cannot commit - no transaction is active.

rcasallas-silabs:
Fix for [#937](https://github.com/project-chip/zap/issues/937)

### DIFF
--- a/src-electron/db/db-api.js
+++ b/src-electron/db/db-api.js
@@ -100,16 +100,18 @@ async function dbBeginTransaction(db) {
  */
 async function dbCommit(db) {
   return new Promise((resolve, reject) => {
-    db.run('COMMIT', [], function (err) {
-      if (err) {
-        env.logError('Failed to COMMIT')
-        reject(err)
-      } else {
-        env.logSql('Executed COMMIT')
-        inTransaction = false
-        resolve()
-      }
-    })
+    if(inTransaction) {
+      db.run('COMMIT', [], function (err) {
+        if (err) {
+          env.logError('Failed to COMMIT')
+          reject(err)
+        } else {
+          env.logSql('Executed COMMIT')
+          inTransaction = false
+          resolve()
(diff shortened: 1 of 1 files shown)


Shirelein (approved):

tecimovic:
@rcasallas-silabs : I ran into one of these recently, and the problem was that if a `ZAP` file that you're reading in was malformed (not a valid JSON), the `try { ... } finally` was incorrect.

If you look at this piece of code in `import.js`:
```
async function importDataFromFile(
  db,
  filePath,
  options = {
    sessionId: null,
    defaultZclMetafile: env.builtinSilabsZclMetafile(),
    postImportScript: null,
    packageMatch: dbEnum.packageMatch.fuzzy,
  }
) {
  let state = await readDataFromFile(filePath, options.defaultZclMetafile)
  try {
    await dbApi.dbBeginTransaction(db)
    let sid
    if (options.sessionId == null) {
      sid = await querySession.createBlankSession(db)
    } else {
      sid = options.sessionId
    }
    let loaderResult = await state.loader(db, state, sid, options.packageMatch)
    if (options.postImportScript != null) {
      await executePostImportScript(
        db,
        loaderResult.sessionId,
        options.postImportScript
      )
    }
    return loaderResult
  } finally {
    await dbApi.dbCommit(db)
  }
}

```

Before my change, the line that does `readDataFromFile` was already inside the `try { ...` and then if it failed, the `finally` would trigger before even getting to the "dbBeginTransaction()".

If this is the same problem, then I would not merge your change in. Whenever we have these transaction missmatches, I'd prefer to hunt them down and clean them properly, rather than just "patch" a problem by doing what you did.

tecimovic (changes requested):
Check if this is due to the malformed .json file, if so, then that was fixed elsewhere.

PR #944, opened 2023-02-22
Fix backwards compatibility for recent zap changes

brdandu:
- Throwing a warning in the notifications page of the zap UI instead of throwing an error in template-engine such that we backwards compatibility can be maintained
- Changes to helper.js files have been made to fall back to old way of implementation in the case of global.resource([key]) not finding the data required
- Github: ZAP#941, ZAP#942

### DIFF
--- a/src-electron/generator/matter/app/zap-templates/templates/app/helper.js
+++ b/src-electron/generator/matter/app/zap-templates/templates/app/helper.js
@@ -49,14 +49,15 @@ function getConfigData(global) {
   if (configData === undefined) {
     let f = global.resource('config-data');
     // NOTE: This has to be sync, so we can use this data in if conditions.
-    let rawData = fs.readFileSync(f, { encoding: 'utf8', flag: 'r' });
-    configData = YAML.parse(rawData);
+    if (f) {
+      let rawData = fs.readFileSync(f, { encoding: 'utf8', flag: 'r' });
+      configData = YAML.parse(rawData);
+    }
   }
   return configData;
 }
 
-function isInConfigList(string, listName)
-{
+function isInConfigList(string, listName) {
   const data = getConfigData(this.global);
   return data[listName].includes(string);
 }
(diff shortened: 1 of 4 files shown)


src-electron/generator/matter/app/zap-templates/templates/app/helper.js
```diff
@@ -828,6 +882,72 @@ function getPythonFieldDefault(type, options) {
   return _getPythonFieldDefault.call(this, type, options);
 }
 
+// Allow-list of enums that we generate as enums, not enum classes.  The goal is
+// to drive this down to 0.
```
bzbarsky-apple, line 886: That is most certainly not the goal for _this_ list.  The goal here is to not touch it.

And I am not sure how you want to deal with the fact that historically this list had entries removed from it....  not sure how far back you want your backwards compat to go.

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -362,8 +363,14 @@ function fetchAvailabilityData(global) {
...
     let f = global.resource('availability-data');
     // NOTE: This has to be sync, so we can use this data in if conditions.
-    let rawData = fs.readFileSync(f, { encoding: 'utf8', flag: 'r' });
-    availabilityData = YAML.parse(rawData);
+    if (f) {
+      let rawData = fs.readFileSync(f, { encoding: 'utf8', flag: 'r' });
+      availabilityData = YAML.parse(rawData);
+    } else {
+      throw new Error(
+        `Resource availability-data not found among the context resources. Check your template.json file.`
```
bzbarsky-apple, line 371: Is this just because the exception is nicer than what `readFileSync` throws if `f` is None?

brdandu: I was thinking of catching it before it even goes ahead anymore.

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
brdandu: :( . May be just start with this set and if there are any new issues which come up then we re-edit this list again.

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
```diff
@@ -102,17 +151,20 @@ function chip_endpoint_generated_functions() {
         hasFunctionArray = true;
       }
 
-      if (configData.ClustersWithShutdownFunctions.includes(clusterName)) {
+      if (
+        configData &&
+        configData.ClustersWithShutdownFunctions.includes(clusterName)
```
bzbarsky-apple, line 156: Note that there was a period when `configData` existed but `configData.ClustersWithShutdownFunctions` did not.  If we need to be compatible with SDKs from that period, we should have a check here and down below.

bzbarsky-apple (approved):

PR #945, opened 2023-02-22
Deal with ordering in the endpoint-config generated structs.

tecimovic:


brdandu (approved):

src-electron/generator/helper-endpointconfig.js
```diff
@@ -418,7 +457,38 @@ function endpoint_reporting_config_defaults(options) {
...
+          break
+        case 'mask':
+          items.push(mask)
+          break
+        case 'mfgCode':
+          items.push(r.mfgCode)
+          break
+        case 'minmax':
+          items.push(
+            `{{ ${r.minOrSource}, ${r.maxOrEndpoint}, ${r.reportableChangeOrTimeout} }}`
```
bzbarsky-apple, line 484: What happens if the minmax struct changes?

tecimovic: Good point. I added minmaxorder to the helper for this case.

PR #948, opened 2023-02-24
Restore matter unit tests and fix backwards compatibility.

tecimovic:


tecimovic:
This work also uncovered the fact that `gen-matter2.test.js` and `gen-matter3.test.js` were largely useless as unit tests. I dabbled a bit initially (hence some commits to the gen-matter3.test.js and the zapt files), but I'll first fix a functional issue here to unblock some people that are trying to release something, and then we can address that in a separate work.

test/gen-template/matter3/t.json
```diff
@@ -96,11 +96,6 @@
       "path": "16.zapt",
       "name": "Test 16",
       "output": "16.out"
-    },
-    {
-      "path": "17.zapt",
-      "name": "Test 17",
-      "output": "17.out"
```
brdandu, line 103: why is this removed?

brdandu (approved):

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
```diff
@@ -352,18 +358,51 @@ function chip_endpoint_cluster_list() {
...
+    let individualItems = [];
+    let orderTokens = order.split(',').map((x) => (x ? x.trim() : ''));
+    orderTokens.forEach((tok) => {
+      switch (tok) {
+        case 'clusterId':
+          individualItems.push(`.clusterId = ${c.clusterId}`);
+          break;
+        case 'attributes':
+          individualItems.push(
+            `.attributes = ZAP_ATTRIBUTE_INDEX(${c.attributeIndex})`
```
bzbarsky-apple, line 370: The whitespace here is different from the original code....  might be OK if the consumer runs something like clang-format anyway, or might not be great for readability if not.    @tecimovic

bzbarsky-apple: This turns out to not be OK on the Matter side; it's causing undesired changes in header outputs.

bzbarsky-apple: Created https://github.com/project-chip/zap/pull/955

PR #958, opened 2023-02-27
Add some more content-based matter tests.

tecimovic:


brdandu:
Other templates to consider https://github.com/project-chip/connectedhomeip/tree/master/src/app/zap-templates/templates/app

test/gen-matter-4.test.js
```diff
@@ -83,6 +83,80 @@ test(
       { disableDeprecationWarnings: true }
     )
     expect(genResult.hasErrors).toEqual(false)
+
+    // Note: A lot of these tests here are for sake of backwards
+    // compatibility. Latest ZAP must be able to generate content for
+    // the old SDKs, so if you changed something that generates
+    // endpoint_config differently, please be very very careful and
+    // make yourself you can answer positively the following question:
```
bzbarsky-apple, line 91:
```suggestion
    // make sure you can answer positively the following question:
```

test/gen-matter-4.test.js
```diff
@@ -83,6 +83,80 @@ test(
...
+    //   will after my changes, zap still be able to generate content
+    //   that will work with an older SDK.
+    //
+    let ept = genResult.content['endpoint_config.h']
+    expect(ept).toContain(`{ \\
+      /* Endpoint: 65534, Cluster: Network Commissioning (server) */ \\
```
bzbarsky-apple, line 97: Shouldn't we also have some tests for non-default `order` values?  Here and in other places.

test/gen-matter-4.test.js
tecimovic: Right....

test/gen-matter-4.test.js
tecimovic: I'll add some tests for that `order="...."` into the template.

PR #964, opened 2023-03-06
Use MTR_DEPRECATED in Darwin availability helper.

ksperling-apple:
See project-chip/connectedhomeip#24994

### DIFF
--- a/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
+++ b/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
@@ -734,7 +734,7 @@ async function availability(clusterName, options) {
   let availabilityStrings = Object.entries(introducedVersions).map(
     ([os, version]) => `${os}(${version}, ${deprecatedVersions[os]})`
   );
-  return `API_DEPRECATED("${
+  return `MTR_DEPRECATED("${
     options.hash.deprecationMessage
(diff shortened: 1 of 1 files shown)


ksperling-apple:
The Matter build failure due to uncommitted changes is expected, these changes will be committed in the Matter repo alongside adopting the next zap version.

bzbarsky-apple (approved):

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -734,7 +734,7 @@ async function availability(clusterName, options) {
   let availabilityStrings = Object.entries(introducedVersions).map(
     ([os, version]) => `${os}(${version}, ${deprecatedVersions[os]})`
   );
-  return `API_DEPRECATED("${
```
brdandu, line 737: Do we need to worry about backward's compatibility here?

bzbarsky-apple: In practice, not really.  I don't think anyone will be using new ZAP to regenerate the Darwin APIs on top of an old SDK revision....

bzbarsky-apple: Yes, having the compatibility check only check app regen would make a lot of sense.

PR #984, opened 2023-03-29
Fixing App.vue so it is asynchronous 

paulr34:
Changing App.vue so Store references including rest API calls are asynchronous.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -47,33 +47,32 @@ async function initializeSession() {
 
 async function initLoad(store) {
   await initializeSession()
-
-  store.dispatch('zap/loadInitialData')
-  store.dispatch('zap/loadOptions', {
+  let promises = []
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -47,33 +47,32 @@ async function initializeSession() {
...
-  })
-  store.dispatch('zap/loadOptions', {
+  })))
+  promises.push(store.dispatch(('zap/loadOptions', {
     key: 'generator',
     type: 'object',
-  })
-  store.dispatch('zap/loadSessionKeyValues')
+  })))
+  promises.push(store.dispatch(('zap/loadSessionKeyValues')))
```
brdandu, line 68: I am not sure if this needs to happen before the rest but either way I am not sure why these were not part of the promises.

brdandu (approved):

PR #985, opened 2023-03-30
Extend all_user_cluster_generated_commands with more info

brdandu:
Extending the all_user_cluster_generated_commands with more information on command arguments

Github: ZAP#971


src-electron/db/query-command.js
```diff
@@ -139,6 +139,12 @@ async function selectAllAvailableClusterCommandDetailsFromEndpointTypes(
       incoming: x.INCOMING,
       outgoing: x.OUTGOING,
       mfgCommandCount: x.MANUFACTURING_SPECIFIC_COMMAND_COUNT,
+      hasSpecificResponse: dbApi.toDbBool(x.RESPONSE_REF),
```
bzbarsky-apple, line 142: I really wish this used our normal command mapping, so we don't keep getting skew like this....

bzbarsky-apple (approved):

src-electron/db/query-command.js
brdandu: I was going for that and saw some mismatches which made me want to do the map cleanup later. I will log a bug a cleanup a lot of these surplus maps and make it all go to db-mapping

PR #990, opened 2023-04-06
Uccli upgrade fix

paulr34:
When Upgrading a Zigbee or Matter App there is one error message 

```Failed to run upgrade apack_zclConfigurator UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "You need to specify at least one zap file."```

This message is not because of the ZAP upgrade process but rather because the ZAP source code checks the .ZAP file for upgrades before finding the .ZAP file in its folder. The result is that even if there is a .zap file in the config folder, the user will be told they need to specify a .zap file.

In order to fix this, changed the order of operations so the code gathers the files before checking the files.

After making this change my tests passed. The files completed their upgrade.

### DIFF
--- a/src-electron/main-process/startup.js
+++ b/src-electron/main-process/startup.js
@@ -177,6 +177,7 @@ function gatherFiles(filesArg, options = { suffix: '.zap', doBlank: true }) {
  * @param {*} output
  */
 async function startConvert(argv, options) {
+  let zapFiles = argv.zapFiles
   let files = gatherFiles(zapFiles, { suffix: '.zap', doBlank: true })
   if (files.length == 0) {
     options.logger(`    👎 no zap files found in: ${zapFiles}`)
--- a/src-electron/main-process/startup.js
+++ b/src-electron/main-process/startup.js
@@ -183,6 +183,7 @@ async function startConvert(argv, options) {
     options.logger(`    👎 no zap files found in: ${zapFiles}`)
     throw `👎 no zap files found in: ${zapFiles}`
   }
+  if (argv.output == null) throw 'You need to specify output file.'
   let output = argv.output
   let conversion_results = argv.results
   options.logger(`🤖 Conversion started
--- a/src-electron/main-process/startup.js
+++ b/src-electron/main-process/startup.js
@@ -177,7 +177,11 @@ function gatherFiles(filesArg, options = { suffix: '.zap', doBlank: true }) {
  * @param {*} output
  */
 async function startConvert(argv, options) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src-electron/main-process/startup.js [hunk 1]
```diff
@@ -780,9 +785,6 @@ async function startUpMainInstance(argv, callbacks) {
   } else if (argv._.includes('server')) {
     return startServer(argv, quitFunction)
   } else if (argv._.includes('convert')) {
-    if (argv.zapFiles.length < 1)
-      throw 'You need to specify at least one zap file.'
-    if (argv.output == null) throw 'You need to specify output file.'
```
tecimovic, line 785: You have deleted the `if ( argv.output == null )` check. There seems to be no replacement for it. Is that ok?

src-electron/main-process/startup.js
```diff
@@ -177,7 +177,12 @@ function gatherFiles(filesArg, options = { suffix: '.zap', doBlank: true }) {
...
  */
 async function startConvert(argv, options) {
-  let files = argv.zapFiles
+  let zapFiles = argv.zapFiles
+  let files = gatherFiles(zapFiles, { suffix: '.zap', doBlank: true })
+  if (files.length == 0) {
+    options.logger(`    👎 no zap files found in: ${zapFiles}`)
+    throw `👎 no zap files found in: ${zapFiles}`
+  }
   let output = argv.output
```
tecimovic, line 186: The argv.output was checked in the caller. Maybe it should be checked here now?

tecimovic (changes requested):
There is an odd logic there with the `argv.output`. How does the startConvert() now assign the outputs if there are multiple files?

src-electron/main-process/startup.js
paulr34: To be consistent with the spec, the check should kept, so I will add it back in.

src-electron/main-process/startup.js
paulr34: Yes, I will add it to startConvert(). I am not sure what you mean about multiple files. The only change I made was deleting the check which simply makes sure the user did not forget to enter a required argument. Any other logic is unchanged. I guess we could add another check to make sure they did not add multiple files as an argument?

src-electron/main-process/startup.js
paulr34: done

tecimovic (approved):

PR #1000, opened 2023-04-18
Add isSupported helper for Darwin.

bzbarsky-apple:
This supports marking this as "provisional", which unlike "removed" can be overridden by a later "introduced".

### DIFF
--- a/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
+++ b/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
@@ -741,20 +741,19 @@ async function availability(clusterName, options) {
 
 /**
  * Utility for wasIntroducedBeforeRelease and isSupported.  Returns undefined if
- * the the thing we are looking at was not officially introduced, otherwise
+ * the the path we are looking at was not officially introduced, otherwise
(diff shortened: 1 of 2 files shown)


src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -739,42 +739,147 @@ async function availability(clusterName, options) {
...
+    if (path1[i] != path2[i]) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+function isSupported(cluster, options) {
+  if (wasRemoved.call(this, cluster, options)) {
```
ksperling-apple, line 845: Is it necessary to call this with `.call` ?

ksperling-apple (approved):

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
bzbarsky-apple: Yes, because it uses `this` to get `this.global`.  I could rejigger it to call some helper function that takes the global as an arg, and call it from there and here, but this was simpler for now.

tecimovic (approved):

PR #1002, opened 2023-04-20
Adding struct item helper based on struct and cluster name

brdandu:
- Add zcl_struct_items_by_struct_and_cluster_name helper for retrieving struct items based on struct and cluster name in helper-zcl.js
- Updating selectAllStructItemsByStructName to handle cluster name as well
- Updating the xml for TargetStruct such that it is now present in binding and access control clusters
- Adding a test to make sure the right struct items are generated based on cluster names
- Github: ZAP#916

### DIFF
--- a/src-electron/generator/helper-zcl.js
+++ b/src-electron/generator/helper-zcl.js
@@ -297,6 +297,32 @@ async function zcl_struct_items_by_struct_name(name, options) {
 }
 
 /**
+ * Block helper iterating over all struct items given the struct name and
+ * cluster name.
+ *
+ * @param name
+ * @param clusterName
+ * @param options
+ * @returns Promise of content.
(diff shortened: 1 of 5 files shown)


src-electron/generator/helper-zcl.js
```diff
@@ -296,6 +296,32 @@ async function zcl_struct_items_by_struct_name(name, options) {
   return templateUtil.templatePromise(this.global, promise)
 }
 
+/**
+ * Block helper iterating over all struct items given the struct and cluster
```
bzbarsky-apple, line 300:
```suggestion
 * Block helper iterating over all struct items given the struct name and cluster
```

bzbarsky-apple (approved):
Thank you!

PR #1004, opened 2023-04-21
Implement the new layout

tbrkollar:
This is a draft pr for ui testing.
I have been working on cypress tests.


tecimovic:
Thanks for putting this out there, @tbrkollar ! We'll review.

tecimovic:
One comment about the matter and zigbee icons....

We've gotten into trouble in the past because we put zigbee icons onto  appbuilder. It was not legal. We had no legal right to use those icons. I don't know how matter icons are treated, but I suspect it's similar? Not anyone can just slam those icons onto anything they want.

So the way we need to treat icons is as follows:
   1.) Zap should by default NOT include either zigbee or matter logos. It should only have it's own ZAP logo.
   2.) `zcl.json` file on the SDK can say: `logo: my-logo.png` or whatever the format is.
   3.) When server reads this, and loads it into the database, it allows the UI to fetch this image and display it in the UI.
   
So Zigbee would say: `logo: "zigbee.png"` and add this PNG into a directory next to the zcl.json. Matter would do similar. And zap is then just a pass-through mechanism and doesn't contain these images itself.

brdandu (changes requested):
- [ ] WHen preview pane is open, highlight it to show that it is open so that the user knows that they have to click it again to close the preview pane. Need to do the same with all the other top level menu bars such as settings and notifications.
- [ ] Generate, extensions and options in the toolbar menu open a new pane where as preview, settings and notifications are an update to the existing pages. We should keep the UI to be consistent across these.
- [ ] Get rid of regenerate button? This is not used anywhere
- [ ] Hide tutorial button until it is in working condition
- [ ] In the endpoint UI, can we get rid of Enabled clusters, enabled attributes and enabled reporting? This is pointless, reports wrong count and will save us space. See attached. Screen Shot 2023-05-08 at 11 51 26 AM
- [ ] For cluster selection, Switch drop down of 'Enable' to check boxes for:  client, server 
Screen Shot 2023-05-08 at 11 52 20 AM
- [ ] Some of the cluster selections is greyed out. We normally consider all clusters as customizable even after the zcl device type selection. Not sure how these became greyed out. Also having an 'Enable all clusters' button at the bottom is counter intuitive See Screen Shot 2023-05-08 at 11 56 04 AM    
  
- [ ] Shouldn't the following change to open all once close all has taken effect: 
Screen Shot 2023-05-08 at 11 57 34 AM      

- [ ] Can we persist the 'close all' button with the filter setting? Everytime I do close all, the filter setting resets.
- [ ] Also when I search something in the cluster search and press 'close all' nothing happens. Perhaps grey it out when something is being searched

paulr34 (approved):

brdandu (approved):
Make sure to squash before merge

paulr34 (approved):

PR #1006, opened 2023-04-25
Resolve issues related to multiple global attributes instances showing in the ZAP UI

brdandu:
- This happens when the matter/zigbee sdks are updated and the same .zap sample app file is opened using zap.
- Updating the attribute tables constraints such that attributes are unique based on cluster_ref, package_ref, code, mfg code and side
-  Added CLUSTER_REF_NOT_NULL and MANUFACTURER_CODE_NOT_NULL to the table because the unique constraint does not work correctly with null values in sqlite. Therefore adding an empty string to them and using those in the unique constraint
- In query-loader#INSERT_ATTRIBUTE_QUERY using ignore such that duplicate attribute insertions are ignored.
- In query-impexp#importAttributeForEndpointType ignoring insert if an endpoint type attribute already exists based on UNIQUE(ENDPOINT_TYPE_REF, ATTRIBUTE_REF, ENDPOINT_TYPE_CLUSTER_REF ). The ZAP UI was running into this when SDK was updated and the saved sample .zap file was opened again.
- JIRA: ZAPP-1079


src-electron/importexport/import-json.js
```diff
@@ -438,6 +439,29 @@ async function jsonDataLoader(db, state, sessionId, packageMatch) {
...
+      if (pkg.type == 'zcl-xml-standalone') {
+        let info = await queryPackage.getPackageByPackageId(db, packageId)
+        let fileContent = await fsp.readFile(info.path)
+        let actualCrc = util.checksum(fileContent)
+        if (info != null && actualCrc != info.crc) {
+          env.logDebug(
+            `CRC missmatch for file ${pkg.path}, (${pkg.crc} vs ${actualCrc}) package id ${pkg.id}, parsing.
+            Deleting package id: ${packageId}`
+          )
+          await queryPackage.deletePackagesByPackageId(db, packageId)
```
tecimovic, line 457: You can't just delete this package. There might have been another session opened which uses this package.

tecimovic: I suggest you load this new package as well, and then make the session use this new package, not the old package. There is no problems if you have multiple standalone packages.

src-electron/importexport/import-json.js
```diff
@@ -438,6 +439,29 @@ async function jsonDataLoader(db, state, sessionId, packageMatch) {
...
         pkg.version
       )
+      // Checking if standalone xml file content changed by checking for a crc
+      // mismatch. If there is a crc mismatch then delete the package so that
+      // it does not come under the list of existing packages.
+      // This check makes sure that if a custom xml is updated and the zap file
+      // is re-opened then the custom xml is reloaded to reflect the correct
+      // configuration.
+      if (pkg.type == 'zcl-xml-standalone') {
+        let info = await queryPackage.getPackageByPackageId(db, packageId)
```
tecimovic, line 449: I thought we have a function for this somewhere. util.readFileContentsAndCrc() or something?

src-electron/zcl/zcl-loader.js
```diff
@@ -195,7 +195,18 @@ async function qualifyZclFile(
       env.logDebug(
         `CRC missmatch for file ${pkg.path}, (${pkg.crc} vs ${actualCrc}) package id ${pkg.id}, parsing.`
       )
-      await queryPackage.updatePathCrc(db, filePath, actualCrc, parentPackageId)
+      // Reloading the zcl xml file when there is a CRC mismatch(changes) on it.
+      // This check makes sure that when a custom xml is edited, deleted and
+      // re-added from ZCL Extensions in ZAP UI then the custom xml package is
+      // updated in the backend to reflect the right configuration.
+      await queryPackage.deletePackagesByPackageId(db, pkg.id)
```
tecimovic, line 202: This deleting of packages still sounds dangerous. How do you know there isn't another session using this package?

tecimovic (changes requested):

src-electron/zcl/zcl-loader.js
brdandu: This is done when the zcl loading happens. Is it possible to have loading happen when an existing session is still in progress?

brdandu: Sorry ignore me. This is the custom xml through zcl extension. I see your point.

paulr34:
@brdandu this bug is fixed so this PR is not needed, please reopen if I am mistaken! thanks!

PR #1008, opened 2023-04-27
Expose the storagePolicy of attributes in attribute iterators.

bzbarsky-apple:


src-electron/db/db-mapping.js
```diff
@@ -118,6 +118,7 @@ exports.map = {
         x.REPORTING_POLICY == dbEnums.reportingPolicy.mandatory ||
         x.REPORTING_POLICY == dbEnums.reportingPolicy.suggested,
       reportingPolicy: x.REPORTING_POLICY,
+      storagePolicy: x.STORAGE_POLICY,
```
brdandu, line 121: What is the difference between STORAGE_OPTION and STORAGE_POLICY?

bzbarsky-apple: STORAGE_POLICY is attached to the attribute definition, independent of cluster/endpoint configuration, and defined by the SDK that limits what STORAGE_OPTION values are allowed.

STORAGE_OPTION is the actual storage option selected in ZAP for the attribute.

For example STORAGE_POLICY can be "any", and then STORAGE_OPTION can be one of "RAM", "NVM", or "External".

Or STORAGE_POLICY can be "attributeAccessInterface", and then STORAGE_OPTION has to be "External".

brdandu: Are the above constraints applied in some form or way in our code? I am thinking if we should add a database constraint for this in the attribute table such that an error is thrown when the above constraints are not met.

bzbarsky-apple: > Are the above constraints applied in some form or way in our code?

In `importAttributeForEndpointType` we have:
```
  if (storagePolicy == dbEnums.storagePolicy.attributeAccessInterface) {
    attribute.storageOption = dbEnums.storageOption.external
  }
```

brdandu (approved):

PR #1016, opened 2023-05-09
Retrieving new description when switching packages

paulr34:
Getting new description 
Screenshot 2023-05-09 at 1 42 29 PM

### DIFF
--- a/src/layouts/ZclConfiguratorLayout.vue
+++ b/src/layouts/ZclConfiguratorLayout.vue
@@ -36,7 +36,7 @@ limitations under the License.
           </q-toolbar-title>
           <q-toolbar-title v-on:click.ctrl="showVersion" v-else>
             Cluster Configurator:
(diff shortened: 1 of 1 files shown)


brdandu (changes requested):
Load 2 packages and check if it is still populating correctly for eg load matter and zigbee into the db and select things from the launcher tab

paulr34:
> Load 2 packages and check if it is still populating correctly for eg load matter and zigbee into the db and select things from the launcher tab

Both packages always load but I did open ZAP linked to Zigbee then closed and opened ZAP linked to Matter. Everything seems to work.

Without this change it says "ZigbeePro Test Data" rather than "Matter SDK ZCL Data" when you open ZAP with Matter linked after opening ZAP with Zigbee linked. This change retrieves the updated description rather than keeping the original one.

brdandu (approved):

PR #1034, opened 2023-05-19
Support multiple device type with primary type

aspecter:


paulr34:
@brdandu @tecimovic @aspecter what about Zigbee?

paulr34:
Primary device types are defined, right? How would the GUI handle this? How would a user know which device types to include in the subset of the superset? They would look at the spec? 

Is the plan to allow any device type combo and just let the user mess up?

PR #1035, opened 2023-05-19
Stop relying on the input .zap file in Matter test codegen.

bzbarsky-apple:
To keep backwards compat with old Matter revisions, we require explicit includeAllClusters=true in the templates to get the new behavior.


andy31415 (approved):

brdandu (approved):
Assuming these changes are just for backwards compatibility and not being used by any of our latest application templates.

bzbarsky-apple:
> Assuming these changes are just for backwards compatibility and not being used by any of our latest application templates.

Our latest application templates would be setting `includeAllClusters=true`.  The support for `includeAllClusters=false` (or unset) is for backwards compat.

PR #1040, opened 2023-05-24
Feature/add multiple device types per endpoint/zap#862

brdandu:


bzbarsky-apple:
@brdandu This will need endpoint_config changes on top of this, right?

brdandu:
> @brdandu This will need endpoint_config changes on top of this, right?

Need to figure out how we want to update endpoint_config with respect to this.

bzbarsky-apple:
> Need to figure out how we want to update endpoint_config with respect to this.

Basically, this bit:

https://github.com/project-chip/zap/blob/6f58a5fcb07a9bf52851f6e1aa8b4372762326fc/src-electron/generator/helper-endpointconfig.js#L661-L666

will need to get the actual list of device types and their versions.  Everything else should Just Work, I believe.

tecimovic (approved):

brdandu:
The zigbee code generation workflow is now producing the correct result.

bzbarsky-apple:
As far as I can tell, endpoint_config.h codegen for Matter is completely broken with this PR as soon as multiple device types are defined on an endpoint.  Steps to reproduce:

1.  Check out Matter.
2. Check out ZAP with these changes, make sure that's the ZAP Matter is using.
3. Run `scripts/codepregen.py --input-glob "*all-clusters-app*"  ./zzz_pregenerated1` in the Matter repository.
4. Run `scripts/tools/zap/run_zaptool.sh examples/all-clusters-app/all-clusters-common/all-clusters-app.zap` int the Matter repository.
5. Select endpoint 1 in the ZAP tool.
6. Click "edit".
7. Add "Matter Dimmable Light" to the "Device" field, so it now has both On/Off Light and Dimmable Light.
8. Click Save
9. Save the ZAP config from the ZAP tool.
10. In the Matter repo, run `scripts/codepregen.py --input-glob "*all-clusters-app*"  ./zzz_pregenerated2`
11. Compare `zzz_pregenerated1/examples/all-clusters-app/all-clusters-common/all-clusters-app/zap/app-templates/zap-generated/endpoint_config.h` to `zzz_pregenerated2/examples/all-clusters-app/all-clusters-common/all-clusters-app/zap/app-templates/zap-generated/endpoint_config.h`

I am attaching the diff I observer here.
[endpoint-diff.txt](https://github.com/project-chip/zap/files/12053573/endpoint-diff.txt)

Note that this has  _5_ fixed endpoints defined, not 4, with all the attributes/clusters from endpoint 1 listed twice.

brdandu:
The zigbee code generation diff failures are generating the correct code now with the correct device version and device id .

PR #1044, opened 2023-05-30
loading initial data after renderer finishes

paulr34:
Attempting to load initial data after API renderer finishes loading new browser/session


paulr34:
redesigning rather than fixing https://github.com/project-chip/zap/discussions/1050

Issue #1042, opened 2023-05-30
[feature request] allow generation of multiple files for a single template


andy31415:
We are looking for developers working on one cluster not conflicting/interacting with developers working on different and unrelated clusters.

Current code generation will often generate a single large file containing all clusters (e.g. serialization, pretty-printing, callback dispatch etc) and this forces serialization of xml merging and every one conflicts with others (and often results in merge conflicts).

It would be great if we could generate per-cluster  outputs:
  - no more merge conflicts (or significantly reduced conflicts)
  - code would look cleaner by `#include <OnOffCluster.h>` instead of `#include <AllClusters.h>`
  - likely better modularity overall, less code would rely on "everyting is visible and available"

PR #1046, opened 2023-05-30
Add iterative templates.

tecimovic:
This is addressing: 
https://github.com/project-chip/zap/issues/1042

Iterative templates, are the template files that have a name like something_{id}.out, for example,
and a property "iterator" set on the individual file.

Depending on iterator (for example "available_clusters") the generator will iterate over all the 
values of an interator, and generate a separate file for each one, substituting {id} with a given value.

There will be different values for "iterator" and for the {pattern} thingies in the output file names.


src-electron/db/query-endpoint-type.js
```diff
@@ -165,14 +165,16 @@ async function selectEndpointType(db, id) {
 async function selectAllClustersDetailsFromEndpointTypes(
   db,
   endpointTypes,
-  options = null
```
brdandu, line 168: Passing options allows other flags coming from the handlebar helper. This helps in not updating the function signature.

brdandu (approved):

paulr34 (approved):

src-electron/db/query-endpoint-type.js
tecimovic: The intent is valid. But the options were passed in a handlebars specific way (options.hash) and I needed to call this functions out of template context. I don't want db layer to be polluted by handlebars specifics.

test/resource/meta/gen-test.json
```diff
@@ -33,6 +37,12 @@
       "path": "access.zapt",
       "name": "Access data example",
       "output": "access.out"
+    },
+    {
+      "path": "cluster_iterator.zapt",
+      "name": "Cluster iterator.",
+      "output": "cluster_out_{define:tolowercase}.out",
+      "iterator": "selectedCluster"
```
bzbarsky-apple, line 45: How does one discover valid "iterator" values, @tecimovic ?  The docs talk about "availableClusters" as a value, but dbenum does not seem to match that, and nothing short of reading the code will discover "selectedCluster"....

tecimovic: I need to update sdk-integration.md doc....

Issue #996, opened 2023-04-17
Changing device type breaks attribute reportability


bzbarsky-apple:
STEPS TO REPRODUCE:

1. In the Matter repo (as of SHA `533f84c9fcce035dd23167128fc89a6901c79274`), run `./scripts/tools/zap/run_zaptool.sh examples/lighting-app/lighting-common/lighting-app.zap`
2. Examine the "Attribute Reporting" view of the "On/Off" cluster on endpoint 1, and see that all attributes are set as reportable.
3. Edit endpoint 1 and change the device type from "Matter Dimmable Light (0x0101)" to "Matter Extended Color Light (0x010D)".  Click the "SAVE" button.
4. Examine the "Attribute Reporting" view of the "On/Off" cluster on endpoint 1, and see that all the non-global attributes are no longer set as reportable.

Similarly, if I just add a new endpoint with device type "Matter Extended Color Light (0x010D)" then all the "On/Off" non-global attributes are set to not reportable in the new endpoint's automatically-enabled "On/Off" cluster.

For comparison, if I create an endpoint with device type "Matter Door Lock (0x000A)" and then enable the "On/Off" cluster (not enabled by default) on that endpoint, then the non-global attributes are correctly (due to Matter SDK defaults) marked as reportable.

@brdandu @tecimovic

paulr34:
![Screenshot 2023-04-24 at 5 27 52 PM](https://user-images.githubusercontent.com/64710345/234120904-f2305ce9-1f8c-4150-8f8f-dd3618de8343.png)
 ^Above is the On/Off cluster for **MA-extendedcolorlight** and below is the On/Off for **Matter On/Off Light Switch**
![Screenshot 2023-04-24 at 5 30 11 PM](https://user-images.githubusercontent.com/64710345/234121212-c3d43281-ad80-4338-ac86-b8d893c3958d.png)

This further leads me to believe that the problem is the device type configuration in this xml file. 

Additionally, I did find an example of non global attribute reporting being enabled by default. Note that it is "DEVICE_TYPE". I am still investigating why that happens. 
<img width="1447" alt="Screenshot 2023-04-24 at 4 13 59 PM" src="https://user-images.githubusercontent.com/64710345/234121463-4475938a-dba8-40e9-8cdb-eb5ab8ce00b0.png">

Any feedback is welcome

bzbarsky-apple:
> because for it to be enabled by default it must be specified in that XML file.

[...]

PR #1058, opened 2023-06-15
Fix button state issue in attribute reporting page

YanchengZhou:
Fix issue #996 
The issue is about when initializing the attribute reporting page, some attributes should have their enabled button turned on, but they are actually turned off.
After the fix, if an attribute has mandatory or suggested reporting policy, then its radio button will be turned on automatically when the page is initialized.

### DIFF
--- a/src/components/ZclAttributeReportingManager.vue
+++ b/src/components/ZclAttributeReportingManager.vue
@@ -247,6 +247,9 @@ export default {
       },
     },
   },
+  created() {
+    this.updateSelectedReporting()
+  },
   data() {
     return {
       pagination: {
@@ -331,6 +334,7 @@ export default {
           field: 'reportable',
         },
       ],
+      initialization: true,
     }
   },
   methods: {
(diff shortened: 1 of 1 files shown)


src/components/ZclAttributeReportingManager.vue
```diff
@@ -221,6 +221,16 @@ export default {
                   .toLowerCase()
                   .includes(this.individualClusterFilterString.toLowerCase())
           })
+
+        resp.forEach((attribute) => {
```
paulr34, line 225: I notice ```attributeData``` is used here ```    <q-table
      class="my-sticky-header-table"
      :rows="attributeData"```

Will this run whenever the user toggles and changes ```this.selectedReporting``` and change it back?

An easy fix, to explain, would be adding a flag like "initialSetup" or something along those lines. After the code here runs, you can have "initialSetup" set to false then add a condition where this code only runs once.

paulr34 (changes requested):

src/components/ZclAttributeReportingManager.vue
YanchengZhou: I did not expect the code to be run whenever ```selectedReporting changes```, and that causes the problem. A flag seems an easy solution here!

YanchengZhou:
I move the part that updates ```this.selectedReporting``` to inside the ```created()``` lifecycle hook. This makes more sense since we only need to call it when the page is initialized (created), thus it will even be more convenient and neat than having a flag. Hope it can pass all CI

paulr34 (approved):

PR #1079, opened 2023-07-20
Fix replace bug of asCleanSymbol in addon-helper

yhoyoon:
The `asCleanSymbol` function in the `addon-helper.js` file has a problem in that only up to 2 spaces can be replaced with underscores.

- `asSnakeCaseLower('Week Day Access Schedules')` -> `'week_day_access _schedules'`

This change modifies the `asCleanSymbol` function to replace all spaces with underscores.

- `asSnakeCaseLower('Week Day Access Schedules')` -> `'week_day_access_schedules'`

### DIFF
--- a/test/gen-template/dotdot/addon-helper.js
+++ b/test/gen-template/dotdot/addon-helper.js
@@ -195,8 +195,7 @@ function isSubstringInString(sub_string, string) {
 
 function asCleanSymbol(label) {
   let l = label.trim()
-  l = l.replace(' ', '_')
-  l = l.replace(' ', '_')
+  l = l.replace(/ /g, '_')
   l = l.replace(/[:/-]/g, '_')
   l = l.replace(/__+/g, '_').toLowerCase()
   return l


brdandu:
@bzbarsky-apple the matter code gen seems to be breaking here. Does not seem related to the change here. Would you know what might be going on here?

yhoyoon:
@brdandu , the matter repo doesn't use this file, so the matter code won't break. However, someone may use that file through zap configuration to use the `asSnakeCaseLower` function.

brdandu:
> @brdandu , the matter repo doesn't use this file, so the matter code won't break. However, anyone who wants to generate custom code may use that file through zap configuration to use the `asSnakeCaseLower` function.
> I chose that example because it represents a potential problem with the function.

@yhoyoon I understand. The only reason I brought that up is because the Matter code regen GitHub job is breaking. It could be just that you need a rebase from Master. Could you rebase your change with the latest on project-chip/zap master?
I think that will fix the failure we are seeing in "Matter code generation ---"

yhoyoon:
@brdandu , thank you! I misunderstood your intentions. I just rebase it with the latest on project-chip/zap master.

brdandu (approved):

yhoyoon:
It looks like a CI failure has occurred that is not related to this PR.
Can I restart CI only with rebase? Is there any other way to solve this issue?

bzbarsky-apple:
@yhoyoon I retriggered the Zigbee job....

bzbarsky-apple:
> @bzbarsky-apple the matter code gen seems to be breaking here.

@brdandu I think this was the temporary "bad merge breaks things on the Matter side" hiccup.

yhoyoon:
@bzbarsky-apple Sorry to bother you, but can you trigger the Zigbee job again?

bzbarsky-apple:
Done.  Though it seems like it's failed a lot.... might not be random.

brdandu:
> Done. Though it seems like it's failed a lot.... might not be random.

The token test failure in zigbee is actually a known issue. There is a .zap file in the tests which was not updated along with the tests when token generation issues were fixed. So it passes and fails in a non deterministic way. However the zap-config diff is an actual issue which was recently resolved with the multiple device type changes so @yhoyoon you may need a rebase. We can merge this after that if it is just the token generation diff.

yhoyoon:
Thanks for confirming the CI issue. I just rebased it.

PR #1092, opened 2023-08-02
Modify Java helper

joonhaengHeo:
When creating a Java jni type, if boxedType is set to false, the type value is not returned. Some scripts have been modified to correct this.


bzbarsky-apple (approved):
OK, though it feels like with some async/await this could be a lot more readable....

yunhanw-google (approved):

PR #1109, opened 2023-08-17
ZCL Upgrade framework

paulr34:


src-shared/db-enum.js
```diff
@@ -135,7 +135,8 @@ exports.sessionOption = {
   manufacturerCodes: 'manufacturerCodes',
   profileCodes: 'profileCodes',
 }
-
+const clusterRevision = '48'
```
tecimovic, line 138: I don't understand this. Why is there a hardcoded number 48 here?

src-electron/db/query-impexp.js
```diff
@@ -678,7 +679,9 @@ WHERE
   if (storagePolicy == dbEnums.storagePolicy.attributeAccessInterface) {
     attribute.storageOption = dbEnums.storageOption.external
   }
-
+  if (attributeId == dbEnums.clusterRevision) {
```
tecimovic, line 682: This literally blows away the value in ZAP file. We can't do this.

tecimovic (changes requested):

src-electron/db/query-impexp.js
tecimovic, line 682: You are also basing the logic on primary key. Nobody is guaranteeing you that the primary key will ever be the same.
You can't base any logic based on primary keys.

PR #1115, opened 2023-08-29
Cleanup exports

paulr34:
Zap is treating database keys like "automatic upgrades" but since the values are not stable this is a mistake.

### DIFF
--- a/src-shared/db-enum.js
+++ b/src-shared/db-enum.js
@@ -138,6 +138,7 @@ exports.sessionOption = {
   clusterSpecification: 'clusterSpecification',
   deviceTypeSpecification: 'deviceTypeSpecification',
 }
+
 const reportingPolicy = {
   mandatory: 'mandatory',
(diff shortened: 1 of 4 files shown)


src-shared/db-enum.js
```diff
@@ -135,7 +135,8 @@ exports.sessionOption = {
   manufacturerCodes: 'manufacturerCodes',
   profileCodes: 'profileCodes',
 }
-
+const clusterRevision = '48'
+exports.clusterRevision = clusterRevision
```
brdandu, line 139: Why do we need this?

brdandu (changes requested):

src-shared/db-enum.js
paulr34: whoops, this was an experiment. deleting, thanks

paulr34:
redesign coming

PR #1117, opened 2023-08-30
Fix non-determinism in the .zap file for endpointType and device type ids

brdandu:
Fixing the non-determinism of endpoint type refs and device type refs in the .zap files by removing it for device type refs and using the index for endpoint type id instead of the primary keys of their corresponding tables since they are non deterministic

### DIFF
--- a/src-electron/db/db-mapping.js
+++ b/src-electron/db/db-mapping.js
@@ -382,6 +382,16 @@ exports.map = {
     }
   },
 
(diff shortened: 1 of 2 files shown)


brdandu:
Note: This will change the endpointType Ids and remove the device type ids the first time you save a .zap file again. However the subsequent files saves will not run into this issue.

bzbarsky-apple (approved):

PR #1121, opened 2023-09-01
Creating a .zap file format with just enabled content

brdandu:
- creating enabled-only-file-format.js for the purpose of saving .zap file with just enabled content
- aLso making sure that the content is easy to read and editable on an editor instead of just the UI
- The file format for saving with enabled only content is 2
- JIRA: ZAPP-1113


brdandu:
Todos: 
- Need to add tests
- Figure out how imp/exp works across different formats

PR #1122, opened 2023-09-01
Fix accessor signatures for string/octstr attributes.

bzbarsky-apple:
See https://github.com/project-chip/connectedhomeip/issues/28357: we are not mutating the actual value the caller will observe.

Note: This is expected to change Matter codegen.  The previous codegen was broken.

### DIFF
--- a/src-electron/generator/matter/app/zap-templates/common/attributes/Accessors.js
+++ b/src-electron/generator/matter/app/zap-templates/common/attributes/Accessors.js
@@ -52,10 +52,13 @@ function canHaveSimpleAccessors(attr) {
 async function accessorGetterType(attr) {
   let type;
   let mayNeedPointer = false;
+  let mayNeedReference = false;
   if (StringHelper.isCharString(attr.type)) {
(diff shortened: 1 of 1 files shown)


src-electron/generator/matter/app/zap-templates/common/attributes/Accessors.js
```diff
@@ -76,6 +79,8 @@ async function accessorGetterType(attr) {
     type = `DataModel::Nullable<${type}> &`;
   } else if (mayNeedPointer) {
     type = `${type} *`;
+  } else if (mayNeedReference) {
```
brdandu, line 82: Would the following give the same result?
else {
type = `${type} &`;
}

bzbarsky-apple: @brdandu For now, yes, unless we add some types that need neither pointer not reference.  I can do that.

brdandu (approved):

Issue #1110, opened 2023-08-18
[Feature request] Add a way to tag default setup values for an attribute


cecille:
Especially in matter, there are a lot of attributes that do not use the ember layer attribute storage, and instead use the override AttributeAccessInterface functionality. These types of attributes should be marked as external so they don't allocate RAM in the storage system despite the fact that they will never hit it.

We have seen many examples come through where attributes were incorrectly tagged as RAM because the ZAP authors were unfamiliar with the underlying implementation. Since we don't expect everyone to read through and understand the code base for every attribute required on their devices, we should provide a way for the SDK authors to tag attributes as external by default.

A similar logic probably applies to some of the other columns, where it would be good to let the SDK authors set up a reasonable set of default starting characteristics since they are most familiar with how the setup should look.

paulr34:
@bzbarsky-apple is this issue a duplicate of https://github.com/project-chip/zap/issues/543?

@cecille my interpretation of this is that there is code which changes the value (as seen in the link) but the global attributes are not properly changed

Or am I misinterpreting the ask?

cecille:
These issues are different as far as I can tell, unless I'm misunderstanding 543. 

The problem here is that people implementing a new configuration for a device won't necessarily know what the implementation of the cluster logic looks like. Hence, they have no idea how to properly set the storage column. It defaults to RAM, which is a safe assumption given no other information, but I'd like a way to supply additional information and adjust the default if the SDK implementation does not use ember storage, so devices aren't wasting RAM space unnecessarily 

I think that possibly applies to other fields as well.

bzbarsky-apple:
Matter already has a way to force attributes to be external, if they are always handled via AttributeAccessInterface,, by adding them to the `attributeAccessInterfaceAttributes` lists in `src/app/zap-templates/zcl/zcl.json` and `src/app/zap-templates/zcl/zcl-with-test-extensions.json`.  What that will do is ignore what the ZAP UI says and just make the attribute external.

Of course it would be better if the ZAP UI reflected reality for such attributes by (1) saying the attribute is external and (2) making the storage field not editable in the UI.

bzbarsky-apple:
@cecille Note that the mechanism I describe is an override, not a hint.  I agree that it might be good to have, for every UI element ZAP presents, a generic ability for an SDK to provide a hint or an override for that element.

Looking at attributes, for Matter, the editable fields seem to be the following:

[...]

PR #1137, opened 2023-09-14
Tagging Attribute Storage Policy properly 

paulr34:
cleaning up storage policy code and fixing issue_1110 in the process https://github.com/project-chip/zap/issues/1110

### DIFF
--- a/src-electron/db/query-config.js
+++ b/src-electron/db/query-config.js
@@ -245,7 +245,7 @@ INTO ENDPOINT_TYPE_ATTRIBUTE (
       cluster.endpointTypeClusterId,
       attributeId,
       staticAttribute.defaultValue ? staticAttribute.defaultValue : '',
-      dbEnum.storageOption.ram,
+      staticAttribute.storagePolicy,
       clusterRef,
       reportMinInterval,
       reportMaxInterval,
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -556,15 +556,15 @@ function prepareCluster(cluster, context, isExtension = false) {
           attribute.$.reportingPolicy
         )
       }
-      let storagePolicy = dbEnum.storagePolicy.any
+      let storagePolicy = dbEnum.storageOption.ram
       if (context.listsUseAttributeAccessInterface && attribute.$.entryType) {
-        storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
+        storagePolicy = dbEnum.storageOption.external
       } else if (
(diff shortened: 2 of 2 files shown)


src-electron/db/query-config.js
```diff
@@ -245,7 +245,7 @@ INTO ENDPOINT_TYPE_ATTRIBUTE (
       cluster.endpointTypeClusterId,
       attributeId,
       staticAttribute.defaultValue ? staticAttribute.defaultValue : '',
-      dbEnum.storageOption.ram,
+      staticAttribute.storagePolicy,
```
brdandu, line 248: What is staticAttribute?

brdandu:
Should this optionality come from the xml for certain attributes?

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -556,15 +556,15 @@ function prepareCluster(cluster, context, isExtension = false) {
           attribute.$.reportingPolicy
         )
       }
-      let storagePolicy = dbEnum.storagePolicy.any
+      let storagePolicy = dbEnum.storageOption.ram
       if (context.listsUseAttributeAccessInterface && attribute.$.entryType) {
-        storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
+        storagePolicy = dbEnum.storageOption.external
```
brdandu, line 561: Is this correct for all use cases @bzbarsky-apple ?

src-electron/db/query-config.js
paulr34: it's an object returned from a query that retrieves attribute data

src-electron/zcl/zcl-loader-silabs.js
paulr34: It is just a bug fix so the database does not have "attributeAccessInterface" as the entry (see photo) but rather "external" which is what t is supposed to be if it is in that list.
Screenshot 2023-09-14 at 10 50 11 AM

paulr34:
> Should this optionality come from the xml for certain attributes?

We have a list in zcl.json called  "attributeAccessInterfaceAttributes" which identifies cluster/attribute pairs which should be deemed "external" for storage policy.

paulr34:
@brdandu I have to fix the conditional so I have to make one more change

src-electron/zcl/zcl-loader-silabs.js
bzbarsky-apple: I'm a little confused.  storagePolicy and storageOption are not the same thing.  Why are we storing storageOption values in storagePolicy @paulr34 ?

Issue #543, opened 2022-06-06
Allow global attributes to have forced storage option


bzbarsky-apple:
We have code in `src-electron/zcl/zcl-loader-silabs.js` that sets storage policy like so:
```
      let storagePolicy = dbEnum.storagePolicy.any
      if (context.listsUseAttributeAccessInterface && attribute.$.entryType) {
        storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
      } else if (
        context.attributeAccessInterfaceAttributes &&
        context.attributeAccessInterfaceAttributes[cluster.name] &&
        context.attributeAccessInterfaceAttributes[cluster.name].includes(name)
      ) {
        storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
      }
```
but that code does not run for global attributes, as far as I can tell.  Then later when we try to use the storage policy in `src-electron/db/query-impexp.js` in `importAttributeForEndpointType` I am not sure _that_ code runs for global attributes either.

So in practice the storage policy bits don't work for global attributes.

tecimovic:
There is currently a mechanism that forced storage option to be external for certain attributes, via this block inside the zcl.json file:
```
  "listsUseAttributeAccessInterface": true,
  "attributeAccessInterfaceAttributes": {
    "Access Control": [
      "ClusterRevision",
      "SubjectsPerAccessControlEntry",
      "TargetsPerAccessControlEntry",
      "AccessControlEntriesPerFabric"
    ],
    "AdministratorCommissioning": [
      "WindowStatus",
      "AdminFabricIndex",
      "AdminVendorId"
    ],
    "Basic": [
      "DataModelRevision",
      "VendorName",
      "VendorID",
      "ProductName",
      "ProductID",
      "Location",
      "HardwareVersion",
    ...
```

If you put a global attribute into this block, however, this doesn't work per cluster. There isn't also any place in the database to put this data per-cluster for a global attribute. We should add this to GLOBAL_ATTRIBUTE_DEFAULT (possibly renaming the table to GLOBAL_ATTRIBUTE_PER_CLUSTER_POLICY).

tecimovic:
So what I plan to do, and @bzbarsky-apple , please review:

1.) Rename GLOBAL_ATTRIBUTE_DEFAULT to GLOBAL_ATTRIBUTE_CLUSTER_POLICY.
2.) Add in there a column STORAGE_POLICY.
[...]

Issue #367, opened 2022-01-11
Should be able to disallow RAM storage for attributes that have to be non-volatile


bzbarsky-apple:
In Matter and Zigbee specs, some attributes are required to be non-volatile.  For non-volatile attributes RAM storage should not be allowed.

For Matter, we'd want this information to come from the XML and default to NVM for such attributes, unless a different storage type is forced via the fix for https://github.com/project-chip/zap/issues/346

brdandu:
From @bzbarsky-apple : Again, this is a thing that makes it too easy to create non-conformant devices, but no one is blocked on this per se.  They can always go and change the storage in the UI...  It's just bad ergonomically.

PR #1140, opened 2023-09-14
properly tagging storage option based on zcl.json

paulr34:


tecimovic (approved):

brdandu (approved):

src-electron/db/query-config.js
```diff
@@ -211,6 +211,14 @@ async function insertOrUpdateAttributeState(
       attributeId,
       clusterRef
     )
+  if (
+    staticAttribute.storagePolicy ==
+    dbEnum.storagePolicy.attributeAccessInterface
+  ) {
+    staticAttribute.storagePolicy = dbEnum.storageOption.external
```
bzbarsky-apple, line 218: This is mixing up _storage policy_ (which can be "any" or "attributeAccessInterface") and _storage option_, which can be "external", "ram", "nvm".  See also https://github.com/project-chip/zap/issues/367

So what I would suggest is that instead of storing "storage option" values in a field that stores storage policy, instead we just use a local variable for the storage option to use for our db insert.

Incidentally, when I was looking over this setup, the `resolve` method on `storagePolicy` in db-enum.js is broken: it has `storagePolicy.attributeAccess` where it should have `storagePolicy.attributeAccessInterface`, no?  Not directly relevant to this PR, but we should fix it here or in a followup.

paulr34: agree, I will make a follow up PR. Thank you.

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -560,6 +560,7 @@ function prepareCluster(cluster, context, isExtension = false) {
       if (context.listsUseAttributeAccessInterface && attribute.$.entryType) {
         storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
       } else if (
+        context.listsUseAttributeAccessInterface &&
```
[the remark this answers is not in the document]
paulr34: I actually thought this was a boolean that indicated whether or not the ```attributeAccessInterface``` was to be referenced or not but now I understand that attribute lists should ALSO be external. Will fix in a follow up. Thank you.

bzbarsky-apple:
Also @paulr34 this does not in fact fix https://github.com/project-chip/zap/issues/543.  What was it trying to fix?

paulr34:
@bzbarsky-apple it fixes this https://github.com/project-chip/zap/issues/1110 but in my opinion it was the first step in the natural progression of fixing all of the attribute storage issues. The front end needs to reflect reality before I can fix anything else. Will work on that one ASAP.

paulr34:
also @bzbarsky-apple this issue https://github.com/project-chip/zap/issues/543 I have this one fixed locally but it requires me to change the XML files. Working on making a couple PR's.

bzbarsky-apple:
> @bzbarsky-apple it fixes this https://github.com/project-chip/zap/issues/1110

Saying that in the description, or at least linking the issue would have been helpful.  No idea how someone is supposed to review a PR if it does not say what it's trying to do.

PR #1153, opened 2023-09-27
Feature/enabled only zap file format2/zapp 1113

brdandu:
Fixing endpoint type command table schema

- incoming and outgoing used to be in the endpoint_type_command table as a single record. However this is an association that is related to the command source and the cluster side. When the cluster side is the same as the command source then it is an outgoing command. If the cluster side and command source are not the same then it is an incoming command. Now we have 2 entries instead of 1 in the endpoint_type_command based on this logic.
- The above allows us to implement the enabled only .zap file correctly or else we lose the incoming commands when the .zap file is saved for the enabled only content since the way the commands were stored before this commit is that the client sourced commands were stored in the client side cluster and the server sourced command is stored in the server cluster which was wrong.
- Updating the endpoint_type_command queries based on the schema changes to the table
- Fixing the import/export for the above schema changes
- Updating the UI code to capture the new schema correctly
- JIRA: ZAPP-1113


src-electron/db/db-mapping.js
```diff
@@ -590,6 +590,8 @@ exports.map = {
       commandRef: x.COMMAND_REF,
       incoming: dbApi.fromDbBool(x.INCOMING),
       outgoing: dbApi.fromDbBool(x.OUTGOING),
+      isIncoming: dbApi.fromDbBool(x.IS_INCOMING),
```
bzbarsky-apple, line 593: What is the difference between IS_INCOMING and INCOMING, and how is an API consumer supposed to know?

brdandu: Let me give an overall picture. I was looking into creating an enabled only .zap file and noticed that when I do that I lost a bunch of commands from the clusters in the .zap file. The reason was because the way commands were stored in the .zap file. The client sourced commands were always stored within the client cluster block and the server sourced command was always stored in the server cluster block. What I had to do is separate those 2 entries into 2 rows i.e. incoming and outgoing had to be 2 separate rows in the endpoint_type_command table. The way they are separated, instead of `incoming` and `outgoing` you now have `isIncoming => 1 indicating incoming and 0 indicating outgoing` and `isEnabled=> 1 for enabled and 0 for disabled` . All the helper queries continue to work the same way by calculating incoming and outgoing using isIncoming and isEnabled to maintain backwards compatibility. Essentially now the endpoint_type_command table records whether the command is incoming or or outgoing and if it is enabled or not.
For future command helpers we can make better use of the isIncoming and isEnabled.

brdandu:
The matter code gen failure seems correct to me by looking at the .zap file as it does not have an incoming but correct me if I am wrong:

`"commands": [
            {
              "name": "RetrieveLogsResponse",
              "code": 1,
              "mfgCode": null,
              "source": "server",
              "incoming": 0,
              "outgoing": 1
            }
          ]`

brdandu:
Corresponding CHIP repo PR: https://github.com/project-chip/connectedhomeip/pull/29509

tecimovic (approved):
It looks ok on the surface. I can't really speak for the depths of the SQL there though. I trust your judgement....

PR #1159, opened 2023-10-03
create mac arch64 package.

jingteng25742:
NOTE: for zap-cli binary, the mac x86 binary is used for the arch64 package
      as a workaround since pkg, the build tool, does not official support for mac-arch64 yet.


tecimovic (approved):
As soon as the CI passes, I'm good with this.

PR #1162, opened 2023-10-04
query attributes with edited global attributes

paulr34:


src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,34 @@ async function getForcedExternalStorage(db, attributeId) {
...
+async function computeStorageTemplate(db, clusterRef, attributes) {
+  let clusterName
+  let forcedExternal
+  clusterName = await queryCluster.selectClusterName(db, clusterRef)
+  attributes.forEach(async (attribute) => {
+    forcedExternal = await getForcedExternalStorage(db, attribute.id)
+    if (
+      forcedExternal.byName &&
+      forcedExternal.byName[clusterName] &&
+      forcedExternal.byName[clusterName].includes(attribute.name)
```
bzbarsky-apple, line 71: Fwiw, you could do: `forcedExternal.byName?.[clusterName]?.includes(attribute.name)`, I expect.

bzbarsky-apple (commented):
This does not seem to fix the output of zcl_attributes....

paulr34:
@tecimovic @brdandu could you guys please take a look at the Zigbee region failures when you have a moment? Thanks

src-electron/generator/helper-zcl.js
```diff
@@ -730,24 +731,31 @@ function zcl_global_commands(options) {
...
+    )
+    attributes = await upgrade.computeStorageTemplate(
+      this.global.db,
+      this.id,
+      attributes
+    )
+  } else {
+    attributes = await queryZcl.selectAllAttributes(this.global.db, packageIds)
+  }
+  if ('removeKeys' in options.hash) {
```
bzbarsky-apple, line 754: The removeKeys bit was not there before, right?  Why is it needed now?

src-electron/generator/helper-zcl.js
```diff
@@ -759,28 +767,37 @@ function zcl_attributes(options) {
...
+      clientAttributes
+    )
+  } else {
+    clientAttributes = await queryZcl.selectAllAttributesBySide(
+      this.global.db,
+      dbEnum.side.client,
+      packageIds
+    )
+  }
+  if ('removeKeys' in options.hash) {
```
bzbarsky-apple, line 796: Again, this is adding removeKeys to something that didn't use to have it, right?

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,48 @@ async function getForcedExternalStorage(db, attributeId) {
   let pkgs = await queryPackage.getPackageRefByAttributeId(db, attributeId)
   let zcl = await queryPackage.getPackageByPackageId(db, pkgs)
   zcl = zcl.path
-  let obj = await fsp.readFile(zcl)
-  let data = JSON.parse(obj)
+  let obj = await fsp.readFile(zcl, 'utf-8')
+  let data
+  if (isJsonString(obj)) {
+    data = JSON.parse(obj)
```
bzbarsky-apple, line 54: So we are parsing as JSON twice?  Why?  Why not just have `maybeParseJsonString` or something that returns the parsed object or `undefined` if parsing throws?

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,48 @@ async function getForcedExternalStorage(db, attributeId) {
...
+  }
   let byName = data?.attributeAccessInterfaceAttributes
   let lists = data?.listsUseAttributeAccessInterface
   let forcedExternal = { byName, lists }
   return forcedExternal
 }
 
+/**
+ * Returns a list of attributes with proper global attributes
```
bzbarsky-apple, line 63: This is ... not very helpful.  Please document what this function _actually_ does.

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,48 @@ async function getForcedExternalStorage(db, attributeId) {
...
+ *
+ * @export
+ * @param {*} db
+ * @param {*} clusterRef
+ * @param {*} attributes
+ * @returns attributes
+ */
+
+async function computeStorageTemplate(db, clusterRef, attributes) {
```
bzbarsky-apple, line 72: This is really a computeStoragePolicy or so, no?

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,48 @@ async function getForcedExternalStorage(db, attributeId) {
...
   return forcedExternal
 }
 
+/**
+ * Returns a list of attributes with proper global attributes
+ *
+ * @export
+ * @param {*} db
+ * @param {*} clusterRef
+ * @param {*} attributes
```
bzbarsky-apple, line 68: This is ... not helpful documentation either.  `attributes` is expected to be an array of objects representing attributes we queried from the db, right?  clusterRef should be the database id (not spec id) of the cluster, I assume?

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,48 @@ async function getForcedExternalStorage(db, attributeId) {
...
+ * @returns attributes
+ */
+
+async function computeStorageTemplate(db, clusterRef, attributes) {
+  let clusterName
+  let forcedExternal
+  clusterName = await queryCluster.selectClusterName(db, clusterRef)
+  return await Promise.all(
+    await attributes.map(async (attribute) => {
```
bzbarsky-apple, line 77: Why do you need the "await attributes" bit?  `attributes.map(async ... )` should give you an array of promises, which is what you want to pass to Promise.all.  Here you're taking the array, creating a promise from it, then awaiting the promise, which adds one extra trip through the event loop, then getting the resulting array and passing it to `all()`... why the extra promise in there?

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,48 @@ async function getForcedExternalStorage(db, attributeId) {
...
+      forcedExternal = await getForcedExternalStorage(db, attribute.id)
+      if (
+        forcedExternal.byName &&
+        forcedExternal.byName[clusterName] &&
+        forcedExternal.byName[clusterName].includes(attribute.name)
```
bzbarsky-apple, line 82: Was there a reason not to do `if (forcedExternal.byName?.[clusterName]?.includes(attribute.name))`?  It's a bit less repetitive....

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +48,48 @@ async function getForcedExternalStorage(db, attributeId) {
...
+      ) {
+        attribute.storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
+        return attribute
+      } else {
+        return attribute
```
bzbarsky-apple, line 87: Why not just `return attribute` outside the `if`?

src-electron/upgrade/upgrade.js
paulr34: At one point this saved me from Buffer errors but those were solved by resolving promises properly. Will remove, thanks.

src-electron/generator/helper-zcl.js
paulr34: yeah, I just figured since it was being used for zcl_attributes_server that I would include it and let someone tell me it was not needed. Thank you, will remove.

src-electron/generator/helper-zcl.js
paulr34: was not sure. will remove, thanks

src-electron/upgrade/upgrade.js
paulr34: will do, thanks

src-electron/upgrade/upgrade.js
paulr34: This is not needed. Thank you. Will remove.

src-electron/upgrade/upgrade.js
paulr34: agree, will fix, thanks

src-electron/upgrade/upgrade.js
paulr34: agree

src-electron/upgrade/upgrade.js
paulr34: @bzbarsky-apple this is the only change I did not push because taking out "await" causes Zigbee generation to fail. I *think* I know why but will do more debug tomorrow and get back to you with a more precise answer. Thank you.

src-electron/upgrade/upgrade.js
paulr34: added better comments and will keep working on them

src-electron/upgrade/upgrade.js
paulr34: This change also causes Zigbee gen to fail.  Will debug more and get back with a clear answer.

src-electron/upgrade/upgrade.js
paulr34: added in separate commit to track CI https://github.com/project-chip/zap/actions/runs/6490297672/job/17625855543

paulr34: @bzbarsky-apple my understanding is that this "await" works as a "try/catch" equivalent. The code fails CI without it. That said, I will debug this more tomorrow and get back to you with more detail.

paulr34: removed

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +40,42 @@ async function getForcedExternalStorage(db, attributeId) {
   let pkgs = await queryPackage.getPackageRefByAttributeId(db, attributeId)
   let zcl = await queryPackage.getPackageByPackageId(db, pkgs)
   zcl = zcl.path
-  let obj = await fsp.readFile(zcl)
-  let data = JSON.parse(obj)
+  let obj = await fsp.readFile(zcl, 'utf-8')
+  let data
+  data = JSON.parse(obj)
```
bzbarsky-apple, line 45: I don't understand this change now....  Why are we declaring and assigning on different lines?

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +40,42 @@ async function getForcedExternalStorage(db, attributeId) {
...
   let byName = data?.attributeAccessInterfaceAttributes
   let lists = data?.listsUseAttributeAccessInterface
   let forcedExternal = { byName, lists }
   return forcedExternal
 }
 
+/**
+ * This function takes a clusterId and an array of attributes associated with that clusterID from the database
```
bzbarsky-apple, line 53:
```suggestion
 * This function takes a clusterId (the database ID, not the specification-defined ID)
```

src-electron/upgrade/upgrade.js
```diff
@@ -40,14 +40,42 @@ async function getForcedExternalStorage(db, attributeId) {
...
+ * and changes the global attributes (attributes with clusterId = null) to represent storage policy
```
bzbarsky-apple, line 54: Nothing below is restricted to global attributes...

paulr34: Thanks, I just added a check "if(attribute.clusterId == null)". The "non global attributes" should already be computed from the loading process. 

This function should only change global attributes.

src-electron/upgrade/upgrade.js
```diff
@@ -22,6 +22,13 @@ const dbEnum = require('../../src-shared/db-enum.js')
 const fs = require('fs')
 const fsp = fs.promises
 
+async function parseJson(json) {
+  try {
+    return JSON.parse(json)
+  } catch (err) {
+    return 'undefined'
```
bzbarsky-apple, line 29:
```suggestion
    return undefined
```

src-electron/upgrade/upgrade.js
```diff
@@ -39,15 +46,49 @@ const fsp = fs.promises
...
+ * That database defined clusterID is used to query the name of the cluster which is in turn used to compute the storage policy for that cluster/attribute pair.
+ *
+ * @export
+ * @param {*} db
+ * @param {*} clusterId (database defined) the clusterId representing a cluster from the database being used in the application
+ * @param {*} attributes an array of objects representing the attributes associated with the cluster
+ * @returns an array of objects representing attributes in the database
+ */
+
+async function computeStoragePolicy(db, clusterId, attributes) {
```
bzbarsky-apple, line 74:
```suggestion
async function computeStoragePolicyForGlobalAttributes(db, clusterId, attributes) {
```
if that's what it does.

bzbarsky-apple (approved):

src-electron/generator/helper-zcl.js
```diff
@@ -730,24 +731,27 @@ function zcl_global_commands(options) {
...
+  let packageIds = await templateUtil.ensureZclPackageIds(this)
+  let attributes = ''
+  if ('id' in this) {
+    // We're functioning inside a nested context with an id, so we will only query for this cluster.
+    attributes = await queryZcl.selectAttributesByClusterIdIncludingGlobal(
+      this.global.db,
+      this.id,
+      packageIds
+    )
+    attributes = await upgrade.computeStoragePolicyForGlobalAttributes(
```
tecimovic, line 746: This doesn't feel right, layer-wise. The point of `upgrade.js` is to perform upgrades when you READ IN the .zap files.
You are here using `upgrade.js` when you're generating stuff.

Not saying that it's the wrong thing, functionally, but I'd then go and move the logic out of `upgrade.js`, put it somewhere else, and maybe have both places call the common place.

src-electron/upgrade/upgrade.js
```diff
@@ -22,6 +22,13 @@ const dbEnum = require('../../src-shared/db-enum.js')
 const fs = require('fs')
 const fsp = fs.promises
 
+async function parseJson(json) {
+  try {
+    return JSON.parse(json)
+  } catch (err) {
+    return undefined
```
tecimovic, line 29: ? You are hiding exception here. Why do we need this? This feels wrong. I don't want for people to blindly start using this function, and then be confused why they get "error: undefined", instead of "error: JSON parsing error".

So this function makes no sense to me.

src-electron/upgrade/upgrade.js
```diff
@@ -39,15 +46,53 @@ const fsp = fs.promises
 async function getForcedExternalStorage(db, attributeId) {
   let pkgs = await queryPackage.getPackageRefByAttributeId(db, attributeId)
   let zcl = await queryPackage.getPackageByPackageId(db, pkgs)
-  zcl = zcl.path
-  let obj = await fsp.readFile(zcl)
-  let data = JSON.parse(obj)
+  zcl = zcl?.path
+  let obj = await fsp.readFile(zcl, 'utf-8')
+  let data = await parseJson(obj)
```
tecimovic, line 51: I don't see why you wrap JSON parsin into a "fake asynchronous call"...  (See my comment above).

src-electron/upgrade/upgrade.js
```diff
@@ -39,15 +46,53 @@ const fsp = fs.promises
...
+
+async function computeStoragePolicyForGlobalAttributes(
+  db,
+  clusterId,
+  attributes
+) {
+  let clusterName
+  let forcedExternal
+  clusterName = await queryCluster.selectClusterName(db, clusterId)
+  return await Promise.all(
```
tecimovic, line 82: If you have an async function, you just want to `return Promise.all(...)`, you don't want to do `return await Promise.all(....)`

By doing it the way you did it, you introduce a fake "synchronize" point, and add another "promise wrapping" on top of it, so it's only killing performance for no benefit.

tecimovic (changes requested):

src-electron/upgrade/upgrade.js
```diff
@@ -39,15 +46,53 @@ const fsp = fs.promises
 async function getForcedExternalStorage(db, attributeId) {
   let pkgs = await queryPackage.getPackageRefByAttributeId(db, attributeId)
   let zcl = await queryPackage.getPackageByPackageId(db, pkgs)
-  zcl = zcl.path
-  let obj = await fsp.readFile(zcl)
-  let data = JSON.parse(obj)
+  zcl = zcl?.path
```
tecimovic, line 49: Somewhere along the way this part is also eating up an "exception condition" (zcl being null or undefined).

src-electron/upgrade/upgrade.js
```diff
@@ -39,15 +46,53 @@ const fsp = fs.promises
 async function getForcedExternalStorage(db, attributeId) {
   let pkgs = await queryPackage.getPackageRefByAttributeId(db, attributeId)
   let zcl = await queryPackage.getPackageByPackageId(db, pkgs)
-  zcl = zcl.path
```
tecimovic, line 42: This feels completely wrong. You are reading a JSON file that is part of the package. All those files should be loaded in during loading time, not here.... But I'd have to see what this does.

paulr34: as discussed, we will fix in follow up PR

src-electron/upgrade/upgrade.js
paulr34: fixed, thanks

src-electron/generator/helper-zcl.js
paulr34: will make follow up PR, thanks

PR #1180, opened 2023-10-23
Expose device class for device types.

bzbarsky-apple:


paulr34 (approved):

src-electron/generator/helper-endpointconfig.js
```diff
@@ -361,7 +361,7 @@ async function endpoint_fixed_device_type_array(options) {
       ret +=
         '{' +
         '0x' +
-        bin.int16ToHex(deviceList[i].deviceId) +
```
brdandu, line 364: Should this change be stacked version dependent?

bzbarsky-apple: @brdandu I'm not sure what you're asking.  Is this a Zigbee vs Matter question?  Or something else?

brdandu: Since this change is restricted to helper-endpointconfig.js which zigbee does not use it is fine for the matter vs zigbee use case. However, I am just thinking out loud.. is there static code in the Matter v1.0 sdk which is looking for the 2 byte device id? In this case, is there a possibility that user can run into issues with this ZAP change and the matter v1.0 application generation and compilation?

bzbarsky-apple: > is there static code in the Matter v1.0 sdk which is looking for the 2 byte device id?

Yes.  But that code is already broken, because 4-byte values get assigned to it other places.

We caught that when we enabled `-Wconversion`, at which point we updated the value to 4-byte.

For those pre-4-byte versions, since they don't use `-Wconversion`, the high two bytes of the 4-byte value here will get silently dropped, which is what was happening before anyway.

brdandu (approved):

src-electron/generator/helper-endpointconfig.js
brdandu: Remaining conversation in https://csamembers.slack.com/archives/C01B01GH64E/p1698155800535739

Issue #1145, opened 2023-09-19
External attributes should not have an editable (or any, really) Default field in the UI


bzbarsky-apple:
For an external attribute the Default field does not make sense.  The value in that field is not used for anything.

Right now, the field exists, and the user can edit it, but then the value is ignored.  This is very misleading and leads to people wasting time setting these values that then get ignored.

I think the UI that would make the most sense is to completely hide the Default field if storage is set to External.  That would avoid showing any currently-set value in it (because that value means nothing), and would avoid users editing the field at all.

Second-best might be just making the field un-editable.

There might be better UI ideas here too, but the important thing is that it should be clear to users that there is no concept of "default value" for external attributes.

paulr34:
https://github.com/project-chip/zap/pull/1181

Issue #1086, opened 2023-07-26
ZAP UI should not allow editing storage setting for things that are forced to be external


bzbarsky-apple:
ZAP has a way to force attributes to be external via SDK settings, but the UI will still allow changing the storage value... which will then be ignored.

The UI should not allow setting things that don't match reality.

paulr34:
@YanchengZhou 

if the SDK forces "storageOption" to be external then the toggle should be disabled. Let's meet and discuss if you can take this on.

paulr34:
https://github.com/project-chip/zap/pull/1181

PR #1181, opened 2023-10-25
disabling GUI buttons of forced External Storage

paulr34:
This PR addresses both issues:
https://github.com/project-chip/zap/issues/1086, 
https://github.com/project-chip/zap/issues/1145

If attribute storage policy is attributeAccessInterface then the storage and the default field are not editable. It also sets default value to null.

The default field is still there but it is disabled.


brdandu:
Seeing a lot of changes to the matter code gen. Is this expected? See all .matter files with diffs.

brdandu (changes requested):
Add tests to show that when external storage option is used then only Matter is affected and not Zigbee

paulr34:
> Seeing a lot of changes to the matter code gen. Is this expected? See all .matter files with diffs.

Yes, they look right. For example, in Access Control we see 
-    callback attribute clusterRevision default = 1;
+    callback attribute clusterRevision;

the default for this external attribute is set to null

paulr34:
> Add tests to show that when external storage option is used then only Matter is affected and not Zigbee

Zigbee doesn't have attributeAccessInterface 

Good catch, I made this commit to limit it to Matter data https://github.com/project-chip/zap/pull/1181/commits/3bc7a3a6116922852290de2843ac02aba660fbe0

As long as Zigbee does not have attributeAccessInterface in their zcl.json then it won't effect it but it would definitely be a good idea to add a cypress test

src-electron/db/query-config.js
```diff
@@ -25,7 +25,7 @@ const dbMapping = require('./db-mapping.js')
 const queryPackage = require('./query-package.js')
 const dbEnum = require('../../src-shared/db-enum.js')
 const queryZcl = require('./query-zcl.js')
-const queryUpgrade = require('../upgrade/upgrade.js')
+const queryUpgrade = require('../matter/matter.js')
```
tecimovic, line 28: As a matter of general rules: we should NOT use the `.js` suffix in require calls.
That allows you to migrate the specific file to a different language (like TS) without having to modify these requires.

src-electron/rest/user-data.js
```diff
@@ -108,7 +109,8 @@ function httpGetPackageNotifications(db) {
 function httpGetPackageNotificationsByPackageId(db) {
   return async (request, response) => {
     let packageId = request.params.packageId
-    let notifications = await queryPackageNotification.getNotificationByPackageId(db, packageId)
+    let notifications =
+      await queryPackageNotification.getNotificationByPackageId(db, packageId)
```
tecimovic, line 113: Why do we have these formatting diffs? Do we not all use the same formatter?

tecimovic (approved):

src-electron/rest/user-data.js
paulr34: not sure

PR #1197, opened 2023-11-01
add mrpActiveThreshold value and ICDOperatesAsLit to the simulated Di…

jmartinez-silabs:
…scoveryCommandResponse

Tested locally with `$ZAP_DEVELOPMENT_PATh,` regen matter and test app (build with `./scripts/examples/gn_build_test_example.sh app1`).

The new variables are generated and usable in the simulated test app.


tecimovic (approved):
We should merge this, but I have a nagging sense that this data should be contributed from SDK, and not hardcoded into the JS file.

paulr34 (approved):

jmartinez-silabs:
@tecimovic Yes I am not sure how it came to be this way. I didn't know the existence of this generation path a few days ago

PR #1199, opened 2023-11-02
update cypress so ZAP will run on Ubuntu 22

wehale:
I had a difficult time getting ZAP to run on my new Ubuntu dev machine, finally found that the versions of Cypress needed to be updated, after making this small change it now runs on Ubuntu 22


tecimovic:
Upgrading Cypress is an elaborate process. I'm closing this PR. 
@wehale : feel free to reopen it once you get to a better state.
`npm run test` is your friend.

PR #1203, opened 2023-11-07
Ci dependencies

paulr34:
upgrading electron, 
deleting extra canvas call that isn't used and breaks everything,
upgrading quasar,
adding python tools to the release
and upgrading cypress


cypress/e2e/attributes/check-cluster-attributes-singleton-in-zigbee-mode.cy.js.skip
```diff
@@ -1,6 +1,6 @@
 /// <reference types="cypress" />
 
-Cypress.on('uncaught:exception', (err, runnable) => {
+Cypress.off('uncaught:exception', (err, runnable) => {
```
brdandu, line 3: Why is this switched off?

cypress/e2e/clusters/cluster-multiple-search.cy.js
```diff
@@ -49,10 +49,8 @@ describe('Add multiple clusters and search', () => {
   })
   it('Search action', () => {
     cy.fixture('data').then((data) => {
-      cy.get(
```
brdandu, line 52: Can you get rid of this diff?

cypress/e2e/custom_xml/custom_xml.cy.js.skip
```diff
@@ -17,7 +17,7 @@ describe('Check ZCL Extension functionalities', () => {
     cy.window()
       .then(function (window) {
         window[rendApi.GLOBAL_SYMBOL_EXECUTE](
-          rendApi.id.open,
+          restApi.ide.open,
```
brdandu, line 20: what is the difference between id and ide?

.github/workflows/release.yml
```diff
@@ -160,6 +160,8 @@ jobs:
         with:
           node-version: 18.x
           cache: 'npm'
+      - name: pip
+        run: pip install setuptools
```
brdandu, line 164: What does setup tools do?

cypress.config.ts
brdandu, line 1: Why is this a typescript file instead of a js file?

cypress/e2e/attribute_validations/bitmap.cy.js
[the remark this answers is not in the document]
paulr34: Cypress update which had to be done. Basically, upgrading something like Electron makes a new dependency tree and the old dependencies are not valid anymore.

The Cypress upgrade required a program to run and change the file extensions and some of the content.

More details here: https://docs.cypress.io/guides/references/configuration

cypress/e2e/attributes/check-cluster-attributes-singleton-in-zigbee-mode.cy.js.skip
paulr34: that is a typo, thank you

cypress/e2e/clusters/cluster-multiple-search.cy.js
paulr34: yes, no problem , thank you

.github/workflows/release.yml
paulr34: So there were a handful of problems that this PR was solving. One of them being that we couldn't build a Windows or a Mac binary anymore. The Mac problem was due to dependency issues. Basically npm install would fail on a Mac before it even got to building a binary.

After the dependency issues were resolved, Mac binary built fine but Windows still had failures. One of the errors was that the distutils (dist folder) python modules had been "end of life". Basically, an alternative to using distutils was to install setuptools. After installing setuptools the Windows binary built.

https://www.geeksforgeeks.org/differences-between-distribute-distutils-setuptools-in-python/

cypress.config.ts
paulr34: This file was auto generated by the Cypress upgrade.

cypress/e2e/attribute_validations/bitmap.cy.js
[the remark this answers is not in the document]
brdandu: Were all these files renamed automatically through some script or you had to figure it out file by file?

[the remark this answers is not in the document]
paulr34: script: https://docs.cypress.io/guides/references/configuration

cypress/e2e/custom_xml/custom_xml.cy.js.skip
brdandu: ok

cypress.config.ts
brdandu: Is this essentially the same content as what was in `cypress.json`. Could this be put in a js file instead and it automatically works?

paulr34: where? Im looking in VScode and don't see it. I didn't manually add that file.

.github/workflows/release.yml
brdandu: Does this information need to be updated to any of our documentation for ZAP's source installation ?

paulr34: yes, probably

cypress.config.ts
brdandu: This one replaced https://github.com/project-chip/zap/blob/ca7937a6aebf3783fee548390fe35e30dc45be55/cypress.json

brdandu (approved):
Approving given the CI passes like it should

tecimovic:
Doing an update is a good thing.

2 comments:
   1.) Run `npm audit`. See a lot of packages that are in various states of "security problem". Can we update many of those?
   2.) Run `npm outdated`. Long list. Can we updated many of those?
   3.) I see electron-builder is outdated. If we're upgrading electron, we should probably also update electron-builder. They go hand in hand.

paulr34:
> Doing an update is a good thing.
> 
> 2 comments: 1.) Run `npm audit`. See a lot of packages that are in various states of "security problem". Can we update many of those? 2.) Run `npm outdated`. Long list. Can we updated many of those? 3.) I see electron-builder is outdated. If we're upgrading electron, we should probably also update electron-builder. They go hand in hand.

Thanks @tecimovic I updated electron-builder. 

1. audit causes failures so I will have to go through them 1 by one. This should be done in a follow up PR.
2.) npm outdated provided some good information and it should be done but in a "phase 2" second PR

agree?

PR #1207, opened 2023-11-09
fix about page bug

paulr34:


tecimovic (approved):

tecimovic:
There are CI failures, otherwise I'm ok with the fix itself.

PR #1211, opened 2023-11-14
Start consuming matter xml.

tecimovic:


paulr34 (approved):

src-electron/zcl/zcl-loader-dotdot.js
```diff
@@ -1255,7 +1260,7 @@ async function loadIndividualDotDotFile(db, filePath) {
  * @param {*} ctx Context of loading.
  * @returns a Promise that resolves with the db.
  */
-async function loadDotdotZcl(db, metafile) {
+async function loadToplevelXmlFile(db, metafile) {
```
brdandu, line 1263: Why are we using the dot dot xml loader?

brdandu: I guess this is just a name change.

Issue #1183, opened 2023-10-26
Monitor and report on validity of configuration


tecimovic:
Problem:
   users are not properly informed if their configuration is "according to spec" and will "pass certification".

Background:
   based on experience with previous "spec and certification" systems, zap is developed with the following UI principle in mind:
     - allow users to modify configuration *however they want* and *whenever they want*.
     - but when data becomes "outside of spec" and "not certificable", then show them warnings/messages/errors/something.
     
So we need mechanisms that covers following use cases:
  1.) You load in a .zap file. .zap file is "not according to spec". As soon as the file loads, the message shows describing this, and the elements that are causing the problem (clusters that should be there, missing clusters that are not there, etc, etc) are highlighted in UI.
   2.) If user deselects a cluster/attribute/command/event that is considered MANDATORY in order to pass certification, the deselection SHOULD SUCCEED, however zap should immediatelly highlight this row as a "warning" and show some message describing what just happend. Something like: "This attribute is considered mandatory for device type DoorLock".
   3.) If user loads in custom XML, which imposes additional "spec" constraints, zap should immediatelly after the load _recalculate_ the state of "up to spec-ness", and pop up appropriate warnings/errors/message if they are required.
   4.) User should have a place in UI where they can view ALL the messages related to "not being up to spec", and can go over them one by one and fix them.
   5.) When generating from command line, there should be a clear warning shown if something is not up to spec.
   6.) We should have a CLI switch "--allowSpecIssues". If this is present, then generation will happen, even though things are not up to spec. If it's not, then zap will simply blow up with a clear error and refuse to generate.
   
And, very obviously:
   this functionality is "spec specific", so thoughts should be given to how this should work on Matter and how it should work on Zigbee. If we need to modify schema to add more data, we should. And things that SHOULD BE driven by the SDK, need to come from SDK. We don't want the code that says: "If matter then X else Y". All that should be driven from the `zcl.json` and it's child files.

jepenven-silabs:
Great issue, some work was already done in the https://github.com/project-chip/connectedhomeip/blob/master/scripts/rules.matterlint but if zap could validate everything that would be great.

paulr34:
@tecimovic so if the user changes anything that came from the XML we should pop a notification as suggested?

tecimovic:
[...]

PR #1212, opened 2023-11-15
cleanup Matter SDK code

paulr34:


brdandu:
What is this PR cleaning up?

paulr34:
> What is this PR cleaning up?

The Matter SDK code. Meaning, the code which is not the Matter spec but SDK specific.

paulr34:
> > What is this PR cleaning up?
> 
> The Matter SDK code. Meaning, the code which is not the Matter spec but SDK specific.

For example, there is a list called "attributeAccessInterfaceAttributes" in zcl.json that has cluster/attribute pairs that must be forced external storage. My understanding is this is not spec code compliant rules but rather SDK code.

Please see https://github.com/project-chip/zap/issues/1183

"But then there are also SDK specific rules and implementation rules, such as some of these "storage" issues and such. These might result in devices not getting certified for implementation issues."

paulr34:
deprioritized

src-electron/db/query-package.js
```diff
@@ -595,6 +595,23 @@ async function getAllPackages(db) {
...
+      db,
+      `SELECT 
+        PACKAGE_REF, 
+        OPTION_CATEGORY, 
+        OPTION_CODE, 
+        OPTION_LABEL
+       FROM 
+        PACKAGE_OPTION
+       WHERE
+        OPTION_CODE = 999999`
```
brdandu, line 652: What is 999999?

src-electron/sdk/matter.js
```diff
@@ -52,15 +37,11 @@ async function getForcedExternalStorageList(db, zcl) {
...
-  zcl = zcl?.path
-  let obj = await fsp.readFile(zcl, 'utf-8')
-  let data = parseJson(obj)
-  let byName = data?.attributeAccessInterfaceAttributes
-  let lists = data?.listsUseAttributeAccessInterface
-  let forcedExternal = { byName, lists }
+async function getForcedExternalStorage(db) {
+  let forcedExternal = await queryPackage.getAttributeAccessInterface(
+    db,
+    '999999'
```
brdandu, line 43: Where does 999999 come from? same as above

src-electron/sdk/matter.js
```diff
@@ -92,17 +73,23 @@ async function computeStoragePolicyForGlobalAttributes(
...
       }
       return attribute
     })
   )
 }
-async function getDisabledStorage(db, zcl) {
-  return await getForcedExternalStorageList(db, zcl)
+
+async function getDisabledStorage(db) {
+  return getForcedExternalStorage(db)
```
brdandu, line 92: Whatever is calling getDisabledStorage can it just call  getForcedExternalStorage ?

src-electron/sdk/matter.js
```diff
@@ -163,19 +155,21 @@ async function computeStorageImport(
...
   attributeName
 ) {
-  if (
-    forcedExternal.byName &&
-    forcedExternal.byName[clusterName] &&
-    forcedExternal.byName[clusterName].includes(attributeName)
-  ) {
-    storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
-  }
+  forcedExternal.map((option) => {
```
brdandu, line 158: Is .map the right function to be used here? wondering if you just need forEach here since you just need to set storage policy

src/components/ZclAttributeManager.vue
```diff
@@ -267,20 +267,23 @@ export default {
...
-      if (
-        this.forcedExternal.byName?.[this.selectedCluster.label]?.includes(name)
-      ) {
-        return true
-      } else {
-        return false
-      }
+    checkForcedExternal(name, selectedCluster) {
+      let forcedExternal
+      this.forcedExternal.map((option) => {
```
brdandu, line 259: Same as above. Did you mean .map here or typo?

brdandu (changes requested):

src-electron/sdk/matter.js
paulr34: test code, needs to be cleaned

src-electron/db/query-package.js
brdandu: AttributeAccessInterface

src-electron/sdk/matter.js
```diff
@@ -23,26 +23,11 @@ const dbEnum = require('../../src-shared/db-enum.js')
...
-  } catch (err) {
-    return undefined
-  }
-}
-
 /**
- * This file implements upgrade rules which are used to upgrade .zap files and xml files
- * to be in sync with the spec
+ * This file implements SDK rules which are used to upgrade .zap files and xml files
+ * to be in sync with the Matter SDK and spec
```
brdandu, line 28: Does this comment need to be removed along with the function definition?

src-electron/sdk/matter.js
```diff
@@ -92,19 +73,20 @@ async function computeStoragePolicyForGlobalAttributes(
     attributes.map(async (attribute) => {
       if (attribute.clusterId == null) {
         forcedExternal = await getForcedExternalStorage(db, attribute.id)
-        if (forcedExternal.byName?.[clusterName]?.includes(attribute.name)) {
-          attribute.storagePolicy =
-            dbEnum.storagePolicy.attributeAccessInterface
-        }
+        forcedExternal.map((option) => {
```
brdandu, line 76: .map may be a typo here? Convert to for loop and break out of it once the if condition is satisfied?

src-electron/sdk/matter.js
```diff
@@ -118,33 +100,38 @@ async function getDisabledStorage(db, zcl) {
...
+async function computeStoragePolicyNewConfig(
+  db,
+  clusterRef,
+  storagePolicy,
+  forcedExternal,
+  attributeName
+) {
+  let clusterName
+  clusterName = await queryCluster.selectClusterName(db, clusterRef)
+  forcedExternal.map((option) => {
```
brdandu, line 124: .map may be a typo here? Convert to for loop and break out of it once the if condition is satisfied?

src-electron/sdk/matter.js
```diff
@@ -118,33 +100,38 @@ async function getDisabledStorage(db, zcl) {
...
 
+async function computeStoragePolicyNewConfig(
+  db,
+  clusterRef,
+  storagePolicy,
+  forcedExternal,
+  attributeName
+) {
+  let clusterName
+  clusterName = await queryCluster.selectClusterName(db, clusterRef)
```
brdandu, line 123: minor issue => let clusterName = await queryCluster.selectClusterName(db, clusterRef)

src-electron/sdk/matter.js
```diff
@@ -163,19 +150,20 @@ async function computeStorageImport(
...
-    forcedExternal.byName[clusterName].includes(attributeName)
-  ) {
-    storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
-  }
+  forcedExternal.forEach((option) => {
+    if (
+      option.optionCategory == clusterName &&
+      option.optionLabel == attributeName
+    ) {
+      storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
```
brdandu, line 158: Need not go through the rest of the elements in the `forcedExternal` once the match has been found. You can break out of the forEach loop using the `return` statement here.
Similar change needs to happen in the above .map function calls above

brdandu (changes requested):

src-electron/db/query-package.js
paulr34: done, thanks

src/components/ZclAttributeManager.vue
paulr34: fixed

src-electron/sdk/matter.js
paulr34: no it does the same thing just in a slightly different way

src-electron/sdk/matter.js
paulr34: oh ya good point

src-electron/sdk/matter.js
paulr34: good point, about to push a fix

src-electron/sdk/matter.js
paulr34: fixed

src-electron/sdk/matter.js
brdandu: Unresolving this since it looks the same. Needs a forEach?

brdandu (changes requested):

brdandu (approved):

src-electron/sdk/matter.js
```diff
@@ -163,19 +149,23 @@ async function computeStorageImport(
...
-  if (
-    forcedExternal.byName &&
-    forcedExternal.byName[clusterName] &&
-    forcedExternal.byName[clusterName].includes(attributeName)
-  ) {
-    storagePolicy = dbEnum.storagePolicy.attributeAccessInterface
-  }
-  return storagePolicy
+  let updatedStoragePolicy = storagePolicy
+  forcedExternal.some((option) => {
```
brdandu, line 153: some is used to check if a condition is met when searching through the array. The purpose of this code seems to be used to update `updatedStoragePolicy` so a foreach and return seems to be the right way to go about this

brdandu (approved):

src-electron/sdk/matter.js
paulr34: The purpose is to change the policy if the condition is met. No need to do any more work after that.

PR #1215, opened 2023-11-21
Change for renderer API integration

pbalint:
This change adds a save button, controllable by a query parameter and glue code so zap is usable by studio v6


src/App.vue
```diff
@@ -34,16 +34,33 @@ limitations under the License.
...
 const storage = require('./util/storage.js')
 
+window.addEventListener('message', (event) => {
+    const eventData = event?.data?.eventData
+    switch (event?.data?.eventId) {
+      case 'theme':
+        window[rendApi.GLOBAL_SYMBOL_EXECUTE](rendApi.id.setDarkTheme, eventData.theme === 'dark')
+        break
+      case 'openFile':
+        window[rendApi.GLOBAL_SYMBOL_EXECUTE](rendApi.id.open, eventData.path)
```
paulr34, line 52: we don't open files with this method anymore

paulr34 (changes requested):

src/App.vue
```diff
@@ -206,6 +223,13 @@ export default defineComponent({
           this.$store.dispatch('zap/updateSelectedUcComponentState', resp)
         }
       )
+
+      this.$onWebSocket(
+        dbEnum.wsCategory.dirtyFlag,
```
paulr34, line 228: The dirty flag is constantly polled from the backend already via web socket

src/components/ZCLToolbar.vue
```diff
@@ -250,6 +264,11 @@ export default {
         )
       }
     },
+    saveChanges() {
```
paulr34, line 267: this is already done in menu.js

src/store/zap/actions.js
```diff
@@ -817,6 +817,10 @@ export function updateSelectedUcComponentState(context, projectInfo) {
   })
 }
 
+export function setDirtyState(context, isDirty) {
```
paulr34, line 820: again, the polling mechanism already handles the dirty state

src/components/ZCLToolbar.vue
pbalint: These changes are done in order to embed ZAP into Studio v6, where there's no electron menu...

src/App.vue
pbalint: Yes, but there's no reflection of that in the UI state, so we can't conditionally show HTML UI elements based on that.
(These changes are done in order to embed ZAP into Studio v6, where there's no electron menu and potentially multiple ZAP UI instances in the same browser instance (sharing the session id)...)

.husky/.gitignore
```diff
@@ -1 +0,0 @@
-_
```
brdandu, line 1: Get rid of this change

.husky/pre-commit
```diff
@@ -1,17 +0,0 @@
...
-npx jsdoc src-electron src-shared -r -d ./generated-html/
-
-echo "🚀   > Eslint..."
-npx eslint --ext .js,.vue src src-electron src-shared src-script test
-
-echo "🚀   > Reset version in package.json to 0.0.0..."
-src-script/zap-update-package-version.js -fake
-
-echo "🚀   > Check versions in package-lock.json..."
-src-script/zap-validate-metafiles.js
```
brdandu, line 17: Why is this removed?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "zap",
-  "version": "0.0.0",
+  "version": "2023.11.21",
```
brdandu, line 3: This should not change

src/boot/ws.js
```diff
@@ -24,6 +24,7 @@ import * as Util from '../util/util.js'
 const tickInterval = 15000 // 15 seconds tick interval for server watchdog.
 
 let eventEmitter = new Events.EventEmitter()
+
```
brdandu, line 27: Get rid of this

.husky/pre-commit
pbalint: Wanted to quickly push draft PR, reverted this once the draft was finalized

paulr34 (approved):

PR #1218, opened 2023-11-27
Fix zapTypeToClusterObjectType to have correct output for acronyms.

bzbarsky-apple:
zapTypeToClusterObjectType needs to be consistent with the actual types we generate for cluster objects, and it was producing different output for a struct (or event) which had an all-caps name.


brdandu:
There is a matter code gen failure

bzbarsky-apple:
@brdandu That codegen failure is because ZAP changed the codegen and no one bothered to update the version in Matter.  It has nothing to do with this PR.

PR #1221, opened 2023-11-29
Feature/spec compliance feature/zap 1194

brdandu:
Summarizing the Matter specification failures in one page and also adding them to the console during file open


test/resource/spec-check-all-clusters-app-matter.zap
paulr34, line 1: why add this?

brdandu: Testing by intentionally removing some device type and cluster requirements so that more use cases can be tested upon file open

PR #1225, opened 2023-12-04
Allow the ZAP UI to load when a custom xml file is not found instead of throwing an error

brdandu:
- The changes to import-json.js allows the ZAP UI to load instead of throwing an error and locking ZAP.
- Now the error message appears in the ZAP notification's pane so that the user can see a detailed error and act upon it
- Minor cleanup
- JIRA: ZAPP-1148


tecimovic (approved):
Looks good to me.

PR #1228, opened 2023-12-05
schema change so new .zap is created

paulr34:


tecimovic:
Technically: We're not deleting the `.zap` directory, we're merely sidelining the existing zap.sqlite and creating a new one.

paulr34:
> Technically: We're not deleting the `.zap` directory, we're merely sidelining the existing zap.sqlite and creating a new one.

okay changed it, thanks

PR #1233, opened 2023-12-08
Adding more logic to pick a better package before picking up the first package in the list of available packages

brdandu:
…

### DIFF
--- a/src-electron/importexport/import-json.js
+++ b/src-electron/importexport/import-json.js
@@ -213,11 +213,30 @@ async function importSinglePackage(db, pkg, zapFilePath, packageMatch) {
       packageType: pkg.type,
     }
   } else {
-    // None exists, so use the first one from 'packages'.
     let p = packages[0]
-    env.logWarning(
-      `None of packages exist, so using first one overall: ${p.id}.`
+    let pkgPaths = packages.map((p) => p.path)
+    //eslint-disable-next-line
+    let filePathToSearch = pkg.path.split(/.*[\/|\\]/).pop()
+    let packageNameMatch = packages.find(
+      (p) => p.path.includes(filePathToSearch) && fs.existsSync(p.path)
     )
+    if (packageMatch) {
+      p = packageNameMatch
+      env.logError(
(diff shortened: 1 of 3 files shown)


tecimovic (approved):
We can do this as a temporary workaround.

But we REALLY have to weed out all the code that EVER "picks the first one", because it doesn't know what to do.

If you can't make an unambiguous decision, you NEED TO defer it back to the user.

src-electron/importexport/import-json.js
```diff
@@ -213,11 +213,30 @@ async function importSinglePackage(db, pkg, zapFilePath, packageMatch) {
...
   } else {
-    // None exists, so use the first one from 'packages'.
     let p = packages[0]
-    env.logWarning(
-      `None of packages exist, so using first one overall: ${p.id}.`
+    let pkgPaths = packages.map((p) => p.path)
+    //eslint-disable-next-line
+    let filePathToSearch = pkg.path.split(/.*[\/|\\]/).pop()
+    let packageNameMatch = packages.find(
+      (p) => p.path.includes(filePathToSearch) && fs.existsSync(p.path)
```
tecimovic, line 221: Generally, we should never use `.existsSync`.

brdandu:
> We can do this as a temporary workaround.
> 
> But we REALLY have to weed out all the code that EVER "picks the first one", because it doesn't know what to do.
> 
> If you can't make an unambiguous decision, you NEED TO defer it back to the user.

Added more logic to this such that the passed zcl and temaplates json files are used when generate is called directly without opening up the UI

paulr34 (commented):
What does this solve?

paulr34:
> We can do this as a temporary workaround.
> 
> 
> 
> But we REALLY have to weed out all the code that EVER "picks the first one", because it doesn't know what to do.
> 
> 
> 
> If you can't make an unambiguous decision, you NEED TO defer it back to the user.

@tecimovic it doesnt "pick the first one". It simply passes in 1 file when the database is broken.

PR #1248, opened 2024-01-18
disabing default if storage is manually changed to external

paulr34:
This PR addresses the issue of attributes that are manually changed to External Storage. 

If an attribute is set to External Storage at any moment then the default value will be uneditable.

Additionally, in this case the value should be set to null to avoid user error.

### DIFF
--- a/src/components/ZclAttributeManager.vue
+++ b/src/components/ZclAttributeManager.vue
@@ -189,7 +189,7 @@ limitations under the License.
                   : ''
               "
               :disable="
-                isDisabled(
+                isDisabledDefault(
                   props.row.id,
                   props.row.label,
                   selectedCluster.id,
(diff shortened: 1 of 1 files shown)


src/components/ZclAttributeManager.vue
```diff
@@ -276,6 +276,17 @@ export default {
         return false
       }
     },
```
brdandu, line 278: When do you use isDisabled v/s isDisabledDefault ?

src/components/ZclAttributeManager.vue
```diff
@@ -276,6 +276,17 @@ export default {
...
+    isDisabledDefault(id, name, selectedClusterId, selectedCluster) {
```
brdandu, line 279: Why do you pass "selectedCluster" as function parameter?

brdandu (changes requested):

src/components/ZclAttributeManager.vue
paulr34: The default value should be disabled whenever the Storage is set to "External".

In a case where the user sets Storage to External but wants to change it back to RAM then it should be editable while the default value is uneditable.

src/components/ZclAttributeManager.vue
paulr34: good catch, will fix, thanks

PR #1249, opened 2024-01-18
Default disable

paulr34:
This PR addresses the issue of attributes that are manually changed to External Storage.

If an attribute is set to External Storage at any moment then the default value will be uneditable.

Additionally, in this case the value should be set to null to avoid user error.

### DIFF
--- a/src/components/ZclAttributeManager.vue
+++ b/src/components/ZclAttributeManager.vue
@@ -178,8 +178,7 @@ limitations under the License.
                 defaultValueCheck(
                   props.row.id,
                   props.row.label,
-                  selectedCluster.id,
-                  selectedCluster.label
+                  selectedCluster.id
                 )
               "
               :error="
--- a/src/components/ZclAttributeManager.vue
+++ b/src/components/ZclAttributeManager.vue
@@ -106,8 +106,7 @@ limitations under the License.
                 isDisabledStorage(
                   props.row.id,
                   props.row.label,
-                  selectedCluster.id,
-                  selectedCluster.label
+                  selectedCluster.id
                 )
               "
               class="col"
@@ -130,14 +129,7 @@ limitations under the License.
               :model-value="selectionSingleton"
               :val="hashAttributeIdClusterId(props.row.id, selectedCluster.id)"
               indeterminate-value="false"
-              :disable="
-                isDisabled(
-                  props.row.id,
-                  props.row.label,
-                  selectedCluster.id,
-                  selectedCluster.label
-                )
-              "
+              :disable="isDisabled(props.row.id, selectedCluster.id)"
               @update:model-value="
                 handleLocalSelection(
                   $event,
@@ -154,14 +146,7 @@ limitations under the License.
               :model-value="selectionBounded"
               :val="hashAttributeIdClusterId(props.row.id, selectedCluster.id)"
               indeterminate-value="false"
-              :disable="
-                isDisabled(
-                  props.row.id,
-                  props.row.label,
-                  selectedCluster.id,
-                  selectedCluster.label
-                )
-              "
+              :disable="isDisabled(props.row.id, selectedCluster.id)"
               @update:model-value="
                 handleLocalSelection(
                   $event,
@@ -188,14 +173,7 @@ limitations under the License.
                   ? 'grey'
                   : ''
               "
-              :disable="
-                isDisabledDefault(
-                  props.row.id,
-                  props.row.label,
-                  selectedCluster.id,
-                  selectedCluster.label
-                )
-              "
+              :disable="isDisabledDefault(props.row.id, selectedCluster.id)"
               :model-value="
                 defaultValueCheck(
                   props.row.id,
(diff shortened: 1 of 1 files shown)


paulr34:
@brdandu I renamed the branch..

Note: I did address your comments in the previous PR

src/components/ZclAttributeManager.vue
```diff
@@ -276,13 +279,33 @@ export default {
         return false
       }
     },
-    isDisabled(id, name, selectedClusterId, selectedCluster) {
```
brdandu, line 279: What about all the places where isDisabled is called from? Don't those need a signature change?

src/components/ZclAttributeManager.vue
```diff
@@ -276,13 +279,33 @@ export default {
...
+        !this.selection.includes(
+          this.hashAttributeIdClusterId(id, selectedClusterId)
+        ) ||
+        this.checkForcedExternal(name) ||
+        this.selectionStorageOption[
+          this.hashAttributeIdClusterId(id, selectedClusterId)
+        ] == 'External'
+      )
+    },
+    isDisabled(id, name, selectedClusterId) {
```
brdandu, line 293: Can these functions have more description as comments? I find it difficult to understand the difference by taking quick look.

src/components/ZclAttributeManager.vue
```diff
@@ -276,13 +279,33 @@ export default {
         return false
       }
     },
-    isDisabled(id, name, selectedClusterId, selectedCluster) {
+    isDisabledDefault(id, name, selectedClusterId) {
+      return (
```
brdandu, line 283: return (isDisabled() ||
         this.selectionStorageOption[
           this.hashAttributeIdClusterId(id, selectedClusterId)
         ] == 'External')
         
         
         avoid the duplicate code

brdandu (changes requested):

src/components/ZclAttributeManager.vue
paulr34: sure, will do

src/components/ZclAttributeManager.vue
paulr34: which part is duplicated? this case requires more conditions..

ill see what I can do to clear it up, thanks

src/components/ZclAttributeManager.vue
paulr34: no the new method isDisabledDefault is a new method for a specific use case

paulr34: basically, rather than changing is disabled by adding if and else statements I just created a new method

brdandu: Are there function calls to isDisabled anywhere else in the code?

paulr34: yes, whenever an attribute is disabled...I added comments

brdandu: That is not what I am asking. Is `isDisabled` function called from anywhere in this file?
You have changed the signature of the function definition from `isDisabled(id, name, selectedClusterId, selectedCluster)` to `isDisabled(id, selectedClusterId)` . 
Now, what about the same change everywhere else in the code where there are function calls to `isDisabled(id, name, selectedClusterId, selectedCluster)`. Don't those need to be changed as well to `isDisabled(id, selectedClusterId)` ?

src/components/ZclAttributeManager.vue
brdandu: Isn't `!this.selection.includes(
         this.hashAttributeIdClusterId(id, selectedClusterId)
       )` part the same in `isDisabled(id, selectedClusterId)` and ` isDisabledDefault(id, name, selectedClusterId)` ?

src/components/ZclAttributeManager.vue
```diff
@@ -276,13 +279,40 @@ export default {
...
+        !this.selection.includes(
+          this.hashAttributeIdClusterId(id, selectedClusterId)
+        ) ||
+        this.selectionStorageOption[
+          this.hashAttributeIdClusterId(id, selectedClusterId)
+        ] == 'External'
+      )
+    },
+    //disabling Storage if forced External AND if attribute is not enabled
```
brdandu, line 293: comment=> Returns true if storage is forced as External and attribute is disabled. ?

src/components/ZclAttributeManager.vue
brdandu, line 1: Add cypress tests to validate the behavior here.

brdandu (changes requested):

tecimovic (approved):

paulr34:
@brdandu thanks for the detailed review. I fixed the method signatures and will work on adding more cypress coverage in a follow up PR. Much appreciated.

src/components/ZclAttributeManager.vue
paulr34: good point, ill add Returns true
I kinda thought it was implied

src/components/ZclAttributeManager.vue
paulr34: yes but this is a different set of conditions. 

do you think I should somehow use isDisabled within isDisabledStorage?

I think that would cause more confusion

paulr34: isDisabled has a condition

isDisabledStorage has that same condition plus another condition

isDisabledDefault has both of the conditions plus another one


I could make isDisabled do all of this work but it would be much harder to read imo

paulr34: Actually, Vue has limitations which would make it even harder to have isDisabled do all of this work. Ill look more into it though...

PR #1251, opened 2024-01-22
fix GUI bug

paulr34:


brdandu:
Looks good to me given the tests pass

PR #1252, opened 2024-01-22
update database and hide default disable

paulr34:
this PR updates the db with default input if the attribute is External or not enabled


paulr34:
needs more work

PR #1257, opened 2024-01-30
Phase 1 - Composite Endpoints

paulr34:
phase 1 - Endpoint Composition


src-electron/db/zap-schema.sql
```diff
@@ -808,6 +808,7 @@ CREATE TABLE IF NOT EXISTS "ENDPOINT" (
   "PROFILE" integer,
   "ENDPOINT_IDENTIFIER" integer,
   "NETWORK_IDENTIFIER" integer,
+  "PARENT" integer,
```
tecimovic: This should be called `PARENT_ENDPOINT_REF`.

src-electron/db/zap-schema.sql
```diff
@@ -808,6 +808,7 @@ CREATE TABLE IF NOT EXISTS "ENDPOINT" (
...
   foreign key (SESSION_REF) references SESSION(SESSION_ID) on delete cascade,
   foreign key (ENDPOINT_TYPE_REF) references ENDPOINT_TYPE(ENDPOINT_TYPE_ID) on delete
   set NULL
```
tecimovic, line 814: There should be a foreign key relationship here, saying that `PARENT_ENDPOINT_REF` needs to point to an existing `ENDPOINT_ID`.

tecimovic (changes requested):
Generally ok in terms of flow, but PLEASE make sure that in the database the PARENT is called PARENT_ENDPOINT_REF and it contains a reference to the primary key of the endpoint that is being referenced to. NOT just the "ENDPOINT_IDENTIFIER", which is just a random number user types in.

We need database integrity, so we need to make references link to primary keys.

src-electron/db/zap-schema.sql
paulr34: yes will do

src-electron/db/zap-schema.sql
```diff
@@ -808,6 +808,7 @@ CREATE TABLE IF NOT EXISTS "ENDPOINT" (
   "PROFILE" integer,
   "ENDPOINT_IDENTIFIER" integer,
   "NETWORK_IDENTIFIER" integer,
+  "PARENT_ENDPOINT_REF" integer NULL,
```
tecimovic, line 811: You need a foreign key constraint here. It needs to be a foreign key into ENDPOINT_ID. 'on delete set nul' feels right.

src-electron/db/query-endpoint.js
```diff
@@ -335,6 +341,32 @@ async function deleteEndpoint(db, id) {
   return dbApi.dbRemove(db, 'DELETE FROM ENDPOINT WHERE ENDPOINT_ID = ?', [id])
 }
 
+async function getParentEndpointRef(db, parentEndpointIdentifier, sessionId) {
```
tecimovic, line 344: I have a feeling that both of these functions could just be one function.
`getParentEndpointRefAndIdentifier` where you retrieve both with a single query. 

Less maintenance overall.

src/boot/ws.js
```diff
@@ -128,7 +128,7 @@ onWebSocket(dbEnum.wsCategory.sessionCreationError, (data) => {
     color: 'negative',
     position: 'top',
     html: true,
-    timeout: 0,
+    timeout: 5000,
```
tecimovic, line 131: ? What are these changes?

src/boot/ws.js
```diff
@@ -155,14 +155,12 @@ onWebSocket(dbEnum.wsCategory.notificationInfo, (data) => {
       color: 'negative',
       position: 'top',
       html: true,
-      timeout: 0,
+      timeout: 5000,
```
tecimovic, line 158: ?

src/store/zap/mutations.js
```diff
@@ -634,7 +649,10 @@ export function updateNotificationCount(state, value) {
 
 export function setDirtyState(state, isDirty) {
   if (state.isDirty != isDirty) {
-    state.isDirty = isDirty;
-    window.parent?.postMessage({eventId: 'dirty', eventData: { isDirty: isDirty }}, '*')
+    state.isDirty = isDirty
```
tecimovic, line 652: Was this file formatted wrong, or are you not using the same formatter?

tecimovic (approved):
1. Add the foreign key constraint.
2. Add a unit test.
3. Make sure regen works.

src-electron/generator/helper-endpointconfig.js
```diff
@@ -93,6 +93,20 @@ function endpoint_fixed_profile_id_array(options) {
   return '{ ' + profileIds.join(', ') + ' }'
 }
 
+/**
+ * Creates array of parent endpoint identifier fields on endpoints
+ *
+ *
+ * @returns C array including the { } brackets
+ */
+function endpoint_fixed_parent_id_array() {
```
brdandu, line 102: Add a block helper which does not depend on language. Leave the language to be decided in the template.

brdandu (changes requested):
There are substantial changes here. Add unit tests to verify that changes are working correctly.

src-electron/db/zap-schema.sql
brdandu: May be an on delete cascade? You may be right though. We probably do not want a user to loose their endpoint altogether.

src-electron/db/query-endpoint.js
brdandu: Also add comments to your function. Is good practice.

src-electron/db/query-impexp.js
```diff
@@ -31,7 +31,13 @@ const queryUpgrade = require('../matter/matter.js')
  * @param {*} endpoint
  * @param {*} endpointTypeRef
```
brdandu, line 32: add function param to comments

src-electron/rest/endpoint.js
```diff
@@ -70,26 +71,48 @@ function httpDeleteEndpointType(db) {
...
+      )
+    }
     let newId = await queryEndpoint.insertEndpoint(
       db,
       sessionId,
       endpointId,
       endpointType,
       networkId,
       profileId,
-      deviceIdentifier
```
brdandu, line 83: This does not look right. Why is device identifier removed here?

brdandu (changes requested):

src-electron/db/zap-schema.sql
paulr34: okay, thanks, will do

src-electron/db/query-endpoint.js
paulr34: yeah, will do, thanks for catching that

src/boot/ws.js
paulr34: bug fix

src/store/zap/mutations.js
paulr34: Not sure, will have to check...

src-electron/generator/helper-endpointconfig.js
paulr34: okay thanks

src-electron/rest/endpoint.js
paulr34: its not used by insertEndpoint because insertEndpoint does a LEFT JOIN from another table

src-electron/db/db-mapping.js
```diff
@@ -448,6 +448,8 @@ exports.map = {
       endpointVersion: x.DEVICE_VERSION, // Left for backwards compatibility
       deviceVersion: x.DEVICE_VERSION,
       deviceIdentifier: x.DEVICE_IDENTIFIER,
+      parentRef: x.PARENT_ENDPOINT_REF,
+      parentEndpointIdentifier: x.PARENT_ENDPOINT_IDENTIFIER,
```
bzbarsky-apple, line 452: As an API consumer...  I assume parentEndpointIdentifier is the endpoint id visible in the spec?  What is parentRef?  Where is any of this documented?

src-electron/generator/helper-endpointconfig.js
```diff
@@ -93,6 +93,20 @@ function endpoint_fixed_profile_id_array(options) {
...
+  return (
+    '{ ' +
+    this.endpoints.map((ep) => ep.parentEndpointIdentifier).join(', ') +
```
bzbarsky-apple, line 105: But not all endpoints have a parent....  Are we using a placeholder value of some sort for "has no parent"?  Something else?

I don't see any tests here exercising this helper.

paulr34: This needs work. 

The idea is the parent value is NULL if there is no parent. 

Working on adding some tests...thanks

src-electron/db/db-mapping.js
paulr34: parentRef is the ENDPOINT_ID of the parent endpoint and parentEndpointIdentifier is the ENDPOINT_IDENTIFIER meaning the "endpoint id visible in the spec" as you mention. parentRef is important for things like checking if an endpoint with identifier x exists in the db...

I will add comments and a documentation page, thank you

src-electron/rest/endpoint.js
brdandu: ok I see that `deviceIdentifier` is not used. Better to fix all the calls to `insertEndpoint` and make sure that only the following arguments are passed to it everywhere in the code.
       db,
       sessionId,
       endpointId,
       endpointType,
       networkId,
       profileId,
       parentEndpointRef

src/boot/ws.js
brdandu: Is there a bug on the button needing a fix?

brdandu (changes requested):

PR #1258, opened 2024-02-05
Fix integer attrib convertion with signed integer type (#1246)

### LINKED ISSUE #1246: ZAP editor should support signed data types
Gecko SDK 4.4.0 with Matter 2.2.0

Issue is described here:

https://community.silabs.com/s/question/0D5Vm000001KPunKAG/zap-editor-signed-field-and-valid-values

ben1hop:


src-electron/validation/validation.js
```diff
@@ -144,22 +149,38 @@ function extractIntegerValue(value) {
...
+    max: getIntegerValue(attribute.max, attribute.type),
   }
 }
 
+function getIntegerValue(attribute, type) {
+  let value = extractIntegerValue(attribute)
+  // if its both a signed type and a hex value: checks for the most significant bit , if its set subtract the 16 bit range from it
+  if (
+    isValidHexString(attribute) &&
+    types.isSignedInteger(type) &&
```
brdandu, line 170: Use something like getSignAndSizeOfZclType instead since it is not hardcoded for integer type strings which could change based on meta data coming from xml. This function is actually searching the database for the types

src-electron/validation/validation.js
brdandu, line 1: Add tests to validation.test.js

src-electron/validation/validation.js
```diff
@@ -144,22 +149,38 @@ function extractIntegerValue(value) {
...
 function getBoundsInteger(attribute) {
   return {
-    min: extractIntegerValue(attribute.min),
-    max: extractIntegerValue(attribute.max),
+    min: getIntegerValue(attribute.min, attribute.type),
+    max: getIntegerValue(attribute.max, attribute.type),
   }
 }
 
+function getIntegerValue(attribute, type) {
```
brdandu, line 165: Use the following format for adding documentation to your functions as a good practice.
```
/** Description
 * 
 * @param {*} param1 
 * @param {*} param2 
 * @returns Integer
 */

**Review by brdandu — CHANGES_REQUESTED** [review]

File: src-electron/validation/validation.js (hunk 1 above)
Comment (ben1hop) [reply src-electron/validation/validation.js:170]: I don't see why is this functions responsibility to decide this. The whole parent function is only handling integer strings , its called in another function which is checked for integer strings. 

In the case the xml is changed , this part is not even going to run.

File: src-electron/validation/validation.js [hunk 3]
```diff
@@ -25,16 +25,29 @@ const queryZcl = require('../db/query-zcl.js')
 const queryConfig = require('../db/query-config.js')
 const queryEndpoint = require('../db/query-endpoint.js')
 const types = require('../util/types.js')
+const queryPackage = require('../db/query-package.js')
 
-async function validateAttribute(db, endpointTypeId, attributeRef, clusterRef) {
```
brdandu, line 29: It is good practice to add function comments for all functions. Add them for all of them while you are at it.

src-electron/validation/validation.js
```diff
@@ -25,16 +25,29 @@ const queryZcl = require('../db/query-zcl.js')
...
+) {
   let endpointAttribute = await queryZcl.selectEndpointTypeAttribute(
     db,
     endpointTypeId,
     attributeRef,
     clusterRef
   )
   let attribute = await queryZcl.selectAttributeById(db, attributeRef)
-  return validateSpecificAttribute(endpointAttribute, attribute)
+  return await validateSpecificAttribute(
```
brdandu, line 44: Why await and not just return directly?

src-electron/validation/validation.js
```diff
@@ -77,15 +96,36 @@ function validateSpecificAttribute(endpointAttribute, attribute) {
...
+          db,
+          zapSessionId,
+          clusterRef
+        ))
+      ) {
         defaultAttributeIssues.push('Out of range')
+      }
     } else {
-      if (!isValidNumberString(endpointAttribute.defaultValue))
+      // TODO this case does the same thing as the one with signed integer
```
brdandu, line 114: Remove todo messages

test/validation.test.js
```diff
@@ -140,56 +224,96 @@ test(
...
   'validate Attribute Test',
-  () => {
+  async () => {
     let fakeEndpointAttribute = {
       defaultValue: '30',
     }
 
     let fakeAttribute = {
-      type: 'UINT16',
+      type: 'int16u',
```
brdandu, line 261: Why is this changed?

brdandu (changes requested):

src-electron/validation/validation.js
ben1hop: i have added to all of my new functions , this one is not mine , i only edited it

src-electron/validation/validation.js
ben1hop: good point , going to remove the await

src-electron/validation/validation.js
ben1hop: i will , however this is not connected to this task and as i could see there are many todos like this are left in the project

this and another function here is in need of a rework , i can open a separate issue for it if you wish

test/validation.test.js
ben1hop: while testing the queryZcl.selectNumberByNameAndClusterId() returned nothing with the upper case type name

my locally built zap tool also uses the lower case format whenever i edit an attribute, why was the other one used at the first place?

brdandu: How was this passing before?
The xml can vary how it represents unsigned 16 bit integers. I believe both representations exist based on matter/zigbee as of today.

src-electron/validation/validation.js
brdandu: To do's in the code are difficult to track. Log issues for these and remove your todos from here.

src-electron/validation/validation.js
brdandu: If you understand the function then please add the description since it is quicker while we are at it. That is how we can strive to make our code more readable but this need not be must do but good to do while at it.

test/validation.test.js
ben1hop: the validation did not check for specific integer types before my implementation, only decided if the input string is an integer and its signed value within bounds

src-electron/validation/validation.js
ben1hop: will do

brdandu (approved):

PR #1259, opened 2024-02-05
Endpoint Composition - Adding helper which has a NULL placeholder + tests

paulr34:
Adding helper which adds Parent Endpoint ID per Endpoint to an array OR NULL if the Endpoint has no Parent

### DIFF
--- a/src-electron/generator/helper-endpointconfig.js
+++ b/src-electron/generator/helper-endpointconfig.js
@@ -102,8 +102,14 @@ function endpoint_fixed_profile_id_array(options) {
 function endpoint_fixed_parent_id_array(options) {
   return (
     '{ ' +
-    this.endpoints.map((ep) => ep.parentEndpointIdentifier).join(', ') +
-    ' }'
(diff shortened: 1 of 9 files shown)


src-electron/generator/helper-endpointconfig.js
```diff
@@ -100,11 +100,15 @@ function endpoint_fixed_profile_id_array(options) {
...
 function endpoint_fixed_parent_id_array(options) {
-  return (
-    '{ ' +
-    this.endpoints.map((ep) => ep.parentEndpointIdentifier).join(', ') +
-    ' }'
-  )
+  let parentIds = []
+  this.endpoints.forEach((ep) => {
+    if (ep.parentEndpointIdentifier == null) {
+      parentIds.push('NULL')
```
tecimovic, line 106: You are hard-coding C `NULL` here. Not the biggest evil, but I thought we have this externalized somewhere?

tecimovic (approved):

PR #1263, opened 2024-02-07
Bug fix editing parent endpoints

paulr34:
Bug fix:

When user edits Endpoint it must return Endpoint Identifier to the GUI and not the database reference

### DIFF
--- a/src-electron/rest/endpoint.js
+++ b/src-electron/rest/endpoint.js
@@ -133,6 +133,29 @@ function httpPatchEndpoint(db) {
   return async (request, response) => {
     let context = request.body
     let sessionIdexport = request.zapSessionId
+    let parentEndpointIdentifier = context.parentEndpointIdentifier
+    let parentEndpointRef = await queryEndpoint.getParentEndpointRef(
+      db,
+      parentEndpointIdentifier,
+      sessionIdexport
+    )
+    if (parentEndpointRef == null && parentEndpointIdentifier != null) {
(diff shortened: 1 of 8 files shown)


src-electron/rest/endpoint.js
```diff
@@ -133,6 +133,29 @@ function httpPatchEndpoint(db) {
...
     let sessionIdexport = request.zapSessionId
+    let parentEndpointIdentifier = context.parentEndpointIdentifier
+    let parentEndpointRef = await queryEndpoint.getParentEndpointRef(
+      db,
+      parentEndpointIdentifier,
+      sessionIdexport
+    )
+    if (parentEndpointRef == null && parentEndpointIdentifier != null) {
+      parentEndpointIdentifier = null
+      notification.setNotification(
```
tecimovic, line 144: You should immediately return the 404 or some other HTTP error here. This is not what we should use notifications for!

Notifications are for asynchronous issues. For immediate failures you should immediatelly send back the appropriate HTTP error code.

tecimovic (changes requested):

src-electron/rest/endpoint.js
paulr34: good idea, will do, thanks

paulr34: @tecimovic I changed it so it errors if the user chooses a Parent that does not exist.

That said, I left in the Notification so the user gets a pop up message stating why...

PR #1264, opened 2024-02-14
await sort

paulr34:
importing needs to wait for the endpoints to sort before logic that depends on sorting

when importing an endpoint with a parent ref, the parent endpoint must already be imported for the endpoint reference to be found

### DIFF
--- a/src-electron/importexport/import-json.js
+++ b/src-electron/importexport/import-json.js
@@ -378,7 +378,7 @@ async function importPackages(
  * @param {*} endpoints
  * @returns list or sorted endpoints based on endpoint type index
  */
-function sortEndpoints(endpoints) {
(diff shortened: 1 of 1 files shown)


src-electron/importexport/import-json.js
```diff
@@ -378,7 +378,7 @@ async function importPackages(
  * @param {*} endpoints
  * @returns list or sorted endpoints based on endpoint type index
  */
-function sortEndpoints(endpoints) {
+async function sortEndpoints(endpoints) {
```
bzbarsky-apple, line 381: I don't get it.  There is nothing async in this function.  Why is it being made async?  This feels like it's covering up a bug somewhere else....

paulr34: agree

PR #1270, opened 2024-02-28
ZAP generation js helper

pgregorr-amazon:
This change in helper.js asObjectiveCClass function, enables the use of an optional struct type prefix for ZAP generation. It’s needed to generate Objective-C class names with the “MC” prefix, which are specific to Matter Casting, tv-casting-app example. 

**Change summary**

**Note:** Most of the changes are whitespace changes made by the git commit.

1.	Added a check for an struct type prefix configuration through the options object. If the new structTypePrefix option is not specified, the default “MTR” struct type prefix will be used in the  asObjectiveCClass ZAP helper function.
2.	Specifying structTypePrefix="MC" in {{asObjectiveCClass type cluster structTypePrefix="MC"}} will generate a struct type with the “MC” prefix.

**Testing**

Verified and tested ZAP generation locally using “./scripts/tools/zap_regen_all.py“ in project-chip/connectedhomeip repo. The Objective-C class names in files outside of the tv-casting-app example, where no structTypePrefix, is specified, are generated with the “MTR” prefix. tv-casting-app example app Objective-C class names are generated with the MC prefix.

### DIFF
--- a/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
+++ b/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
@@ -264,14 +264,8 @@ async function asObjectiveCClass(type, cluster, options) {
         hash: { preserveAcronyms: preserveAcronyms },
       });
     }
-    // Use a custom prefix if specified.
-    if (options.hash.structTypePrefix) {
-      return (
-        options.hash.structTypePrefix +
-        `${cluster}Cluster${appHelper.asUpperCamelCase(type)}`
-      );
-    }
-    return `MTR${cluster}Cluster${appHelper.asUpperCamelCase(type)}`;
+    // Use a custom prefix if specified, otherwise default to "MTR" for backwards compat.
+    return `${options.hash.structTypePrefix || "MTR"}${cluster}Cluster${appHelper.asUpperCamelCase(type)}`;
   }
 
   return 'NSNumber';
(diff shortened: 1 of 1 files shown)


src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -262,6 +264,13 @@ async function asObjectiveCClass(type, cluster, options) {
         hash: { preserveAcronyms: preserveAcronyms },
       });
     }
+    // Use a custom prefix if specified.
+    if (options.hash.structTypePrefix) {
+      return (
+        options.hash.structTypePrefix +
+        `${cluster}Cluster${appHelper.asUpperCamelCase(type)}`
+      );
+    }
```
sharadb-amazon, line 273: I believe this is the only "real" change in this diff and everything else is a whitespace change?

sharadb-amazon (approved):

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
pgregorr-amazon: That is correct. During git commit, Pretty quick reformat... introduced the whitespace changes

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -262,6 +264,13 @@ async function asObjectiveCClass(type, cluster, options) {
...
     return `MTR${cluster}Cluster${appHelper.asUpperCamelCase(type)}`;
```
bzbarsky-apple, line 274: How about this instead:
```suggestion
    // Use a custom prefix if specified, otherwise default to "MTR" for backwards compat.
    return `${options.hash.structTypePrefix || "MTR"}${cluster}Cluster${appHelper.asUpperCamelCase(type)}`;
```

paulr34 (approved):

paulr34 (approved):
approve  under assumption that tests pass

bzbarsky-apple (approved):

PR #1274, opened 2024-03-07
Fix UI cluster closing issues while using text based filtering

ben1hop:
Fix the issue on the UI when using text filtering neither the close all button nor the clusters itself couldn't open and close the clusters correctly.

### DIFF
--- a/src/components/ZclClusterManager.vue
+++ b/src/components/ZclClusterManager.vue
@@ -57,6 +57,7 @@ limitations under the License.
             dense
             @update:model-value="changeDomainFilter($event)"
             data-test="filter-input"
+            data-cy="cluster-domain-filter"
           />
         </div>
 
@@ -70,6 +71,7 @@ limitations under the License.
             @clear="changeFilterString('')"
             :model-value="filterString"
             data-test="search-clusters"
+            data-cy="cluster-text-filter"
           >
             <template v-slot:prepend>
               <q-icon name="search" />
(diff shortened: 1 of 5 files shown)


src/store/zap/mutations.js
```diff
@@ -488,23 +488,22 @@ export function setDomainFilter(state, filterEnabledClusterPair) {
...
-      ),
+      value: openDomainValue,
     })
   })
 }
 
 export function doActionFilter(state, filterEnabledClusterPair) {
   let filter = filterEnabledClusterPair.filter
   // When we close all, we also clear all filters.
-  resetFilters(state)
```
ben1hop, line 507: a close all button should act only for closing the opened clusters , not clearing only one of the two filter options

src/components/ZclClusterManager.vue
```diff
@@ -301,7 +301,7 @@ export default {
       })
     },
     getDomainOpenState(domainName) {
-      return this.openDomains[domainName] || this.filterString != ''
```
ben1hop, line 304: this extra condition blocks the closing actions on the UI

src/store/zap/mutations.js
```diff
@@ -488,23 +488,22 @@ export function setDomainFilter(state, filterEnabledClusterPair) {
   let filter = filterEnabledClusterPair.filter
   state.clusterManager.filter = filter
   state.domains.map((domainName) => {
+    const openDomainValue =
+      state.clusterManager.filterString === ''
+        ? filter.domainFilterFn(domainName, state.clusterManager.openDomains, {
+            enabledClusters: filterEnabledClusterPair.enabledClusters,
+          })
+        : true
```
ben1hop, line 496: checking the existence of the filter string here instead , if its not missing then open the clusters just as it before

PR #1278, opened 2024-03-07
matter centric UI

paulr34:
separating Zigbee and Matter

### DIFF
--- a/cypress/e2e/clusters/cluster-filter.cy.js
+++ b/cypress/e2e/clusters/cluster-filter.cy.js
@@ -30,29 +30,6 @@ describe('Testing cluster filters', () => {
     }
   )
   it(
-    'enable power configuration cluster',
-    { retries: { runMode: 2, openMode: 2 } },
-    () => {
-      cy.get('[data-test="filter-input"]').click({ force: true })
-      cy.get('.q-virtual-scroll__content > :nth-child(1)').click()
-      cy.fixture('data').then((data) => {
-        cy.get('tbody').children().should('contain', data.cluster2)
-      })
-      cy.get('#General').click({ force: true })
-      cy.get(
-        '#General > .q-expansion-item__container > .q-expansion-item__content > .q-card > .q-card__section > .row >  .q-table__container > .q-table__middle > .q-table > tbody > :nth-child(2) > :nth-child(6) > .q-field > .q-field__inner > .q-field__control'
-      ).click({ force: true })
-      cy.fixture('data').then((data) => {
-        cy.get('.q-virtual-scroll__content > :nth-child(3)')
-          .contains(data.server1)
-          .click()
-      })
-      cy.get(
-        '.v-step-7 > .q-field > .q-field__inner > .q-field__control'
-      ).click({ force: true })
-    }
-  )
-  it(
     'checks if power configuration exists',
     { retries: { runMode: 2, openMode: 2 } },
     () => {
--- a/cypress/e2e/clusters/cluster-filter.cy.js
+++ b/cypress/e2e/clusters/cluster-filter.cy.js
@@ -29,16 +29,4 @@ describe('Testing cluster filters', () => {
       })
     }
   )
-  it(
-    'checks if power configuration exists',
-    { retries: { runMode: 2, openMode: 2 } },
-    () => {
-      cy.get('.q-virtual-scroll__content > :nth-child(2)').click({
-        force: true,
-      })
-      cy.fixture('data').then((data) => {
-        cy.get('tbody').children().should('contain', data.cluster2)
-      })
-    }
-  )
 })
--- a/src/components/ZclAttributeManager.vue
+++ b/src/components/ZclAttributeManager.vue
@@ -123,8 +123,14 @@ limitations under the License.
               "
             />
           </q-td>
-          <q-td key="singleton" :props="props" auto-width>
+          <q-td
+            key="singleton"
+            v-if="enableSingleton"
+            :props="props"
+            auto-width
+          >
             <q-checkbox
+              v-if="enableSingleton"
               class="q-mt-xs"
               :model-value="selectionSingleton"
               :val="hashAttributeIdClusterId(props.row.id, selectedCluster.id)"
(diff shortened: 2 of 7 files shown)


paulr34:
@brdandu @tecimovic @ben1hop 

Two things: 
1. Since I removed the option to Enable "Client" for a Matter cluster, the Cypress tests diverged from Zigbee. We are either blocked by this or we delete the tests temporarily until we successfully separate Matter and Zigbee Cypress tests. 
2. If a User incorrectly loads a .zap file with a Cluster Enabled as a "Client", should we automatically change it to "Server" for them?

src/components/ZclAttributeManager.vue
```diff
@@ -410,97 +417,115 @@ export default {
...
       forcedExternal: [],
+      enableSingleton: false,
+      enableBounded: false,
     }
   },
+  mounted() {
+    const enableZigbeeFeatures =
+      this.$store.state.zap.selectedZapConfig?.zclProperties?.category ===
+      'zigbee'
+    this.enableSingleton = enableZigbeeFeatures
```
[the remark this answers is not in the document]
paulr34: It comes from the .zap file already so why implement the same thing twice?

cypress/e2e/clusters/cluster-filter.cy.js
```diff
@@ -29,39 +29,4 @@ describe('Testing cluster filters', () => {
       })
     }
   )
-  it(
-    'enable power configuration cluster',
```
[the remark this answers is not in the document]
paulr34: Because the two GUI are diverging and need their own tests

src/components/ZclAttributeManager.vue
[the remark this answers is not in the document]
paulr34: How is it more scalable?

- It comes from the .zap file and gsdk already
- We already use it, for example, when a user selects zcl.json from the zap config page

cypress/e2e/clusters/cluster-filter.cy.js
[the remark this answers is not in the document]
paulr34: @brdandu we run the same tests for Zigbee and Matter. We need to fix that.

src/components/ZclAttributeManager.vue
```diff
@@ -410,97 +417,115 @@ export default {
...
-        },
-      ],
+      columns: [],
       forcedExternal: [],
+      enableSingleton: false,
+      enableBounded: false,
     }
   },
+  mounted() {
+    const enableZigbeeFeatures =
```
tecimovic, line 427: I always complain if UI is polluted by words "zigbee" and "matter". UI should be driven by "features" and which of those are "zigbee" and which are "matter" is defined in the `zcl.json.

tecimovic: So the way this should be implemented is not like this...

You should do the following:
   1.) Matter UI options should say: "singletonsEnabled=false" inside matter zcl.json file.
   2.) ZIgbee UI options should say: "singletonsEnabled=true" inside zigbee zcl.json file.
   3.) UI should then simply act on this data, not have this hardcoded business logic of:
      if category is matter, then disable singletons.

src/components/ZclCreateModifyEndpoint.vue
```diff
@@ -254,7 +255,11 @@ export default {
     } else {
       this.shownEndpoint.endpointIdentifier = this.getSmallestUnusedEndpointId()
     }
-
+    const enableZigbeeFeatures =
```
tecimovic, line 258: Same comment as earlier. Please stop polluting UI code with "zigbee" and "matter".

src/components/ZclDomainClusterView.vue
```diff
@@ -232,6 +232,24 @@ export default {
   props: ['domainName', 'clusters'],
   mixins: [CommonMixin],
   computed: {
+    clusterSelectionOptions() {
+      if (
+        this.$store.state.zap.selectedZapConfig?.zclProperties?.category ===
+        'matter'
```
tecimovic, line 238: Ditto

tecimovic (changes requested):
UI should be feature driven, not category driven.
All you did here was put a lot of "business logic" into UI. We should not do that.

UI should be driven from "features".
"Features" should be mapped to categories in zcl.json files.

UI should NEVER EVER say:
   "if category == zigbee then do this, else do that"
UI should ONLY react to values of UI options loaded into the DB.

src/components/ZclAttributeManager.vue
paulr34: the logic is the same either way

src/components/ZclDomainClusterView.vue
paulr34: yeah, I am about to push a fix to centralize this but as discussed:
1.) there is no avoiding conditionals
2.) zcl.json changes would be much harder for a ZAP developer to make 
3.) we already implement this from a .zap file so if we want to do a redesign then it would have to be a new ticket

src/components/ZclCreateModifyEndpoint.vue
paulr34: so instead of doing a high level conditional that we can derive data from, you suggest doing a conditional for every single component?

paulr34:
> UI should be feature driven, not category driven. All you did here was put a lot of "business logic" into UI. We should not do that.
> 
> UI should be driven from "features". "Features" should be mapped to categories in zcl.json files.
> 
> UI should NEVER EVER say: "if category == zigbee then do this, else do that" UI should ONLY react to values of UI options loaded into the DB.

I don't agree because if we know the GUI is a specific protocol that implies data which means we would be doing unnecessary work retrieving information that is already implied.

src/util/ui-options.js
```diff
@@ -0,0 +1,31 @@
+// mixin.js
```
tecimovic, line 1: This is in public space. Whenever you create a new file, you need to put that Copyright blah blah blah stanza up on top.

tecimovic (approved):

src/util/ui-options.js
paulr34: Oh thank you! I'll add that!

PR #1287, opened 2024-03-26
bug fix - duplicating endpoints

paulr34:
Fixing bug when selecting API_MATURITY from a table where it does not exist

: Error: SQLITE_ERROR: no such column: API_MATURITY"}
⇝ (node:2539) UnhandledPromiseRejectionWarning: Error: SQLITE_ERROR: no such column: API_MATURITY

### DIFF
--- a/src-electron/db/query-attribute.js
+++ b/src-electron/db/query-attribute.js
@@ -54,8 +54,7 @@ async function selectEndpointTypeAttributesByEndpointTypeRefAndClusterRef(
       INCLUDED_REPORTABLE,
       MIN_INTERVAL,
       MAX_INTERVAL,
-      REPORTABLE_CHANGE,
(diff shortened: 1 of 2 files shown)


src-electron/db/query-attribute.js
```diff
@@ -54,8 +54,7 @@ async function selectEndpointTypeAttributesByEndpointTypeRefAndClusterRef(
       INCLUDED_REPORTABLE,
       MIN_INTERVAL,
       MAX_INTERVAL,
-      REPORTABLE_CHANGE,
-      API_MATURITY
+      REPORTABLE_CHANGE
```
brdandu, line 57: I have this change on another PR. This column does not exist in the table.

brdandu (approved):

PR #1288, opened 2024-03-28
Studio V6 Integration

paulr34:
changing session storage depending on what is given from parent iframe

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -141,6 +141,46 @@ export default defineComponent({
     },
   },
   methods: {
+    parseQueryString() {
+      const querystring = require('querystring')
+      let search = window.location.search
+
+      if (search[0] === '?') {
+        search = search.substring(1)
+      }
+      this.query = querystring.parse(search)
+    },
+
+    setSessionUuid() {
+      if (window.sessionStorage.getItem('session_uuid') == null) {
+        window.sessionStorage.setItem('session_uuid', uuidv4())
+      }
+      if (this.query[`stsApplicationId`]) {
+        let currentSessionUuid =
+          window.sessionStorage.getItem('session_uuid') || ''
+        let updatedSessionUuid =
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -249,6 +251,17 @@ export default defineComponent({
     },
   },
   created() {
+    // Parse the query string into the front end.
+    const querystring = require('querystring')
```
tecimovic, line 255: I don't like stray `require` sitting in the code like this. It should be at the top of the file.

tecimovic (changes requested):
I think there is more to this....

We have logic that allows to attach to the old session. Where is that logic? Is that logic affected?

With that "setItem('session_id')", we're again "spreading around" the session id logic. I would like all logic that touches the session id to be accumulated in one place, not having to hunt it down between the place where we have it no, to here.

src/App.vue
paulr34: okay will change, thanks

src/App.vue
```diff
@@ -249,6 +251,17 @@ export default defineComponent({
...
+    let search = window.location.search
+
+    if (search[0] === '?') {
+      search = search.substring(1)
+    }
+    let query = querystring.parse(search)
+    if (query[`stsApplicationId`]) {
+      window.sessionStorage.setItem('session_uuid', query[`stsApplicationId`])
```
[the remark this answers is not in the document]
paulr34: okay I addressed this concern, lmk what you think

paulr34:
> I think there is more to this....
> 
> We have logic that allows to attach to the old session. Where is that logic? Is that logic affected?
> 
> With that "setItem('session_id')", we're again "spreading around" the session id logic. I would like all logic that touches the session id to be accumulated in one place, not having to hunt it down between the place where we have it no, to here.

Fixed

PR #1290, opened 2024-03-29
further cleanup and documentation

paulr34:


src-electron/sdk/matter.js
```diff
@@ -110,17 +106,29 @@ async function computeStorageOptionNewConfig(storagePolicy) {
   }
   return storageOption
 }
-
+/**
+ * This asynchronous function computes and returns the new configuration for a storage policy.
+ *
+ * @param {Object} db - The database instance.
+ * @param {Object} clusterRef - The reference to the cluster.
```
brdandu, line 113: clusterRef -> Number?

paulr34: hm. you're right

src-electron/sdk/matter.js
```diff
@@ -110,17 +106,29 @@ async function computeStorageOptionNewConfig(storagePolicy) {
...
   clusterRef,
   storagePolicy,
   forcedExternal,
   attributeName
 ) {
-  let clusterName
-  clusterName = await queryCluster.selectClusterName(db, clusterRef)
-  forcedExternal.map((option) => {
+  let clusterName = await queryCluster.selectClusterName(db, clusterRef)
+  forcedExternal.some((option) => {
```
tecimovic, line 130: I don't get this .some(). Why are we using .some()? 
why are we not doing `for ( ... of ...)` ?

tecimovic: If we are using .some(), we should properly return a value back.

tecimovic (approved):
Some minor issues, but generally fine....

brdandu (approved):

PR #1291, opened 2024-03-29
general cleanup

paulr34:
With a lot to do in a short amount of time, sometimes things get a little messy. 

This PR is an attempt to cleanup the code by modularizing and creating/following quality standards.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -42,11 +42,16 @@ const observable = require('./util/observable.js')
 const dbEnum = require(`../src-shared/db-enum.js`)
 const storage = require('./util/storage.js')
 
-window.addEventListener('message', (event) => {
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -42,83 +42,82 @@ const observable = require('./util/observable.js')
...
           window[rendApi.GLOBAL_SYMBOL_EXECUTE](rendApi.id.save)
         }
         break
     }
-  }, 
+  },
   false
 )
 
-async function initLoad(store) {
```
tecimovic, line 61: This whole set of  changes seems counter-productive. It's not doing the RIGHT THING', it's only doing a different thing.

tecimovic (changes requested):

PR #1296, opened 2024-04-03
Cleaning up the custom xml UI update as well as the custom xml package uploading such that there are no duplicates

brdandu:
- JIRA: ZAPP-1351


paulr34 (approved):

paulr34 (dismissed):

brdandu:
> cluster extensions(attributes, cmd) are not showing up in the UI. Looking into it.

fixed

brdandu:
remaining tasks: Add unit tests for custom xml

brdandu:
See https://github.com/project-chip/zap/pull/1299 PR instead since that has the fix on top of the multi protocol revert

PR #1299, opened 2024-04-08
Bug/fix custom xml issues/zapp 1351

brdandu:
Reverting the multi protocol commit and fixing custom xml issues on top of it.


src-electron/db/query-endpoint.js
```diff
@@ -58,6 +58,75 @@ ORDER BY E1.ENDPOINT_IDENTIFIER
...
+  E1.PROFILE,
+  E1.ENDPOINT_IDENTIFIER,
+  E1.NETWORK_IDENTIFIER,
+  PACKAGE.CATEGORY,
+  E2.ENDPOINT_ID AS PARENT_ENDPOINT_REF,
+  E2.ENDPOINT_IDENTIFIER AS PARENT_ENDPOINT_IDENTIFIER
+FROM
+  ENDPOINT AS E1
+LEFT JOIN
+  ENDPOINT AS E2
```
tecimovic, line 91: I got confused by now.... Why exactly do we need the parent-child relationship in the endpoints? What does a "parent endpoint" mean?

test/server-bare.test.js
```diff
@@ -145,9 +145,34 @@ describe('Session specific tests', () => {
         .insertPathCrc(db, 'PATH', 32, dbEnum.packageType.zclProperties)
         .then((pkg) => {
           packageId = pkg
+          return packageId
         })
         .then(() =>
-          queryPackage.insertSessionPackage(db, sessionId, packageId)
```
tecimovic, line 150: We should rewrite these tests to use await/async instead of the old .then() patterns...

tecimovic (approved):
There is a LOT of change here. I am not able to fully review this line by line, but generally, if you've covered all the unit tests with odd custom XML cases, and they are working as expected, then I'm ok with this.

src-electron/main-process/startup.js
```diff
@@ -584,20 +584,27 @@ async function generateSingleFile(
...
-    throw new Error(`Generation failed: ${zapFile}`)
+    if (genResult.hasErrors) {
+      console.log(JSON.stringify(genResult.errors))
+      throw new Error(`Generation failed: ${zapFile}`)
+    }
+    genResults.push(genResult)
   }
 
-  return genResult
+  return genResults
```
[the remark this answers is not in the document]
brdandu: For multi-protocol there can be more than one top level generation packages. Hence the array. Each top level template package can have a gen template package associated with it.

src-electron/db/query-endpoint.js
brdandu: An endpoint which can be a composition of other endpoints. So that relationship is established by parent endpoints

test/server-bare.test.js
brdandu: I agree. We should probably do that for our entire code base.

brdandu: Logged https://github.com/project-chip/zap/issues/1302

PR #1300, opened 2024-04-08
Feature/feature map attribute and device type conformance/zapp 1346

brdandu:
This feature provides automatic setting of feature map attribute based on mandatory features from the device type conformance. The changes should also warn existing app users with warnings during .zap file open


brdandu:
Corresponding CHIP repo PR: https://github.com/project-chip/connectedhomeip/pull/32930

paulr34 (approved):

PR #1301, opened 2024-04-10
Externalize Helpers framework

paulr34:


paulr34:
@tecimovic TODO: add sessionId and db to context object

test/gen-template/matter/external-addon-helper.js
```diff
@@ -0,0 +1,31 @@
...
+// This is an example of an external addon helper for templates.
+
+async function test_external_addon_helper(api, context) {
+  //add queries here
+  return 'This is example of test external addon helper.'
+}
+function initialize_helpers(api, context) {
+  api.registerHelpers(
+    'test_external_addon_helper',
+    test_external_addon_helper.bind(null, api, context),
```
tecimovic, line 26: I don't understand this. You are binding the function with `null` as this, and passing api and context. Why are you creating a bound function here? Why not just use the actual function? I assumed we will have a wrapper function here, not a bound instance of the same function...

test/gen-template/matter/external-addon-helper.js
```diff
@@ -0,0 +1,31 @@
...
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+// This is an example of an external addon helper for templates.
+
+async function test_external_addon_helper(api, context) {
+  //add queries here
+  return 'This is example of test external addon helper.'
```
tecimovic, line 21: You have a very simplistic unit test, simply returning a constant string. You also bind function, so I'm very lost as to what is the actual API and context that is passed to this function.

Can you create a unit test that uses some queries from the `api` and gets some data from the context?

Also note, that the helpers have their own "hash" properties passed. It would be nice to see if that mechanism works fully.

Suggestion:
   create a helper that actually does something real-like, like retrieve a "longest cluster name" or something similar, and then test it with that data. Then also make it add a `key="value"` type of argument, that will modify its behavior.

tecimovic (commented):
I would like a unit test that registers a helper that is actually using some of the API functions that were asker for:
   - retrieve ALL clusters/events/attributes/commands
   - retrieve SELECTED clusters/events/attributes/commands
etc.

There is also that issue of binding a function that I commented in the code. Not 100% I understand that use...

test/gen-template/matter/external-addon-helper.js
paulr34: Okay will do

test/gen-template/matter/external-addon-helper.js
paulr34: Ok I'll look into it, thanks

test/gen-template/matter/external-addon-helper.js
paulr34: done

paulr34:
@tecimovic thanks for your review, I addressed your concerns 👍

brdandu (changes requested):

test/gen-template/matter/external-addon-helper.js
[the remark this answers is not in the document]
paulr34: Good point, but I am just following the current setup of helper files. In order to achieve common files for both Matter and Zigbee we would have to change how all the add-on helpers are organized. 

We can do a reorg in a separate PR , agree?

test/gen-template/matter/external-addon-helper.js
```diff
@@ -0,0 +1,74 @@
...
+  let attributes = await api.availableAttributes(this)
+  let totalAttributes = attributes.length
+  return totalAttributes
+}
+async function test_external_addon_all_commands_helper(api) {
+  let commands = await api.availableCommands(this)
+  let totalCommands = commands.length
+  return totalCommands
+}
+async function test_external_addon_all_clusters_helper(api) {
```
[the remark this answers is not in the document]
paulr34: okay, I would keep "this" until someone said otherwise

src-electron/util/script-api.js
```diff
@@ -523,7 +580,48 @@ async function enableOutgoingCommand(
...
+          thrownObject = err
+          thrownObject.message = err.message + locMsg
+        }
+      }
+      throw thrownObject
+    }
+  }
+}
 
+function registerHelpers(singleHelper, registerHelper, context, api) {
```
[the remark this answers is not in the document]
paulr34: @tecimovic I think I made the changes you wanted. Lmk..

test/gen-template/matter/external-addon-helper.js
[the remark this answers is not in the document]
paulr34: instead of context it gets "this". At first, I tried passing "context" but there are lots of helper functions and helpers of helper functions that are designed to deal with "this" so for now it seemed better to stick with it.

Moving helpers in general from "this" to context will require more work than just passing in as an argument

src-electron/util/helper-api.js
```diff
@@ -0,0 +1,91 @@
+/**
+ *
+ *
+ *    Copyright (c) 2021 Silicon Labs
```
tecimovic, line 4: It's 2024.

src-electron/util/helper.js
```diff
@@ -0,0 +1,45 @@
+/**
+ *
+ *    Copyright (c) 2021 Silicon Labs
```
tecimovic, line 3: It's 2024.

test/gen-template/matter/external-addon-helper.js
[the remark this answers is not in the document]
tecimovic: I don't want to convert ALL the helpers to use context, just the "external" ones.
True, the Handlebars is using "this" heavily and I always found that use arcane, and it was always confusing what "this" was, like it always is with Javascript.

That's why I hoped that for the external helpers, instead of letting users guess what "this" is, we don't even bother them with it, since we are wrapping these helpers anyway.

But it's fine temporarily to use "this" and then before you hand this out to final users, change "this" into an explicit "context" argument.

As a general rule: if you can avoid making people use "this", avoid it.
But Handlebars is making you use "this", so for our internal helpers, we should continue using "this" and just be sad about it.

test/gen-template/zigbee/external-addon-helper.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+// This is an example of an addon helper for templates.
+
+function test_external_addon_helper(api, context) {
+  return 'This is example of test external addon helper.'
+}
+function initialize_helpers(api, context) {
+  api.registerHelpers(
+    'test_external_addon_helper',
+    test_external_addon_helper.bind(null, api, context),
```
tecimovic, line 25: I thought we removed the .bind()?

tecimovic (commented):
I would like to add a file "external-helpers.md" into the doc/ directory, where you write up the "Developers guide to using external helpers". Then you simply point whoever wants to write external helpers to that file. You describe what do you need to do to have an external helper and link to generated API.

tecimovic (approved):

test/gen-template/matter/external-addon-helper.js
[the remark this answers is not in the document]
paulr34: ok changed from "this" to "context"

PR #1305, opened 2024-04-16
changing to correct name

paulr34:


brdandu (approved):

apack.json
brdandu, line 1: There is one more reference in ZclCustomZclView `You can use this functionality to add custom ZCL clusters or
            commands to the Zigbee Clusters Configurator`

brdandu (approved):

apack.json
paulr34: ok thanks

tecimovic (approved):

PR #1308, opened 2024-04-18
Make options and extensions separate pages instead of dialogs like the rest of the menu items

ben1hop:


paulr34 (approved):

test/ui.test.js
```diff
@@ -162,14 +162,14 @@ describe('Component mounting test', () => {
...
+    'Options',
     () => {
-      const wrapper = shallowMount(ZclGeneralOptionsBar, {
+      const wrapper = shallowMount(Options, {
         global: {
           plugins: [ZapStore()],
         },
       })
-      expect(wrapper.html().length).toBeGreaterThan(100)
+      expect(wrapper.html().length).toBeGreaterThan(50)
```
brdandu, line 172: What is this count?

test/ui.test.js
```diff
@@ -217,7 +217,7 @@ describe('Component mounting test', () => {
...
```
brdandu, line 220: same what is this count?

test/ui.test.js
ben1hop: no idea but didnt seem that important and they were failing

brdandu: could you debug and check what is under wrapper just to make sure this is not anything to be worried about?

ben1hop: its a simple vue wrapper

ben1hop: so this is checks for the vue components html code length , this is reduced because its a new page now not a dialog in another component

test/ui.test.js
ben1hop: same here , since the extensions dialog is removed the components html string length will be shorter

PR #1310, opened 2024-04-23
Studio V6 Integration- updating  Dirty State and saving files

paulr34:
Updating Dirty State and sending the status to Studio V6 via Post Message

### DIFF
--- a/src-shared/rest-api.js
+++ b/src-shared/rest-api.js
@@ -34,7 +34,7 @@ const uri = {
   deletePackageNotification: 'deletePackageNotification',
   duplicateEndpoint: '/duplicateEndpoint',
   duplicateEndpointType: '/duplicateEndpointType',
-  dirtyFlag: '/dirtyFlag',
+  getDirtyFlag: '/getDirtyFlag',
   option: '/option',
   uiOptions: '/uiOptions',
   commandUpdate: '/command/update',
--- a/src/store/zap/actions.js
+++ b/src/store/zap/actions.js
@@ -833,8 +833,10 @@ export function updateSelectedUcComponentState(context, projectInfo) {
   })
 }
 
-export function setDirtyState(context, isDirty) {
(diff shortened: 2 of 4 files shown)


src-shared/rest-api.js
```diff
@@ -34,7 +34,7 @@ const uri = {
   deletePackageNotification: 'deletePackageNotification',
   duplicateEndpoint: '/duplicateEndpoint',
   duplicateEndpointType: '/duplicateEndpointType',
-  dirtyFlag: '/dirtyFlag',
```
tecimovic, line 37: Why has the previous API gotten removed?

src/store/zap/actions.js
```diff
@@ -833,8 +833,10 @@ export function updateSelectedUcComponentState(context, projectInfo) {
   })
 }
 
-export function setDirtyState(context, isDirty) {
-  context.commit('setDirtyState', isDirty)
+export function setDirtyState(context) {
```
tecimovic, line 836: This makes little sense.... Why are we putting dirty state into the store?

tecimovic (commented):
I don't understand this change. What are we trying to achieve?

src/store/zap/actions.js
paulr34: this was previously written code that sends the dirty state value in the store to Studio V6

the value was never being updated 

this PR adds a REST_API which updates it

src-shared/rest-api.js
paulr34: adding back. I thought it was not used but it is in the studio-rest-api files..

paulr34:
> I don't understand this change. What are we trying to achieve?

Previously, we adding Studio V6 integration code which sends a postMessage to the parent DOM, Studio V6. This message tells Studio the state of the ZAP file. There was a bug where the state was never changing because we were using websocket incorrectly.

PR #1311, opened 2024-04-24
bug fix-Matter GUI

paulr34:
Matter GUI should allow client clusters but not show attributes or commands if a client cluster

### DIFF
--- a/cypress/e2e/attributes/check-required-attributes.cy.js
+++ b/cypress/e2e/attributes/check-required-attributes.cy.js
@@ -16,22 +16,4 @@ describe('Testing disabling enabled attributes', () => {
       cy.gotoAttributePage(data.endpoint1, data.cluster1)
     })
   })
-  it(
-    'getting enabled attribute and disable it',
-    { retries: { runMode: 2, openMode: 2 } },
-    () => {
-      cy.get(
-        '.table_body:first > :nth-child(2) > .q-mt-xs > .q-toggle__inner'
-      ).click()
-      cy.get(
-        '.table_body:first > :nth-child(2) > .q-mt-xs > .q-toggle__inner'
-      ).should('be.visible')
-      cy.get(
-        '.table_body:first > :nth-child(2) > .q-mt-xs > .q-toggle__inner'
-      ).click()
-      cy.contains(
-        '.table_body:first > :nth-child(2) > .q-mt-xs > .q-toggle__inner'
-      ).should('not.exist')
-    }
-  )
 })
--- a/src/components/ZclAttributeManager.vue
+++ b/src/components/ZclAttributeManager.vue
@@ -39,7 +39,11 @@ limitations under the License.
         <q-tr
           :props="props"
           class="table_body"
-          v-if="!globalLists.includes(props.row.label)"
+          v-if="
+            (!globalLists.includes(props.row.label) &&
+              props.row.side === 'server') ||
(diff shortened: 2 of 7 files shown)


src/components/ZclAttributeManager.vue
```diff
@@ -39,7 +39,11 @@ limitations under the License.
         <q-tr
           :props="props"
           class="table_body"
-          v-if="!globalLists.includes(props.row.label)"
+          v-show="
+            (!globalLists.includes(props.row.label) &&
+              props.row.side === 'server') ||
```
brdandu, line 44: This should be matter only change. Other protocols such as Zigbee should not be affected by this. Is that the case here?

cypress/e2e/attributes/check-required-attributes.cy.js
```diff
@@ -16,22 +16,4 @@ describe('Testing disabling enabled attributes', () => {
       cy.gotoAttributePage(data.endpoint1, data.cluster1)
     })
   })
-  it(
```
brdandu, line 19: Why is this UI test removed?

src/components/ZclAttributeReportingManager.vue
```diff
@@ -29,7 +29,10 @@ limitations under the License.
       data-cy="Attributes Reporting"
     >
       <template v-slot:body="props">
-        <q-tr :props="props">
+        <q-tr
+          :props="props"
+          v-show="props.row.side === 'server' || !enableServerOnly"
+        >
```
brdandu, line 35: Does matter have attribute reporting? Should this whole vue component not show up for matter?

src/components/ZclCommandManager.vue
```diff
@@ -35,7 +35,10 @@ limitations under the License.
           </q-tr>
         </template>
         <template v-slot:body="props">
-          <q-tr :props="props">
+          <q-tr
+            :props="props"
+            v-show="props.row.source === 'server' || !enableServerOnly"
```
brdandu, line 40: Should we just be showing commands which have source as server?
Are there no response commands which should be shown when command source is client?

brdandu (changes requested):

src/components/ZclAttributeManager.vue
paulr34: it does do that `            (!globalLists.includes(props.row.label) &&
              props.row.side === 'server') ||
            !enableServerOnly`

cypress/e2e/attributes/check-required-attributes.cy.js
paulr34: its checking for a client side attribute which means it is an incorrectly written test and should be remade in a separate PR

src/components/ZclAttributeReportingManager.vue
paulr34: good point, ill add that

src/components/ZclCommandManager.vue
paulr34: I thought that it was discussed we don't need to show Client commands but not 100 percent sure

cypress/e2e/attributes/check-required-attributes.cy.js
brdandu: Shouldn't it still be tested for protocols other than matter?

paulr34:
going with a different design

PR #1313, opened 2024-05-01
GUI_fix

paulr34:
GUI fix

### DIFF
--- a/src-electron/db/query-session.js
+++ b/src-electron/db/query-session.js
@@ -681,18 +681,6 @@ async function deleteSession(db, sessionId) {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


paulr34:
> This makes little sense.... You are deleting all non-dirty sessions upon startup. How about the sessions that are CURRENTLY active, and the user is editing?
> 
> After the user clicks "Save" the session is no longer dirty. And the UI is still up.
> 
> But then a second instance of zap kicks in (because of generation) and simply wipes that session out?
> 
> You should NEVER delete an active session whether it's dirty or not.
> 
> Even so, I would never delete a session that is not more than a week or or something.
> 
> Remember: you can "restore" old sessions through that intermediate pre-session page....

huh? how is a user editing a ZAP instance prior to startup?

tecimovic (approved):

PR #1318, opened 2024-05-06
Fix generation issues with respect to template category

brdandu:
- Add more device type information in selectAllEndpointTypes query such as category and package id
- Update user_endpoints query to consider categories such that templates are only generated for their corresponding queries
- Update helpers to use selectAllEndpointTypes instead of exportEndpointTypes to only include endpoints based on the correct category. If category is not specified then generate for everything.
- Fix ensureTemplatePackageId to also check for this.global.genTemplatePackageId

### DIFF
--- a/src-electron/generator/template-util.js
+++ b/src-electron/generator/template-util.js
@@ -206,6 +206,8 @@ async function ensureZclPackageIds(context) {
 async function ensureTemplatePackageId(context) {
   if ('templatePackageId' in context.global) {
     return context.global.templatePackageId
+  } else if ('genTemplatePackageId' in context.global) {
+    return context.global.genTemplatePackageId
   } else {
     let pkgs = await queryPackage.getSessionPackagesByType(
       context.global.db,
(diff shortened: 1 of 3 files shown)


paulr34 (approved):

src-electron/generator/template-util.js
```diff
@@ -206,6 +206,8 @@ async function ensureZclPackageIds(context) {
 async function ensureTemplatePackageId(context) {
   if ('templatePackageId' in context.global) {
     return context.global.templatePackageId
+  } else if ('genTemplatePackageId' in context.global) {
```
tecimovic, line 209: I'm confused by this. Why are we returning one or the other?

tecimovic (approved):
There are few comments in the code that should be addressed, but if someone is being blocked, feel free to merge and then address later.

src-electron/generator/template-util.js
brdandu: I am not seeing templatePackageId under context.global. My guess is this somehow got changed to genTemplatePackageId somewhere down the line. So I thought keep `templatePackageId` for backwards compatibility just in case and capture the missing `genTemplatePackageId` here.

PR #1327, opened 2024-05-22
fix zigbee regen CI

paulr34:


brdandu (approved):
Approving with another issue to be created to revert this once the GitHub issue is fixed. Seems like a Github release bug.

PR #1330, opened 2024-05-23
Temperature type

paulr34:
Phase 1: user defined types

### DIFF
--- a/src-electron/generator/helper-endpointconfig.js
+++ b/src-electron/generator/helper-endpointconfig.js
@@ -651,7 +651,7 @@ function determineAttributeDefaultValue(
   typeSize,
   isNullable
 ) {
-  if (specifiedDefault !== null || !isNullable) {
+  if (specifiedDefault && (specifiedDefault !== null || !isNullable)) {
     return specifiedDefault
   }
 
@@ -659,7 +659,6 @@ function determineAttributeDefaultValue(
     // Handled elsewhere.
     return null
   }
-
   if (types.isSignedInteger(type)) {
     return '0x80' + '00'.repeat(typeSize - 1)
   }
--- a/test/endpoint-config.test.js
+++ b/test/endpoint-config.test.js
@@ -216,7 +216,7 @@ test(
       '#define FIXED_PROFILE_IDS { 0x0107, 0x0104, 0x0104 }'
     )
     expect(epc).toContain('#define FIXED_ENDPOINT_TYPES { 0, 1, 2 }')
-    expect(epc).toContain('#define GENERATED_DEFAULTS_COUNT (12)')
+    expect(epc).toContain('#define GENERATED_DEFAULTS_COUNT (13)')
     expect(epc).toContain(
       `17, 'T', 'e', 's', 't', ' ', 'm', 'a', 'n', 'u', 'f', 'a', 'c', 't', 'u', 'r', 'e', 'r',`
     )
--- a/test/endpoint-config.test.js
+++ b/test/endpoint-config.test.js
@@ -216,7 +216,7 @@ test(
       '#define FIXED_PROFILE_IDS { 0x0107, 0x0104, 0x0104 }'
     )
     expect(epc).toContain('#define FIXED_ENDPOINT_TYPES { 0, 1, 2 }')
-    expect(epc).toContain('#define GENERATED_DEFAULTS_COUNT (13)')
+    expect(epc).toContain('#define GENERATED_DEFAULTS_COUNT (19)')
     expect(epc).toContain(
       `17, 'T', 'e', 's', 't', ' ', 'm', 'a', 'n', 'u', 'f', 'a', 'c', 't', 'u', 'r', 'e', 'r',`
     )
--- a/src-electron/util/types.js
+++ b/src-electron/util/types.js
@@ -286,6 +286,7 @@ function isSignedInteger(type) {
     case 'int48s':
     case 'int56s':
     case 'int64s':
+    case 'temperature':
       return true
     default:
       return false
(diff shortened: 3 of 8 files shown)


src-electron/generator/helper-endpointconfig.js
```diff
@@ -651,15 +651,14 @@ function determineAttributeDefaultValue(
   typeSize,
   isNullable
 ) {
-  if (specifiedDefault !== null || !isNullable) {
+  if (specifiedDefault && (specifiedDefault !== null || !isNullable)) {
```
brdandu, line 654: is `specifiedDefault !== null ` and `!isNullable` different ?

brdandu: I am confused about why you need `specifiedDefault !== null`

paulr34: not totally sure if they are always the same

paulr34: from looking at endpoint-config.c, 

`  { 0x00000000, ZAP_TYPE(TEMPERATURE), 2, ZAP_ATTRIBUTE_MASK(NULLABLE), ZAP_SIMPLE_DEFAULT(0x8000) }, /* LocalTemperature */  \`

they are different entities

brdandu: Is `specifiedDefault` not handling `specifiedDefault !== null` ?

paulr34: specifiedDefault was undefined and not null which caused the bug

paulr34: we probably don't need `specifiedDefault !== null'

paulr34: ill do a phase 2 next week

test/endpoint-config.test.js
```diff
@@ -216,7 +216,7 @@ test(
       '#define FIXED_PROFILE_IDS { 0x0107, 0x0104, 0x0104 }'
     )
     expect(epc).toContain('#define FIXED_ENDPOINT_TYPES { 0, 1, 2 }')
-    expect(epc).toContain('#define GENERATED_DEFAULTS_COUNT (12)')
+    expect(epc).toContain('#define GENERATED_DEFAULTS_COUNT (19)')
```
paulr34, line 219: @tecimovic @brdandu this is interesting

src-electron/generator/helper-endpointconfig.js
brdandu: ok the worrying part here is `specifiedDefault` can be a bit off for attributes which are boolean(false) or number(0) or String(empty) . Do not add  `specifiedDefault` to this condition unless sure what is the issue without it and how this condition solved it.

test/endpoint-config.test.js
brdandu: That does not make sense. This has to be related to the if condition change.

paulr34: yeah it definitely is, but the condition was completely broken before

src-electron/generator/helper-endpointconfig.js
paulr34: Im confused, I didn't add anything except a check to make sure it is defined

test/endpoint-config.test.js
tecimovic: Hm....

tecimovic (approved):
You can push this temporarily, but we need to review.

If it's releasing a bunch of folks, we should go ahead to commit, but it's suspicious.

src-electron/generator/helper-endpointconfig.js
brdandu: Example use case which can run into an issue: There is a boolean attribute and has a default value of false
in your condition if (specifiedDefault && (specifiedDefault !== null || !isNullable)) { you will not return false but return
 // Assume everything else is an unsigned integer.
  return '0x' + 'FF'.repeat(typeSize)
  
  Log a bug to revert this and thoroughly check the issue at hand

test/endpoint-config.test.js
brdandu: Log a bug to thoroughly investigate and validate this change

src-electron/generator/helper-endpointconfig.js
bzbarsky-apple, line 654: I don't understand this code.  If `specifiedDefault` is truthy, then certainly `specifiedDefault !== null` is true, and then it doesn't matter what's going on with `isNullable`, so why do we have `isNullable` at all?

paulr34: Sometimes "specifiedDefault" is not truthy AND not null. Sometimes it's in an empty state.

paulr34: For example, when it's in this "empty" state it will fail tests checking if it's "undefined" or "null" because it's something else

bzbarsky-apple: > Sometimes "specifiedDefault" is not truthy AND not null. Sometimes it's in an empty state.

Yes, sure.  Lots of falsy things in JS are not null: `false`, `0`, `""`, `undefined`.

My poiunt is, this condition as written is completely identical to ` if (specifiedDefault) {`, so why is it not written that way?

And if the input is in fact 0, do we really want to output `0xFF` and whatnot?

paulr34: @bzbarsky-apple thanks for your review. I reverted the commit and will make a new PR.

paulr34: @bzbarsky-apple when it comes to this code, I was under the impression that the javascript representation of specifiedDefault being 0 would be a string which would return true 

That being said, I reverted this because if there is an empty string then it will return false which as discussed on slack is a bug

PR #1333, opened 2024-05-30
Custom type phase 1

paulr34:


src-electron/db/query-atomic.js
```diff
@@ -82,6 +82,27 @@ async function selectAtomicById(db, id) {
...
 
+/**
+ * Checks if a type by a given name is signed.
+ *
+ * @param {object} db - The database connection object.
+ * @param {string} name - The name of the type.
+ * @param {Array} sessionPackages - An array of session packages.
+ * @returns {Promise<boolean>} - A promise that resolves to true if the type is signed, false otherwise.
+ */
+async function isTypeSignedByNameAndPackage(db, name, sessionPackages) {
```
brdandu, line 93: This just checks the atomic table not the entire set of data types that are not in the atomic table.

src-electron/db/query-atomic.js
```diff
@@ -82,6 +82,27 @@ async function selectAtomicById(db, id) {
...
+  const sessionPackage = sessionPackages[0].packageRef
+  const rows = await dbApi.dbAll(
```
brdandu, line 95: use dbGet so you don't need to get 0th index of array below.

src-electron/db/query-atomic.js
paulr34: yeah, the DATA_TYPE table does not have "IS_SIGNED" which makes sense because non-atomic types are not signed or unsigned by nature

for example, a struct (non-atomic) and an integer (atomic)

src-electron/db/query-atomic.js
paulr34: good idea, thanks

src-electron/db/query-atomic.js
brdandu: Is number table the only table which has a sign? If you use the function name I mentioned before you will not have to worry about a specific table. It will account all data types and it will be scalable to future changes as well.
for eg: some day or may be even today number table could have entries which the atomic table does not have. Then the sign won't be captured properly

Also you do not need to write this query. see query-atomic#selectAtomicType. It does the same thing. Re-use that.

paulr34: You mean the ATOMIC table? Yes it's the only table with IS_SIGNED

brdandu: Approving it given that the the code redundant query in this function is switched to using selectAtomicType which is cached and does the same thing. Get rid of `await dbApi.dbGet(
     db,
     `SELECT IS_SIGNED FROM ATOMIC WHERE NAME = ? AND PACKAGE_REF = ?`,
     [name, sessionPackage]
   )`

brdandu (approved):

src-electron/db/query-atomic.js
brdandu: Also Atomic table is not the only table that is signed. See schema for number table which also has a sign.

brdandu (approved):

src-electron/db/query-atomic.js
paulr34: Okay thank you I'll check on that before merging

PR #1335, opened 2024-06-05
Fix attribute validation issues

ben1hop:


src-electron/validation/validation.js
```diff
@@ -291,26 +286,24 @@ async function getIntegerFromAttribute(attribute, typeSize, isSigned) {
...
-  attribType
-) {
+async function getIntegerAttributeSize(db, zapSessionId, attribType) {
   let packageIds = await queryPackage.getSessionZclPackageIds(db, zapSessionId)
-  let attribData = await queryZcl.selectNumberByNameAndClusterId(
+  const attribData = await types.getSignAndSizeOfZclType(
     db,
     attribType,
-    clusterRef,
     packageIds
```
brdandu, line 297: Hoping we don't have issues if 2 clusters define the same type name. 2 clusters can define a data type with the same name but different types and sizes.

brdandu: Wondering if keeping queryZcl.selectNumberByNameAndClusterId is safer but I will leave it to you.

brdandu (approved):

src-electron/validation/validation.js
ben1hop: it does not work on some types , for example bitmaps

paulr34 (approved):

PR #1336, opened 2024-06-10
show NULL when nullable and default is null

paulr34:
when default value is null and the attribute isNullable then the defaultValue is NULL not empty

### DIFF
--- a/src/components/ZclAttributeManager.vue
+++ b/src/components/ZclAttributeManager.vue
@@ -185,7 +185,12 @@ limitations under the License.
                   : ''
               "
               :disable="isDisabledDefault(props.row.id, selectedCluster.id)"
-              :model-value="defaultValueCheck(props.row.id, selectedCluster.id)"
+              :model-value="
+                props.row.isNullable &&
+                defaultValueCheck(props.row.id, selectedCluster.id) === null
+                  ? 'NULL'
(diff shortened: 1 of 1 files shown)


src/components/ZclAttributeManager.vue
```diff
@@ -185,7 +185,12 @@ limitations under the License.
                   : ''
               "
               :disable="isDisabledDefault(props.row.id, selectedCluster.id)"
-              :model-value="defaultValueCheck(props.row.id, selectedCluster.id)"
+              :model-value="
+                props.row.isNullable &&
+                defaultValueCheck(props.row.id, selectedCluster.id) === null
+                  ? 'NULL'
```
tecimovic, line 191: I am not happy that we hardcode "NULL" here. This is mixing up the C concept of "NULL" with a concept of "nil" from the spec.

What if we're not generating for C, so we don't have "NULL"?

tecimovic (approved):

src/components/ZclAttributeManager.vue
```diff
@@ -211,11 +216,6 @@ limitations under the License.
                   color="secondary"
                   label="Null"
                   size="sm"
-                  :disabled="
```
brdandu, line 214: Why is this change needed?

PR #1345, opened 2024-06-24
Matter generation performance improvements

tecimovic:
0. Some dependencies cleanup.
1. Matter performance enhancements (mostly from package caching).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2329,9 +2329,9 @@
       }
     },
     "node_modules/@cypress/code-coverage": {
-      "version": "3.12.9",
-      "resolved": "https://registry.npmjs.org/@cypress/code-coverage/-/code-coverage-3.12.9.tgz",
-      "integrity": "sha512-MIVISl2WOYzV0DgRZXs9wxYRBMM5eN2n/p4PN0wPJCATllfv8nqqyMC1TGWBrf8Di+QMikUjbSZ20w/5+9w4Lg==",
+      "version": "3.12.39",
+      "resolved": "https://registry.npmjs.org/@cypress/code-coverage/-/code-coverage-3.12.39.tgz",
+      "integrity": "sha512-ja7I/GRmkSAW9e3O7pideWcNUEHao0WT6sRyXQEURoxkJUASJssJ7Kb/bd3eMYmkUCiD5CRFqWR5BGF4mWVaUw==",
       "dev": true,
       "dependencies": {
         "@cypress/webpack-preprocessor": "^6.0.0",
@@ -2339,8 +2339,8 @@
         "dayjs": "1.11.10",
         "debug": "4.3.4",
         "execa": "4.1.0",
-        "globby": "11.0.4",
-        "istanbul-lib-coverage": "3.0.0",
+        "globby": "11.1.0",
+        "istanbul-lib-coverage": "^3.0.0",
         "js-yaml": "4.1.0",
         "nyc": "15.1.0"
       },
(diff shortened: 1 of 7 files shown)


tecimovic:
25% of time with Matter regen is spent in querying the DB for  full package info out of packageId, to obtain category. Given that we already have that info, but then choose to ONLY cache the ID, that's a bit dumb.

So I'm changing the logic so that the entire package info is cached, and then you can get ID or category or  whatever else you need from that, instead of having to run silly amount of "getPackageByPackageId()" queries.

paulr34 (approved):

src-electron/db/query-package.js
```diff
@@ -26,7 +26,7 @@ const dbEnum = require('../../src-shared/db-enum.js')
 const querySession = require('./query-session')
```
brdandu, line 26: Did anything change in this file? I might sendup having a big conflict on this from my PR.

src-electron/generator/template-util.js
```diff
@@ -197,6 +204,20 @@ async function ensureZclPackageIds(context) {
   }
 }
 
+async function ensureTemplatePackageCategory(context) {
```
brdandu, line 207: Missed some comments for the new function

package-lock.json
brdandu, line 1: Why did the package-lock.json file change?

src-electron/generator/template-util.js
```diff
@@ -142,23 +142,30 @@ async function ensureZclPackageId(context) {
...
 async function ensureZclPackageIds(context) {
-  // Get category of templates.json
-  let pkg = await queryPackage.getPackageByPackageId(
-    context.global.db,
-    context.global.genTemplatePackageId
-  )
-  let packageCategory = pkg ? pkg.category : null
+  let packageCategory = null
+  if (context.global.genTemplatePackage != null) {
+    packageCategory = context.global.genTemplatePackage.category
```
brdandu, line 147: So we do not have genTemplatePackageInfo in the context object already?
Wondering if this info already existed as well?

src-electron/db/query-package.js
tecimovic: I think it's just a default formatter. We need to sync up with the formatter, so we all use the same. I didn't change any queries.

package-lock.json
tecimovic: Because along the way I ran `npm update` and `npm install`.

src-electron/generator/template-util.js
tecimovic: I'll add them....

src-electron/generator/template-util.js
tecimovic: We were just storing the IDs. I made the whole package object be stored in the cache. That help in cases where you need a category, since you have it already there, instead of having to look it up from the DB. That alone solved 25% of performance issues.

brdandu (approved):

src-electron/generator/template-util.js
brdandu: Should we be doing `packageCategory = await ensureTemplatePackageCategory(context)` ?

src-electron/generator/template-util.js
```diff
@@ -229,11 +256,10 @@ async function ensureTemplatePackageId(context) {
...
  */
 async function ensureEndpointTypeIds(context) {
-  let pkg = await queryPackage.getPackageByPackageId(
-    context.global.db,
-    context.global.genTemplatePackageId
-  )
-  let packageCategory = pkg.category
+  let packageCategory = null
+  if (context.global.genTemplatePackage != null) {
+    packageCategory = context.global.genTemplatePackage.category
```
brdandu, line 261: Should we be doing `packageCategory = await ensureTemplatePackageCategory(context)` ?

src-electron/generator/template-util.js
tecimovic: Yep.

brdandu: I meant instead of doing `packageCategory = context.global.genTemplatePackage.category`. Does that code need to be updated to `packageCategory = await ensureTemplatePackageCategory(context)`

PR #1347, opened 2024-06-24
Fixing the top level zcl file registration when the gsdk is updated from one branch to another

brdandu:
- There is a check for whether a zcl.json or any of its child xml packages being updated. If updated, then the top level package is added again along with its child packages with the is_in_sync column as true for the top level package. However the old top level package's is_in_sync is turned to false to point out an outdated package.
- All the query-pacakge queries have been updated to use the is_in_sync = 1 package such that the old one is not picked up by mistake.
- Updated the zap schema, added is_in_sync to package table.
- JIRA: ZAPP-1079


paulr34 (approved):

src-electron/db/query-package.js
```diff
@@ -339,7 +357,22 @@ async function updatePathCrc(db, path, crc, parentId) {
...
+ * @param {*} db
+ * @param {*} packageRef
+ * @param {*} isInSync
+ * @returns Promise of an update.
+ */
+async function updatePackageIsInSync(db, packageRef, isInSync) {
+  return dbApi.dbUpdate(
+    db,
+    'UPDATE PACKAGE SET IS_IN_SYNC = ? WHERE PACKAGE_ID = ?',
+    [isInSync, packageRef]
```
tecimovic, line 375: Should we not use `dbApi.toDbBool()` here?
I am always cautious about taking arguments from JS and turning them into DB boolean values (0/1).

tecimovic (commented):
I thought there would have to be a changed to `db-mapping.js` somewhere? TO add to the `package` map the new in-sync column?

tecimovic:
> I thought there would have to be a changed to `db-mapping.js` somewhere? TO add to the `package` map the new in-sync column?

Oh, right... Maybe you don't even need to query for that explicitely. You just used it implicitely in the queries...

tecimovic (approved):

src-electron/db/query-package.js
brdandu: Good point.

PR #1348, opened 2024-06-25
Generation performance enhancements

tecimovic:


package.json
brdandu, line 1: Can we add documentation to our GitHub developer documentation on how to do profiling for generation?

src-electron/generator/generation-engine.js
```diff
@@ -752,7 +764,9 @@ async function generateSingleTemplate(
     disableDeprecationWarnings: false,
   }
 ) {
-  env.logInfo(`Start generating from template: ${genTemplateJsonPackage?.path}`)
+  let genStart = process.hrtime.bigint()
+  //console.log(`Start generating from template: ${singleTemplatePkg?.path}`)
```
brdandu, line 768: Should we keep the console.log? I think it is generally nice to see this during all our generations instead of seeing it in some hidden form unless logInfo does console.log.

src-electron/generator/generation-engine.js
```diff
@@ -774,8 +788,12 @@ async function generateSingleTemplate(
       genResult.stats[result.key] = result.stats
     }
     genResult.partial = true
+    let nsDuration = process.hrtime.bigint() - genStart
+    //console.log(`Finish generating from template: ${singleTemplatePkg?.path}: ${util.duration(nsDuration)}`)
     env.logInfo(
-      `Finish generating from template: ${genTemplateJsonPackage?.path}`
+      `Finish generating from template: ${
+        singleTemplatePkg?.path
+      }: ${util.duration(nsDuration)}`
```
brdandu, line 796: Also, do we continue to get the cumulative generation time like we did before?

test/gen-matter-3-1.test.js
brdandu, line 1: Confused what really happened here between these test.js files

paulr34 (approved):

package.json
tecimovic: Yeah, I will do that....

src-electron/generator/generation-engine.js
tecimovic: Console.log prints out the totals. I think it would be too much clutter if we print out these individual templates.

test/gen-matter-3-1.test.js
tecimovic: At some point the numbers shifted.
We have test data: matter1/2/3/4.
Then we had tests: matter1/2/3/4.

However Matter test 3 started using data 4. So when I added data 5, I would create test 4 and that felt stupid.
So I moved matter4 test to be named matter-3-1, since it uses test data 3 and then a new test data 4 can be used by matter test 4, so that the numbers align.

PR #1350, opened 2024-06-28
feat: add device type's icons to cards

tbrkollar:


src/components/ZclEndpointCard.vue
```diff
@@ -289,6 +300,15 @@ export default {
     }
   },
   methods: {
+    getDeviceCategory(packageRef) {
+      let category = ''
+      this.$store.state.zap.selectedZapConfig.zclProperties.forEach((item) => {
+        if (item.id === packageRef) {
+          category = item.category
```
brdandu, line 307: `return category` once it has been populated.

src/util/common-mixin.js
```diff
@@ -266,17 +266,30 @@ export default {
         (x) => this.sdkExtClusterCode(x) === clusterRoleName
       )
     },
+    createLogoSrc(isTiny, category, isSelected = false) {
+      return (
+        (isTiny ? '/logo/tiny/' : '/logo/') +
+        category +
```
brdandu, line 272: What happens when category is empty string (''). Some of our very old GSDKs did not have category.

src/util/common-mixin.js
```diff
@@ -266,17 +266,30 @@ export default {
...
-                '.svg'
+              this.createLogoSrc(
+                this.$store.state.zap.isMultiConfig,
+                selectedZapConfig[i].category
+              )
+              // (this.$store.state.zap.isMultiConfig ? '/logo/tiny/' : '/logo/') +
+              //   selectedZapConfig[i].category +
+              //   '_logo' +
+              //   (this.$q.dark.isActive ? '_white' : '') +
+              //   '.svg'
```
brdandu, line 292: Get rid of comments

src/components/ZclEndpointCard.vue
```diff
@@ -289,6 +300,15 @@ export default {
     }
   },
   methods: {
+    getDeviceCategory(packageRef) {
```
brdandu, line 303: Add comments to function

src/util/common-mixin.js
```diff
@@ -266,17 +266,30 @@ export default {
         (x) => this.sdkExtClusterCode(x) === clusterRoleName
       )
     },
+    createLogoSrc(isTiny, category, isSelected = false) {
```
brdandu, line 269: Add comments to function.

brdandu (changes requested):

src/components/ZclEndpointCard.vue
brdandu: Keep the comment format as below for this function too
`/**
      *  Specifies the path to the logo image
      *
      * @param {*} isTiny -
      *                   if value "true", it's used as a tiny logo
      *                   if value "false", it's used as a normal logo
      * @param {*} category -
      *                   if value "zigbee", it's used as a zigbee logo
      *                   if value "matter", it's used as a matter logo
      *                   if value "multiprotocol", it's used as a multiprotocol logo
      * @param {*} isSelected -
      *                   if value "true", it's used as a dark / selected version
      *                   if value "false", it's used as a light version
      * @returns Returns a string value for the src image property
      */`

src/components/ZclEndpointCard.vue
```diff
@@ -289,6 +300,19 @@ export default {
...
   },
   methods: {
+    // Calculate readable category string
+    getDeviceCategory(packageRef) {
+      let category = ''
+      this.$store.state.zap.selectedZapConfig.zclProperties.forEach((item) => {
+        if (item.id === packageRef) {
+          if (item.category) {
+            category = item.category
+            return
```
brdandu, line 310: Just realized the returning in a forEach does not really end the loop but it continues going through the loop regardless.

Can you switch `this.$store.state.zap.selectedZapConfig.zclProperties.forEach((item) => {
         if (item.id === packageRef) {
           if (item.category) {
             category = item.category
             return
           }
         }
       })`
       
to

`
let zclProperty = this.$store.state.zap.selectedZapConfig.zclProperties.find((item) => (item.id === packageRef && item.category))
if (zclProperty) {
category = zclProperty.category
} 
return category
`

brdandu: You can also switch to a normal for loop and return instead of using forEach

brdandu (approved):

src/components/ZclEndpointCard.vue
tbrkollar: you are right, thanks. i will fix it

PR #1351, opened 2024-07-01
Fixing the endpoint type event schema

brdandu:
- Getting rid of the endpoint_type_ref from endpoint_type_event table since the endpoint type cluster table already has this. Normalizing the schema between endpoint type event and endpoint type cluster.
- Updating the schema after changes to endpoint_type_event table
- JIRA: ZAPP-1129


tecimovic (approved):
Seems mostly ok.

Issue #1356, opened 2024-07-11
Default value handling broken for nullable strings and other situations


bzbarsky-apple:
The default value handling in `collectAttributes` in `helper-endpointconfig.js` is broken, in the sense that it generates code that does not compile.

Specifically, when generating long defaults there are the following cases:

1. String attribute with null default value.  In this case, we do either `def = ['0xFF']` or `def = ['0xFF', '0xFF']` depending on whether it's a long string.
2. Something else, in which we do `def = types.longTypeDefaultValue(stuff)`.  What this returns is:
    a. If value is null, a string ending in `", "`
    b. If value is NaN, a string that does NOT end in a comma.
    c. Otherwise a string ending in a comma (with no space after it).

We then treat `def` as a string (even though sometimes it's an array, for case 1 above).

Note that this means that there is no comma after the value in cases 1 and 2b, and those cases will then not compile.

There might be other places that are broken here too; I did not go through all the possible codepaths.  Someone needs to do that.

This is breaking people who make the mistake of marking a nullable string as null in their .zap file.

paulr34:
@bzbarsky-apple "make the mistake of marking a nullable string as null" 

do you mean this is a mistake because ideally it should not be allowed OR a mistake because ZAP causes compilation errors?

bzbarsky-apple:
> OR a mistake because ZAP causes compilation errors?

A "mistake" because it leads to a broken codegen and compilation errors.  Sorry I should have been clearer that it's not a "mistake" in that they are doing something wrong, just a "mistake" in that they are expecting ZAP to behave and it's not.

paulr34:
https://github.com/project-chip/zap/commit/69b35220ae0d521b58e651c4b56c21d3b0c1e960
unit test https://github.com/project-chip/zap/commit/4fd8efde6c3c78743c76cf5e1a6a493163d57490

paulr34:
closing this and made a new issue for auditing "collect_attributes" https://github.com/project-chip/zap/issues/1392

PR #1357, opened 2024-07-11
issue_1356

paulr34:
creating two separate flows because the default value is can be an array or a string

https://github.com/project-chip/zap/issues/1356


paulr34:
initial work for creating different code paths for different types

PR #1365, opened 2024-07-22
Resolved the intermittent packageId mismatch error in gen-test.test.js and zcl-loader-consecutive.test.js when running locally

dhchandw:
- Made separate test in `zcl-loader.test.js` for checking reloading of ZCL packages on XML file changes
- This test was previously combined with the  'test Silabs zcl data loading in memory' test
- Added `zcl-builtin/silabs/zcl-special.json` and `zcl-builtin/silabs/general-special.xml` for this particular test to use
- Introduced functions in `src-electron/util/env.js` to retrieve the above files
- This test logic was previously changing `zcl-builtin/silabs/general.xml` used by other tests causing unintentional packageId mismatch.

**Fix:** Resolved the intermittent packageId mismatch error in gen-test.test.js and zcl-loader-consecutive.test.js when running unit tests locally.


brdandu:
Make sure your commit message also mentions. The more the details in a commit message the better.

> Moved logic for checking reloading of ZCL packages on XML file changes to zcl-loader-reload.test.js
Added zcl-special.json and general-special.xml for the test above
Introduced functions in src-electron/util/env.js to retrieve the above files
Resolved the intermittent packageId mismatch error in gen-test.test.js and zcl-loader-consecutive.test.js when running unit tests locally.

Make sure to mention the jira/github issue in your commit as well for future references to look back at associated changes.

test/zcl-loader-reload.test.js
brdandu, line 1: Since you have updated the xml and json files. Can this continue to exist in the same zcl-loader.test.js

src-electron/util/env.js
```diff
@@ -31,10 +31,22 @@ export function defaultFileFormat() {
   return saveFileFormat
 }
 
+/**
+ *
+ * @returns path to zcl.json file
+ */
```
brdandu, line 37: Minor but for future references have all functions go into the following format:

> /**
 * Description:
 * Example:
 * @param {*} param1
 * @param {*} param2
 * @returns: returns ---
 */

paulr34 (approved):

PR #1366, opened 2024-07-23
Feature/modify cluster configuration

tbrkollar:


src/components/ZclClusterView.vue
```diff
@@ -171,6 +174,25 @@ export default {
...
+        return this.getDeviceCategory(
+          this.zclDeviceTypes[
+            this.endpointDeviceTypeRef[this.selectedEndpointId][0]
+          ]?.packageRef
+        )
+      },
+    },
+    enableEventsTab: {
+      get() {
+        return this.category !== dbEnum.helperCategory.zigbee
```
brdandu, line 188: Should this be =='matter' instead and also show if it is null. Ideally stacks(gsdk json files) should be enabling the UI components. If we really want to make this more scalable.

brdandu: Log a bug to drive some of these things via the gsdk. It is fine for now.

brdandu (approved):

PR #1368, opened 2024-07-24
Fix fails in gen-zigbee unit tests

ethanzhouyc:
- Fix possible fails in gen-zigbee-x.test.js unit tests due to inconsistent execution order in async functions
    
- Replace .then() with await and replace forEach with for loop in ensurePackagesAndPopulateSessionOptions to ensure proper async flow
    
- Keep using promises in other parts of the function for faster parallel execution

### DIFF
--- a/src-electron/util/util.js
+++ b/src-electron/util/util.js
@@ -213,143 +213,136 @@ async function ensurePackagesAndPopulateSessionOptions(
         sessionPartitionIndex++
       }
     } else {
-      let genTemplateJsonPromise = queryPackage
-        .getPackagesByType(db, dbEnum.packageType.genTemplatesJson)
-        .then((rows) => {
-          let packageId
-          if (
-            selectedGenTemplatePackages &&
-            selectedGenTemplatePackages.length > 0
-          ) {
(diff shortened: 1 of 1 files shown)


paulr34 (approved):

dhchandw:
Mentioning the function changed and mentioning its role in these tests might be useful.

src-electron/util/util.js
```diff
@@ -213,143 +213,136 @@ async function ensurePackagesAndPopulateSessionOptions(
...
+      let rows = await queryPackage.getPackagesByType(
+        db,
+        dbEnum.packageType.genTemplatesJson
+      )
+      let packageId
+      if (
+        selectedGenTemplatePackages &&
+        selectedGenTemplatePackages.length > 0
+      ) {
+        for (const gen of selectedGenTemplatePackages) {
```
brdandu, line 225: Minor stuff `const genPkgId` gives more readability but not a big deal.

src-electron/util/util.js
brdandu, line 1: It would be good to break this long function into smaller pieces but that can be done on a later date so do not worry about it now.

brdandu (approved):

PR #1373, opened 2024-07-25
Fix zcl_struct_items_by_struct_and_cluster_name to work for global structs.

bzbarsky-apple:


andy31415 (approved):

src-electron/generator/helper-zcl.js
```diff
@@ -323,6 +323,17 @@ async function zcl_struct_items_by_struct_and_cluster_name(
   options
 ) {
   let packageIds = await templateUtil.ensureZclPackageIds(this)
```
andy31415, line 325: should we update the documentation to explain how global clusters take effect? then cluster name has no meaning.

bzbarsky-apple: Done.

andy31415 (approved):

PR #1374, opened 2024-07-26
Fix Darwin name determination for global structs and enums.

bzbarsky-apple:
The naming pattern is a bit different from cluster-specific ones.

### DIFF
--- a/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
+++ b/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
@@ -240,6 +240,18 @@ async function asObjectiveCClass(type, cluster, options) {
   }
 
   if (isStruct) {
+    const structObj = await zclQuery.selectStructByName(
+      this.global.db,
+      type,
+      pkgIds
+    );
+    if (structObj.structClusterCount == 0) {
+      // This is a global struct.
+      return `${
+        options.hash.structTypePrefix || 'MTR'
+      }DataType${appHelper.asUpperCamelCase(type)}`;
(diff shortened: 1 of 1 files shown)


ksperling-apple (approved):

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
brdandu, line 1: Can this lead to backwards compatibility issues if someone has new zap and old gsdk?
Should these changes be tied to some classification of a new vs old gsdk?

bzbarsky-apple: It can't, because in old SDK there are no global structs and clusterName is never 'Globals'.

brdandu (approved):

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -339,6 +353,12 @@ function commandHasRequiredField(command) {
  * "Enum" bits on the enum names while we're here.
  */
 function objCEnumName(clusterName, enumLabel, options) {
+  if (clusterName == 'Globals') {
```
brdandu, line 356: We should probably start defining these loose hanging strings in db-enum for better maintenance in the future.

bzbarsky-apple: Possible.  What should _really_ happen is that this helper should not be in ZAP at all, but in the SDK.  ;)

PR #1376, opened 2024-07-26
Add support for global types in the Python Matter codegen.

bzbarsky-apple:


brdandu (approved):
Don't want to slow you down but unit tests for these global helpers would be good for the future.

bzbarsky-apple:
No argument; I'll see if I can get the folks maintaining the Python stuff in general to add some tests, and will think about some for the Darwin bits.

Or better yet, move some of these helpers out of ZAP and back into the Matter SDK.  ;)

jmartinez-silabs (approved):

Issue #1375, opened 2024-07-26
ZAP GUI - loading large data


paulr34:
When using the ZAP GUI it is very frustrating having the application glitch while I am scrolling through Attributes

Steps to reproduce:
1. create an application with a lot of attributes (thermostat for example)
2. within thermostat cluster settings scroll through the attributes and notice sometimes disabled attributes look enabled and sometimes the screen is completely empty

note: it does seem to fix itself after 

Another way to reproduce this is by scrolling through device types like in the picture below
<img width="436" alt="Screenshot 2024-07-26 at 11 16 46 AM" src="https://github.com/user-attachments/assets/1ac505a4-f28f-4c7a-8e0f-a9e0d95feee6">

PART 2:

we need a new ZAP GUI feature which handles large data sets

ethanzhouyc:
It is caused by this [virtual-scroll](https://github.com/project-chip/zap/blob/b0d702c32b2f0683b305d3007a199e803b09c5d2/src/components/ZclAttributeManager.vue#L31) attribute in the <q-table>. It helps improve performance by only displaying a small amount of the entire list and updates the list as users scroll up and down. So adding this attribute will bring side effects: 
1. The toggle's abnormal enable/disable behavior; 
2. When scrolling too fast, rows disappear because new rows are still being loaded.

Maybe we should consider deleting this attribute as it affects user experience.

paulr34:
Sounds promising. Can you please experiment and let me know?

paulr34:
It turns out this problem requires a new feature to fix it. Changing from bug to "enhancement".

ethanzhouyc:
For current solutions, using virtual-scroll can lead to GUI issues, and preloading all data for a large cluster introduces significant latency. We need to develop a new feature that efficiently loads of large datasets without compromising on performance.

PR #1377, opened 2024-07-26
Fix glitchy scrolling in attributes by removing virtual-scroll

ethanzhouyc:
- Solved issue #1375

- Removed virtual-scroll from q-table component in the ZAP GUI


paulr34 (approved):

paulr34 (changes requested):
let's turn this PR into a feature rather than a bug fix so the entire problem is fixed...I have updated this issue https://github.com/project-chip/zap/issues/1375

paulr34 (approved):

PR #1381, opened 2024-07-31
fixing issue_1356

paulr34:
when a nullable string is null, the javascript code should be a string and not an array so the generated code compiles


tecimovic (approved):
Not really sure why we had arrays there in the first place. Purely technically, arrays would be nice, but I see that the `longTypeDefaultValue` doesn't return array either, so I guess this is at least makes it consistently un-nice....

PR #1383, opened 2024-07-31
feat: modify zclclusterview popup to page

tbrkollar:


src/components/ZclClusterView.vue
```diff
@@ -198,12 +207,7 @@ export default {
       this.tab = val
     },
   },
-  mounted() {
```
brdandu, line 201: Why is this removed?

src/components/ZclClusterView.vue
```diff
@@ -222,20 +226,10 @@ export default {
     setIndividualClusterFilterString(filterString) {
       this.$store.dispatch('zap/setIndividualClusterFilterString', filterString)
     },
-    onResize(size) {
```
brdandu, line 225: Why is this removed?

src/components/ZclClusterView.vue
```diff
@@ -222,20 +226,10 @@ export default {
...
-    },
-    calculateTableSize() {
-      this.tableHeight = '30px'
-      this.tableWidth = '500px'
-    },
   },
   data() {
     return {
       tab: 'attributes',
-      tableHeight: '30px',
```
brdandu, line 237: What does this change?

src/components/ZclCreateModifyEndpoint.vue
```diff
@@ -58,7 +58,7 @@ limitations under the License.
             class="col v-step-2"
             use-input
             :multiple="enableMultipleDevice"
-            :use-chips="enableMultipleDevice"
+            use-chips
```
brdandu, line 61: What does this do?

src/layouts/ZclLayout.vue
brdandu, line 1: Were the 3 .vue files that are deleted not used at all?

test/ui.test.js
```diff
@@ -185,18 +183,7 @@ describe('Component mounting test', () => {
...
-    () => {
-      const wrapper = shallowMount(ZclClusterLayout, {
-        global: {
-          plugins: [ZapStore()],
-        },
-      })
-      expect(wrapper.html().length).toBeGreaterThan(90)
-    },
-    timeout.short()
-  )
```
brdandu, line 199: Why are these 2 tests removed?

src/layouts/ZclLayout.vue
tbrkollar: we are not using there anymore

src/components/ZclCreateModifyEndpoint.vue
tbrkollar: in standalone zigbee mode, we could not use the backspace / remove btn on the new card endpoint card. But if we modify this, it is working well. (I had a meeting with @paulr34, @dhchandw, @ethanzhouyc and this was one of the topics.)

src/components/ZclClusterView.vue
tbrkollar: it is not needed anymore. It was necessary for the table's height in popup.

src/components/ZclClusterView.vue
tbrkollar: similar to the previous comment

brdandu (approved):

PR #1384, opened 2024-07-31
add unit test for issue1356

paulr34:
unit test for types.nullStringDefaultValue(a.type)


tecimovic (approved):
Nice. A small refactoring and a small unit test. Makes things better, one line of the code at a time!

PR #1386, opened 2024-08-01
Updating the schema to not trigger device type compliance warnings for optional clusters

brdandu:
- Updating the loader to not trigger device type compliance warnings for optional clusters along with its corresponding attributes
- Adding selectEndpointTypeClusterFromEndpointIdentifierAndAttributeRef to retrieve endpoint type cluster information for the above
- Adding tests for warnings associated to optional clusters and its attributes in a device type.
- Github: ZAP#1267


tecimovic (approved):
Looks good in general, but I did get a little bit lost with SQL. I'll trust you on that.../.

Issue #1387, opened 2024-08-01
Custom Cluster Remains in Generated Files After Custom XML Removal


dhchandw:
In ZAP, when a custom XML with a custom cluster is loaded into a **Matter session**, and an endpoint is configured to have this custom cluster toggled on, the generated files incorrectly retain the custom cluster even after the XML is deleted from the session.

**Steps to Reproduce:**
- Open a ZAP matter session.
- Load a custom XML that includes a custom cluster.
- Create an endpoint and configure it to have the custom cluster toggled on.
- Delete the custom XML from the session.
- Generate the files.

**Expected Behavior:**
The generated files should not include the custom cluster after the XML is deleted from the session.

**Actual Behavior:**
The generated files incorrectly still include the custom cluster, even though the XML has been removed. This is only noticed in the dynamic files (endpoint-config.c, endpoints.out, ...)

PR #1389, opened 2024-08-05
Adding tests for Matter custom xml support

dhchandw:
Added tests for **Matter custom xml support** for the following use cases:
- loads a file with a custom cluster and an extension of a cluster and checks it
- loads a second file with a further extension of the extended cluster and checks it
- tests generation with custom xml
- loading bad custom xml -> throw warning
- Add 2 custom files with manufacturer name/code and attribute code collisions -> throw error.
    - This should be an SQL error, it is currently broken and should be fixed in PR #1352 is merged.
- test generation after deleting the custom xml
    - This doesn't work as expected - <https://github.com/project-chip/zap/issues/1387>
- Add custom xml, make changes, delete, add again -> make sure that generated content has new edits
     - Effected by <https://github.com/project-chip/zap/issues/1387> and <https://github.com/project-chip/zap/issues/1391>
- Loading zap file with custom xml that does not exist - throw error
- Loading zap file with custom xml and generating

Once PR #1352 is merged and #1387 and #1391 are resolved - tests should be updated appropriately.


paulr34 (approved):

brdandu (approved):
- One edge case worth consideration is if the custom xml path is valid but does not exist in the database. what happens? We have users sometime deleting the .zap repo and then opening their .zap files. Ideally that behavior should not break ZAP. Worth testing that as well. Approving the PR however if you want to merge and add that later.
- Also link this PR to your issue for easy reference in the future.
- You have pretty good description on the PR. Add that to your git commit message it is useful information. Also add the git/jira issue associated with your PR to the git commit eg(JIRA: ZAPP-123 or Github: ZAP#123)

Issue #1397, opened 2024-08-09
List type attributes should be forced external


paulr34:
List type attributes should be locked to External 

Example, the list attributes should be be locked as External like Cluster Revision is
<img width="1279" alt="Screenshot 2024-08-09 at 3 44 02 PM" src="https://github.com/user-attachments/assets/4a543ccf-9e12-487b-9616-8375cb17f17e">

PR #1398, opened 2024-08-09
Issue 1397

paulr34:
Fixes https://github.com/project-chip/zap/issues/1397

This pull request retrieves the cluster/attribute pairs for all attributeAccessInterface attributes, including those of the list type.

Background: The ATTRIBUTE table designates CLUSTER_REF as null for global attributes.

To incorporate cluster/attribute pairs that are global attributes, the attributeAccessInterface list from zcl.json needs to be imported into an alternative table, PACKAGE_OPTION.

Given that PACKAGE_OPTION exclusively contains pairs enumerated in zcl.json and does not encompass all attributeAccessInterface pairings, it becomes essential to query both the ATTRIBUTE and PACKAGE_OPTION tables.

### DIFF
--- a/src-electron/db/query-config.js
+++ b/src-electron/db/query-config.js
@@ -235,7 +235,10 @@ async function insertOrUpdateAttributeState(
     )
     staticAttribute.defaultValue = featureMapDefaultValue
   }
-  let forcedExternal = await queryUpgrade.getForcedExternalStorage(db)
+  let forcedExternal = await queryUpgrade.getForcedExternalStorage(
+    db,
+    staticAttribute.packageRef
+  )
   staticAttribute.storagePolicy =
     await queryUpgrade.computeStoragePolicyNewConfig(
       db,
--- a/src-electron/db/query-impexp.js
+++ b/src-electron/db/query-impexp.js
@@ -737,7 +737,10 @@ WHERE
     attribute.reportable = false
   }
   if (attributeId) {
-    forcedExternal = await queryUpgrade.getForcedExternalStorage(db)
+    forcedExternal = await queryUpgrade.getForcedExternalStorage(
+      db,
+      packageIds[0]
+    )
     storagePolicy = await queryUpgrade.computeStorageImport(
       db,
       cluster.name,
--- a/src-electron/db/query-impexp.js
+++ b/src-electron/db/query-impexp.js
@@ -737,10 +737,7 @@ WHERE
     attribute.reportable = false
   }
   if (attributeId) {
-    forcedExternal = await queryUpgrade.getForcedExternalStorage(
-      db,
-      packageIds[0]
-    )
+    forcedExternal = await queryUpgrade.getForcedExternalStorage(db, packageIds)
     storagePolicy = await queryUpgrade.computeStorageImport(
       db,
       cluster.name,
--- a/src-electron/db/query-package.js
+++ b/src-electron/db/query-package.js
@@ -706,19 +706,14 @@ async function getAllPackages(db) {
  */
 async function getAttributeAccessInterface(db, code, packageIds) {
   try {
-    // Ensure packageIds is always an array
-    if (!Array.isArray(packageIds)) {
-      packageIds = [packageIds]
-    }
-
     let packageRefCondition = `po.PACKAGE_REF = ?`
     let attributePackageRefCondition = `a.PACKAGE_REF = ?`
     let queryParams = [code, ...packageIds, code, ...packageIds]
 
-    // Since packageIds is now always an array, adjust the query and parameters accordingly
-    const placeholders = packageIds.map(() => '?').join(', ')
-    packageRefCondition = `po.PACKAGE_REF IN (${placeholders})`
-    attributePackageRefCondition = `a.PACKAGE_REF IN (${placeholders})`
+    // Use dbApi.toInClause to automatically create the placeholders for the IN clause
+    const inClause = dbApi.toInClause(packageIds)
+    packageRefCondition = `po.PACKAGE_REF IN (${inClause})`
+    attributePackageRefCondition = `a.PACKAGE_REF IN (${inClause})`
 
     const extendedQuery = `
       SELECT
--- a/src-electron/db/query-package.js
+++ b/src-electron/db/query-package.js
@@ -704,34 +704,38 @@ async function getAllPackages(db) {
  * @returns {Promise<Array>} A promise that resolves to an array of option objects, each containing the option category, code, and label.
  */
 async function getAttributeAccessInterface(db, code, packageId) {
-  const extendedQuery = `
-    SELECT
-        po.OPTION_CATEGORY,
-        po.OPTION_CODE,
-        po.OPTION_LABEL
-    FROM
-        PACKAGE_OPTION po
-    WHERE
-        po.OPTION_CODE = ?
-        AND po.PACKAGE_REF = ?
+  try {
+    const extendedQuery = `
+      SELECT
+          po.OPTION_CATEGORY,
+          po.OPTION_CODE,
+          po.OPTION_LABEL
+      FROM
+          PACKAGE_OPTION po
+      WHERE
+          po.OPTION_CODE = ?
+          AND po.PACKAGE_REF = ?
 
-    UNION 
+      UNION 
 
-    SELECT
-        c.NAME AS OPTION_CATEGORY,
-        a.STORAGE_POLICY AS OPTION_CODE,
-        a.NAME AS OPTION_LABEL
-    FROM
-        ATTRIBUTE a
-    LEFT JOIN CLUSTER c ON a.CLUSTER_REF = c.CLUSTER_ID
-    WHERE
-        a.STORAGE_POLICY = ?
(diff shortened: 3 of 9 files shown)


src-electron/db/query-package.js
```diff
@@ -690,22 +690,46 @@ async function getAllPackages(db) {
...
+    SELECT
+        a.PACKAGE_REF,
+        c.NAME AS OPTION_CATEGORY,
+        a.STORAGE_POLICY AS OPTION_CODE,
+        a.NAME AS OPTION_LABEL
+    FROM
+        ATTRIBUTE a
+    LEFT JOIN CLUSTER c ON a.CLUSTER_REF = c.CLUSTER_ID
+    WHERE
+        a.STORAGE_POLICY = ?
```
brdandu, line 727: What happens if you just do the following with no union?

`SELECT
         a.PACKAGE_REF,
         c.NAME AS OPTION_CATEGORY,
         a.STORAGE_POLICY AS OPTION_CODE,
         a.NAME AS OPTION_LABEL
     FROM
         ATTRIBUTE a
     LEFT JOIN CLUSTER c ON a.CLUSTER_REF = c.CLUSTER_ID
     WHERE
         a.STORAGE_POLICY = ?`

paulr34: that's only querying the ATTRIBUTE table so it won't work

brdandu: What will it miss and how won't that work?

paulr34: The PACKAGE_OPTION table and all the global attributes

brdandu: I see a couple of issues here:
- You are retrieving all the package option entries which have accessAttributeInterface. What happens when there are more than 1 Matter SDKs loaded? Will you be pulling in package option entries from another sdk which you are not concerned with?
- You can retrieve global attributes by checking if their cluster_ref is null

src-electron/db/query-package.js
brdandu, line 1: Add unit tests to cover the code changes here.

src-electron/db/query-package.js
paulr34: No I can't get the Cluster if it's null

paulr34: Good point about the second SDK I'll check on that

src-electron/db/query-package.js
paulr34: Good point will do

src-electron/rest/user-data.js
```diff
@@ -255,12 +255,30 @@ function httpPostCluster(db) {
...
 function httpPostForcedExternal(db) {
   return async (request, response) => {
-    let forcedExternal = await upgrade.getForcedExternalStorage(db)
+    let sessionId = request.zapSessionId
+    let packages = await queryPackage.getPackageSessionPackagePairBySessionId(
+      db,
+      sessionId
+    )
+    let packageId = packages[0].pkg.id
+    let forcedExternal = await upgrade.getForcedExternalStorage(db, packageId)
```
brdandu, line 277: pass packageIds here and at other places instead of packageId

test/matter.test.js
```diff
@@ -0,0 +1,41 @@
...
+)
+test('compute storage policy new config', async () => {
+  const result1 = await matter.computeStorageOptionNewConfig(
+    dbEnum.storagePolicy.attributeAccessInterface
+  )
+  expect(result1).toEqual(dbEnum.storageOption.external)
+  const result2 = await matter.computeStorageOptionNewConfig(
+    dbEnum.storagePolicy.any
+  )
+  expect(result2).toEqual(dbEnum.storageOption.ram)
```
brdandu, line 40: To test the query above. You might want to catch a cluster attribute accessInterface and a global attribute access interface. That way if someone ever messes that for you it will be caught in that test.

src-electron/db/query-impexp.js
```diff
@@ -737,7 +737,10 @@ WHERE
     attribute.reportable = false
   }
   if (attributeId) {
-    forcedExternal = await queryUpgrade.getForcedExternalStorage(db)
+    forcedExternal = await queryUpgrade.getForcedExternalStorage(
+      db,
+      packageIds[0]
```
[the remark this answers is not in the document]
paulr34: Okay sure sounds fine

src-electron/db/query-package.js
```diff
@@ -690,23 +690,52 @@ async function getAllPackages(db) {
...
+ *
+ * @param {Object} db - The database connection object.
+ * @param {string} code - The option code or storage policy code to query for.
+ * @param {number} packageId - The ID of the package to which the options are related.
+ * @returns {Promise<Array>} A promise that resolves to an array of option objects, each containing the option category, code, and label.
+ */
+async function getAttributeAccessInterface(db, code, packageId) {
+  try {
+    const extendedQuery = `
+      SELECT
```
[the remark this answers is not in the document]
paulr34: > Add a comment stating that this query is used to get the global attributes of clusters which cannot be gathered by the subsequent query since that only gathers access interface of attributes with a cluster reference.
> 
> 
> 
> I am confused on why we would want this interface in the attribute table and also in the zcl.json(Option_category). Seems like this data should just be in the attribute table(xml). May be with exceptions to global attributes but ideally it should all be in the xml and not package_option.

It cannot be in the Attribute table because that would conflate the ideas. They are global attributes.

The rules are not in the spec. They are driven by the SDK so they should not be included in a table that represents spec data. 

Hope that clears it up

src-electron/db/query-package.js
```diff
@@ -690,23 +690,52 @@ async function getAllPackages(db) {
...
+          po.OPTION_CATEGORY,
+          po.OPTION_CODE,
+          po.OPTION_LABEL
+      FROM
+          PACKAGE_OPTION po
+      WHERE
+          po.OPTION_CODE = ?
+          AND po.PACKAGE_REF = ?
```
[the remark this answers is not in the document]
paulr34: Okay

ethanzhouyc (approved):

test/matter.test.js
paulr34: the global attribute will be set as attributeAccessInterface so it already does that. ill make more PRs to enhance the test coverage for these SDK functions. thank you

src-electron/db/query-config.js
```diff
@@ -235,7 +235,10 @@ async function insertOrUpdateAttributeState(
     )
     staticAttribute.defaultValue = featureMapDefaultValue
   }
-  let forcedExternal = await queryUpgrade.getForcedExternalStorage(db)
+  let forcedExternal = await queryUpgrade.getForcedExternalStorage(
+    db,
+    staticAttribute.packageRef
```
brdandu, line 240: [staticAttribute.packageRef]

src-electron/db/query-package.js
```diff
@@ -690,23 +690,67 @@ async function getAllPackages(db) {
...
+ * @param {string} code - The option code or storage policy code to query for.
+ * @param {number} packageId - The ID of the package to which the options are related.
+ * @returns {Promise<Array>} A promise that resolves to an array of option objects, each containing the option category, code, and label.
+ */
+async function getAttributeAccessInterface(db, code, packageIds) {
+  try {
+    // Ensure packageIds is always an array
+    if (!Array.isArray(packageIds)) {
+      packageIds = [packageIds]
+    }
```
brdandu, line 712: You don't need this if you pass arrays in all function calls.

src-electron/db/query-package.js
```diff
@@ -690,23 +690,68 @@ async function getAllPackages(db) {
...
-        OPTION_CODE,
-        OPTION_LABEL
-       FROM
-        PACKAGE_OPTION
-       WHERE
-        OPTION_CODE = ?`,
-      [code]
-    )
-    .then((rows) => rows.map(dbMapping.map.options))
+    let packageRefCondition = `po.PACKAGE_REF = ?`
```
brdandu, line 714: Can avoid all this complex logic if you pass single instance of packageId as [pkgId].

src-electron/db/query-package.js
```diff
@@ -690,23 +690,68 @@ async function getAllPackages(db) {
...
+    let attributePackageRefCondition = `a.PACKAGE_REF = ?`
+    let queryParams = [code, ...packageIds, code, ...packageIds]
+
+    // Since packageIds is now always an array, adjust the query and parameters accordingly
+    const placeholders = packageIds.map(() => '?').join(', ')
```
brdandu, line 719: You can do the standard:

IN (${dbApi.toInClause(packageIds)})

brdandu (approved):
Approving but would like to see array arguments and cleanup as per that.

src-electron/db/query-package.js
paulr34: ill make a phase 2 PR, thank you for your review. much appreciated

paulr34: actually, fixing now

src-electron/db/query-config.js
paulr34: thanks I fixed this in getForcedExternalStorage

Issue #1399, opened 2024-08-12
Row number discrepancy in attributes of Matter Clusters


ethanzhouyc:
**Steps to reproduce:**
1. `npm run zapmatter`
2. create an endpoint on any Matter device type (e.g. HA Color Dimmable Light)
3. navigate to the attributes of `Identify` cluster (or any other cluster)

**Issue:**
The row number displayed at the bottom of the page consistently shows 4 more rows than are actually present. For example, the row number indicates there are 8 rows, but only 4 rows are visible in the cluster.

**Screenshot**
<img width="1505" alt="Screenshot 2024-08-12 at 11 11 46 AM" src="https://github.com/user-attachments/assets/50dd455d-eda0-4dc6-9236-08e3551e3216">

PR #1401, opened 2024-08-12
Fix row number discrepancy in Matter attributes

ethanzhouyc:
- fix issue #1399

- exclude global attributes from relevantAttributeData to make row number aligns with rendered rows

### DIFF
--- a/src/util/ui-options.js
+++ b/src/util/ui-options.js
@@ -26,12 +26,6 @@ export default {
       enableServerOnly: false,
       enableSingleton: false,
       enableBounded: false,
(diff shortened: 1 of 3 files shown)


src/util/ui-options.js
```diff
@@ -26,12 +26,6 @@ export default {
       enableServerOnly: false,
       enableSingleton: false,
       enableBounded: false,
-      globalLists: [
```
paulr34, line 29: @ethanzhouyc could you leave this here and reference it?

PR #1410, opened 2024-08-22
Syncing Multi-protocol Endpoint Type Attribute Default Values

brdandu:
Syncing the endpoint type attribute default values across endpoints as per the multi-protocol.json and attribute_mapping table

- Adding a trigger to sync endpoint type attribute default values across the same endpoint identifiers when it belongs to the attribute mapping table(multi-protocol.json)
- Adding unit tests for the above into multi-protocol.test.js
- JIRA: ZAPP-1384


tecimovic (approved):
Looks ok. But I can only trust that that TRIGGER code is ok. I don't fully understand that.

PR #1414, opened 2024-08-30
Update eslint configuration and adding comment to all functions which do not have them.

brdandu:
- Update eslint to throw errors when functions don't have documentation.
- Ignoring eslint missing comments for spa, generated-html and src-electron/generator/matter
- Update comments on all the js functions
- Add eslint to pre-commit hook
- JIRA: ZAPP-1143


tecimovic (approved):
Good stuff.

Issue #1382, opened 2024-07-31
Switching to endpoint configuration from the options, extensions and notifications page.


dhchandw:
- Currently, the only way to go back to a particular endpoint configuration page from the options, extensions and notifications page is by clicking on the back button.

- **Suggested addition** - clicking on a particular endpoint on the left takes the user to the endpoint configuration page of that particular endpoint. 

- This seems intuitive since  the options, extensions and notifications page are not endpoint specific (currently).

- Note: notifications page will be changed to be a panel on the right side identical to the preview page.

<img width="1624" alt="Screenshot 2024-07-31 at 9 17 15 AM" src="https://github.com/user-attachments/assets/56045fdd-2c67-478c-8cd6-6d9faddbffa8">

I'd appreciate any thoughts on this.

tbrkollar:
notification page will not exist in the future. It will be a panel on the right site.

paulr34:
I see a contradiction. The endpoint is still highlighted even though "Options" is Endpoint agnostic

What do you think about removing the highlight in this case. That way your feature will be more intuitive to the user because they are "reselecting" an endpoint and subsequently returning to the "Endpoint" view.

dhchandw:
@paulr34 that makes sense. The endpoint shouldn't be highlighted unless we are on a page that is specific to it. Ideally, when we are on the options page for example, none of the endpoints should be highlighted but clicking on one of the endpoints should take us to the config page for that endpoint. 

I think right now the endpoint that is highlighted just signifies that this is the endpoint config you go back to when you click the back button.

PR #1418, opened 2024-09-04
can go back to endpoint config from extensions and options page

dhchandw:
#1382

### DIFF
--- a/src/components/ZclEndpointCard.vue
+++ b/src/components/ZclEndpointCard.vue
@@ -18,7 +18,7 @@ limitations under the License.
   <div class="q-mx-md q-mb-sm">
     <q-card
       :class="{ 'active v-step-5': isSelectedEndpoint }"
-      @click="setSelectedEndpointType(endpointReference)"
+      @click="handleEndpointCardClick(endpointReference)"
       flat
     >
       <div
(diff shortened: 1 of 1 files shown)


src/components/ZclEndpointCard.vue
```diff
@@ -21,111 +21,85 @@ limitations under the License.
...
+                createLogoSrc(
+                  true,
+                  getDeviceCategory(deviceType[0]?.packageRef),
+                  isSelectedEndpoint
+                )
+              "
+              alt=""
+            />
+            <strong>
+              Endpoint - {{ getFormattedEndpointId(endpointReference) }}</strong
```
tbrkollar, line 42: remove "Endpoint -" text and add "#" character. I think it was a rebase conflict.

src/components/ZclEndpointCard.vue
```diff
@@ -21,111 +21,85 @@ limitations under the License.
...
-            dense
-            v-close-popup
-            size="sm"
-            icon="o_delete"
-            @click="handleDeletionDialog"
-            data-test="delete-endpoint"
-          >
-            <q-tooltip> Delete </q-tooltip>
-          </q-btn>
+      <router-link :to="'/'" class="no-decoration">
```
tbrkollar, line 24: not use router-link for this. add @click="" handler and a function for this in a script tag.

src/components/ZclEndpointCard.vue
```diff
@@ -590,5 +596,9 @@ export default {
     color: #fff;
     background: var(--q-primary);
   }
+  .no-decoration {
```
tbrkollar, line 599: if you use a function it is not necessary

dhchandw:
use this.router.push instead

src/components/ZclEndpointCard.vue
```diff
@@ -408,6 +408,10 @@ export default {
             this.selectedReporting.push(resolvedReference)
         })
       })
+    },
+    handleEndpointCardClick(endpointReference) {
```
brdandu, line 412: When do we put a function on common-mixing vs here? If more than one places calls this function?
Add comments to this method?

brdandu (changes requested):

brdandu (approved):

PR #1427, opened 2024-09-17
dynamically enabling Matter/Zigbee only components for CMP

paulr34:
CMP components cacheing

### DIFF
--- a/src/components/ZclAttributeManager.vue
+++ b/src/components/ZclAttributeManager.vue
@@ -415,9 +415,7 @@ export default {
         sortBy: 'clientServer'
       },
       columns: [],
-      forcedExternal: [],
-      enableSingleton: false,
-      enableBounded: false
+      forcedExternal: []
     }
   },
   mounted() {
--- a/src/util/ui-options.js
+++ b/src/util/ui-options.js
@@ -106,13 +106,6 @@ export default {
       return this.enableMatterFeatures
     },
     enableServerOnly() {
-      if (this.cmpEnableMatterFeatures) {
-        console.log(this.cmpEnableMatterFeatures)
-        return true
-      } else if (this.cmpEnableZigbeeFeatures) {
-        console.log(this.cmpEnableMatterFeatures)
-        return false
-      }
       return this.enableMatterFeatures
     }
   }
--- a/src/util/ui-options.js
+++ b/src/util/ui-options.js
@@ -18,6 +18,16 @@
 export default {
   data() {
     return {
+      cmpEnableProfileId: false,
+      cmpEnableNetworkId: false,
+      cmpEnableMultipleDevice: false,
+      cmpEnablePrimaryDevice: false,
+      cmpEnableParentEndpoint: false,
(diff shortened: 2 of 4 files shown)


src/util/ui-options.js
```diff
@@ -90,7 +90,7 @@ export default {
       return this.enableMatterFeatures
     },
     enableServerOnly() {
-      return this.enableMatterFeatures && !this.enableZigbeeFeatures
```
brdandu, line 93: What about client clusters which belong to endpoints with Zigbee's zcl device types? Those should still be configurable.

src/util/ui-options.js
brdandu, line 1: Add cypress tests to make sure that client clusters of zigbee endpoint is still configurable but not matter endpoints in the case of CMP.

brdandu (changes requested):

src/util/ui-options.js
paulr34: we do not have the framework to make Zigbee or matter specific cypress tests, do you think that should block this?

src/util/ui-options.js
paulr34: yes it appears there is a bigger problem here

src/util/ui-options.js
paulr34: we definitely need cypress tests you're right!

tbrkollar: Maybe you know, but  I had started to implement a solution for detecting zigbee and matter cases in cypress. It has not been finished yet but it is working now. So we have an attribute that helps me detect which mode is active in the Cypress test. This is the "mode" attribute. 

` cy.fixture('data').then((data) => {
      if(data.mode === 'matter') {
        //code
      }
       if(data.mode === 'zigbee') {
        //code
      }
     
    })`

paulr34: that's great!

src/components/ZclAttributeManager.vue
```diff
@@ -415,9 +415,7 @@ export default {
         sortBy: 'clientServer'
       },
       columns: [],
-      forcedExternal: [],
-      enableSingleton: false,
-      enableBounded: false
```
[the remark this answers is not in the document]
paulr34: because that code is out of date with the changes to ui-options.js

src/components/ZclAttributeManager.vue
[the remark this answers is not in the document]
paulr34: there is no ticket for this because while I was solving the 1499 I noticed the whole thing was broken

src/util/ui-options.js
paulr34: it does not disable it

paulr34: to be clear, ZAP master branch, as of now, when running CMP mode never distinguishes between Zigbee and Matter. That logic is broken. So this PR fixes that.

tbrkollar (approved):
I've taken a look at this branch and I think my CMP changes are still working. So I approved this pr. Great job, @paulr34 .

PR #1431, opened 2024-09-23
Update the .zap file when Studio project updates to a new GSDK

brdandu:
- Update the uc_upgrade target for apack.json
- Add the upgrade zap file function which updates the top level json packages for the .zap file based on the gsdk being used for the project.
- The code goes through a temporary directory where .zap files are stored for Studio and updates all the .zap files in that directory
- The top level upgrade packages are matched based on type and category
- Other minor cleanup
- JIRA: ZAPP-1362


src-electron/main-process/startup.js
```diff
@@ -192,6 +192,173 @@ async function noopConvert(resultsFile, logger) {
   }
 }
 
+/**
+ * Find all zap files in a given directory
+ * @param {*} dir
+ * @returns all .zap files in the given directory
+ */
+function findZapFiles(dir) {
```
tecimovic, line 200: I thought there was already some logic that finds all the zap files in a directory? We use that for the file opening, since we only get "project directory" from UC...

tecimovic (approved):

src-electron/main-process/startup.js
brdandu: I just knew about the ${tempContentFolder} which has the .zap file for upgrade process. It is a temporary directory where the upgrade is performed. Just pass that here. Do not know about any other logic here.

paulr34 (approved):

PR #1433, opened 2024-09-25
Remove temp state because uc_upgrade now works on Studio without tempstate

brdandu:
JIRA: ZAPP-1362

### DIFF
--- a/apack.json
+++ b/apack.json
@@ -61,7 +61,7 @@
       "cmd": "$(zap-cli) --version"
     },
     "uc_generate": {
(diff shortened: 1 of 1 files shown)


paulr34:
Ive been testing this and it appears to be working well. Leaning towards approving but am going to run through some more use cases to make sure.

paulr34 (approved):
Approved

tecimovic (commented):
Paul was not wrong we need to test this well, but in general it's a right thing.

brdandu:
> Paul was not wrong we need to test this well, but in general it's a right thing.

I have tested this for matter and zigbee apps. Went through the matter application upgrade process with Michael Rupp from the matter team as well showing how it passes now.

ethanzhouyc (approved):

paulr34 (changes requested):
I would like @tecimovic to approve before merge

paulr34:
@tecimovic @brdandu 

we need testing for this change or else any change after is a risk, right?

tecimovic (approved):
Right thing to do, the question is have we done enough testing. But I assume we did, so I'm ok with this.

paulr34 (approved):

PR #1434, opened 2024-09-25
fixing CMP GUI so it does not always react + bug fix

paulr34:


src/components/ZclEndpointCard.vue
```diff
@@ -419,6 +419,15 @@ export default {
       if (this.$route.path !== '/') {
         this.$router.push({ path: '/' })
       }
+      if (this.$store.state.zap.isMultiConfig) {
+        if (categoryTmp === 'zigbee') {
```
tbrkollar, line 423: i think we should use helperCategory in this condition
dbEnum.helperCategory.matter
dbEnum.helperCategory.zigbee

src/components/ZclEndpointCard.vue
```diff
@@ -491,11 +500,27 @@ export default {
         return this.$store.state.zap.endpointView.parentEndpointIdentifier
       }
     },
+    showParentEndpointIdentifier: {
+      get() {
+        return (
+          this.getDeviceCategory(this.deviceType[0].packageRef) === 'matter'
```
tbrkollar, line 506: same here

src/components/ZclEndpointCard.vue
```diff
@@ -491,11 +500,27 @@ export default {
...
+    },
     profileId: {
       get() {
         return this.$store.state.zap.endpointView.profileId
       }
     },
+    showProfileId: {
+      get() {
+        return (
+          this.getDeviceCategory(this.deviceType[0].packageRef) === 'zigbee' &&
```
tbrkollar, line 518: and here. dbEnum.helperCategory.zigbee

src/components/ZclEndpointCard.vue
paulr34: thank you, fixed

tbrkollar (approved):
if tests are passed

PR #1436, opened 2024-09-27
bug fix so endpoint card is not reactive

paulr34:


src/components/ZclEndpointCard.vue
```diff
@@ -519,8 +519,7 @@ export default {
       get() {
         return (
           this.getDeviceCategory(this.deviceType[0].packageRef) === 'zigbee' &&
```
tbrkollar, line 521: please modify the zigbee word in this condition to DbEnum.helperCategory.zigbee

ethanzhouyc (approved):

src/components/ZclEndpointCard.vue
paulr34: fixed, thank you

tbrkollar (approved):

PR #1442, opened 2024-10-09
Update logic to process domain data

ethanzhouyc:
- Added support to process domain data from XML in both formats: `<domain>General</domain>` (current) and `<domain name="General"/>` (new)
- Make ZAP compatible with parsing CHIP XMLs updated by Alchemy


paulr34 (approved):

ethanzhouyc:
Plans of updating CHIP XMLs changed we do not need to support the new domain format anymore.

brdandu (approved):

PR #1443, opened 2024-10-09
Update the zcl and template packages during generation

brdandu:
- Create a flag called --upgradeZapFile which can be passed during generation to update the package paths during generation. add this to apack.json
- During generation use the flag to determine if packages should be updated in the .zap file. If the .zap file does not already have the packages mentioned in generation(uc_generate) then the .zap file is updated with those packages and saved.
- JIRA: ZAPP-1456

### DIFF
--- a/apack.json
+++ b/apack.json
@@ -61,7 +61,7 @@
       "cmd": "$(zap-cli) --version"
     },
     "uc_generate": {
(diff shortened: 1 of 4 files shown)


paulr34 (approved):

tecimovic (approved):
I'll say this: I don't like this. We're "on-the-fly"overwriting the input file. Someone, somewhere will say: "Why do I have a diff in this file? I didn't change anything!"

So I don't like this functionality in general, because I think it has a potential of causing more damage than not.

But it's hidden behind the CLI switch, so if someone doesn't want to use it, they can simply not pass in that switch. So I'm ok with merging this, because this is all optional, I just don't want this to me missused as some kind of silver bullet.

brdandu:
> I'll say this: I don't like this. We're "on-the-fly"overwriting the input file. Someone, somewhere will say: "Why do I have a diff in this file? I didn't change anything!"
> 
> So I don't like this functionality in general, because I think it has a potential of causing more damage than not.
> 
> But it's hidden behind the CLI switch, so if someone doesn't want to use it, they can simply not pass in that switch. So I'm ok with merging this, because this is all optional, I just don't want this to me missused as some kind of silver bullet.

Our chat for reference https://silabsiot.slack.com/archives/D67QEA77C/p1729626363562279

PR #1444, opened 2024-10-09
ZAPP-1183 added clean script to build release packages

dhchandw:
* added clean script to build release packages

* script can be used in CI and locally

* modified instructions to build binaries locally

* modified release workflow to use build-release-package.js script

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -210,15 +210,9 @@ jobs:
       - name: Build & Release for Windows / Mac universal binary on macOS
         if: startsWith(matrix.os, 'macos')
         run: |
-          npm run pack:win
-          npm run pkg:win
-          ls ./dist/
-          npm run pack:cli:win
+          node src-script/build-release-package.js --platform w
 
-          npm run pack:mac
-          npm run pkg:mac
-          ls ./dist/
-          npm run pack:cli:mac
+          node src-script/build-release-package.js --platform m
 
         env:
           GH_TOKEN: ${{ secrets.github_token }}
@@ -230,10 +224,7 @@ jobs:
       - name: Build & Release for Linux
         if: startsWith(matrix.os, 'ubuntu')
         run: |
-          npm run pack:linux
-          npm run pkg:linux
-          ls ./dist/
-          npm run pack:cli:linux
+          node src-script/build-release-package.js --platform l
 
           mv dist/zap-linux-amd64.deb dist/zap-linux-x64.deb
           mv dist/zap-linux-x86_64.rpm dist/zap-linux-x64.rpm
--- a/src-script/build-release-package.js
+++ b/src-script/build-release-package.js
@@ -1,0 +1,143 @@
+#!/usr/bin/env node
+/**
+ *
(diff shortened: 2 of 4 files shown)


src-script/build-release-package.js
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env node
```
brdandu, line 1: Just for my own recollection. Forgot what this was. Why do we do this?

src-script/build-release-package.js
```diff
@@ -0,0 +1,143 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+const scriptUtil = require('./script-util.js')
+const os = require('os')
+const path = require('path')
+const fs = require('fs')
+const yargs = require('yargs/yargs')
+const { hideBin } = require('yargs/helpers')
```
brdandu, line 24: What does hideBin do?

.github/workflows/release.yml
```diff
@@ -210,15 +210,9 @@ jobs:
       - name: Build & Release for Windows / Mac universal binary on macOS
         if: startsWith(matrix.os, 'macos')
         run: |
-          npm run pack:win
-          npm run pkg:win
-          ls ./dist/
-          npm run pack:cli:win
+          node src-script/build-release-package.js --platform w
```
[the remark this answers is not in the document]
dhchandw: This is set by the workflow script in the setup phase.

src-script/build-release-package.js
```diff
@@ -0,0 +1,143 @@
...
+      console.log(`Building for Mac... Output: ${outputPath}`)
+      await scriptUtil.executeCmd({}, 'npm', ['run', 'pack:mac']) // Building electron app
+      await scriptUtil.executeCmd({}, 'npm', ['run', 'pkg:mac']) // Building zap-cli
+      await scriptUtil.executeCmd({}, 'npm', ['run', 'pack:cli:mac']) // Adding zap-cli to zip file
+      if (outputPath) {
+        await scriptUtil.executeCmd({}, 'mv', [
+          './dist/zap-mac-x64.zip',
+          path.join(outputPath, 'zap-mac-x64.zip')
+        ])
+        await scriptUtil.executeCmd({}, 'mv', [
```
brdandu, line 43: Should we comment this one out because it does not even work?

src-script/build-release-package.js
dhchandw: #!/usr/bin/env node is an instance of a [shebang line](https://en.wikipedia.org/wiki/Shebang_(Unix)): the very first line in an executable plain-text file on Unix-like platforms that tells the system what interpreter to pass that file to for execution, via the command line following the magic #! prefix (called shebang).

From stack overflow^

brdandu (approved):

src-script/build-release-package.js
dhchandw: It just helper that simplifies command line parsing

ethanzhouyc (approved):

Issue #1441, opened 2024-10-09
ZAP Tool GUI: Events and Commands states not copied when copying endpoints


hacker-cb:
Events and Commands enabled states not copied when copying endpoints.

**How to reproduce:**

1. Open ZAP Tool
2. Create endpoint with enabled commands and events
3. Copy endpoint
4. Copied endpoint will have all attributes and events disabled.

**Original endpoint:**

<img width="1459" alt="image" src="https://github.com/user-attachments/assets/92bc4d6a-d5c3-4551-a711-495797418686">

**Duplicated endpoint:**

<img width="1464" alt="image" src="https://github.com/user-attachments/assets/6da4b2bd-0eed-496d-aa25-2f34728d04ef">

PR #1461, opened 2024-10-21
Copy commands and events when duplicating endpoint

ethanzhouyc:
- When duplicating an endpoint, only attribute states are copied. This PR extends to copy command and event states.

- fix issue #1441

- Jira: GHM_ZAP-457


paulr34 (approved):
perfect, thank you

dhchandw (approved):

PR #1467, opened 2024-10-22
Adding support for Quality tag in Commands

dhchandw:
* Added support for qualities in commands
* Added isLargeMessage quality
* Modified relevant queries and template helpers
* Also fixed issues with loading of attribute qualities
* Added tests for both command and attribute qualities


src-electron/db/db-mapping.js
```diff
@@ -125,7 +125,7 @@ exports.map = {
       isArray: x.ARRAY_TYPE ? 1 : 0,
       mustUseTimedWrite: dbApi.fromDbBool(x.MUST_USE_TIMED_WRITE),
       apiMaturity: x.API_MATURITY,
-      isChangeComitted: dbApi.fromDbBool(x.IS_CHANGE_COMITTED),
+      isChangeOmitted: dbApi.fromDbBool(x.IS_CHANGE_OMITTED),
```
brdandu, line 128: Thanks for fixing this as well!

src-electron/db/db-mapping.js
```diff
@@ -215,7 +215,8 @@ exports.map = {
       hasSpecificResponse: dbApi.toDbBool(x.RESPONSE_REF),
       isIncoming: x.INCOMING,
       isOutgoing: x.OUTGOING,
-      isDefaultResponseEnabled: x.IS_DEFAULT_RESPONSE_ENABLED
+      isDefaultResponseEnabled: x.IS_DEFAULT_RESPONSE_ENABLED,
+      isLargeMessage: dbApi.fromDbBool(x.IS_LARGE_MESSAGE)
```
brdandu, line 219: Is this the only one that the quality takes are bringing in for now in terms of commands?

src-electron/db/db-mapping.js
```diff
@@ -700,7 +701,8 @@ exports.map = {
       incoming: dbApi.fromDbBool(x.INCOMING),
       outgoing: dbApi.fromDbBool(x.OUTGOING),
       isIncoming: dbApi.fromDbBool(x.IS_INCOMING),
-      isEnabled: dbApi.fromDbBool(x.IS_ENABLED)
+      isEnabled: dbApi.fromDbBool(x.IS_ENABLED),
+      isLargeMessage: dbApi.fromDbBool(x.IS_LARGE_MESSAGE)
```
brdandu, line 705: endpointTypeCommand(any map which has a corresponding table) normally denotes to only the fields in the endpoint_type_command table. So if another table's entries are needed then create/use endpointTypeCommandExtended for it and add this there.

brdandu: I believe we have messed this up for a certain set of tables/maps but that was the idea behind the proper mapping.

src-electron/db/query-command.js
```diff
@@ -1160,6 +1163,7 @@ SELECT
   CO.IS_FABRIC_SCOPED,
   CO.RESPONSE_REF,
   CO.RESPONSE_NAME,
+  CO.IS_LARGE_MESSAGE,
```
brdandu, line 1166: Add this to the `mapFunction` above for this to be really used.

src-electron/db/zap-schema.sql
```diff
@@ -182,6 +182,7 @@ CREATE TABLE IF NOT EXISTS "COMMAND" (
   "RESPONSE_NAME" integer,
   "RESPONSE_REF" integer,
   "IS_DEFAULT_RESPONSE_ENABLED" integer,
+  "IS_LARGE_MESSAGE" integer,
```
brdandu, line 185: Is this the only quality tag attribute which is currently being added by CSA xml? Wondering if there are additional ones and if it is quicker to add those right now while you are at it.

brdandu (changes requested):

src-electron/db/db-mapping.js
dhchandw: seems to be for now based of our discussion

paulr34 (approved):

PR #1471, opened 2024-10-24
Handling Cluster Extension loading when custom XML is reloaded

dhchandw:
* Making sure cluster extension is not duplicated when custom XML is reloaded
* Added a MANUFACTURER_CODE_DERIVED column since NULL!=NULL in SQL
* Added SIDE to the UNIQUE constraint in attributes
* Added logic to catch duplicates during loading before insertion


test/resource/custom-cluster/matter-bad-custom.xml
```diff
@@ -11,6 +12,7 @@
 <clusterExtension code="0x0006">
     // intentional undefined type errors
     <attribute side="server" code="0xFFF10000" define="SAMPLE_MFG_SPECIFIC_TRANSITION_TIME_6" type="INT9" min="0x0000" max="0xFFFF" writable="true" default="0x0000" optional="true">Sample Mfg Specific Attribute 6</attribute>
+    <attribute side="server" code="0xFFF10000" define="SAMPLE_MFG_SPECIFIC_TRANSITION_TIME_6_DUPLICATE" type="INT8" min="0x0000" max="0xFFFF" writable="true" default="0x0000" optional="true">Sample Mfg Specific Attribute 6 Duplicate</attribute>
     <attribute side="server" code="0xFFF20001" define="SAMPLE_MFG_SPECIFIC_TRANSITION_TIME_8" type="INT16U" min="0x0000" max="0xFFFF" writable="true" default="0x0000" optional="true">Sample Mfg Specific Attribute 8</attribute>
```
brdandu, line 16: Just to extend testing add a duplicate command to standard xml as well so you get code coverage on insertClusters and insertClusterExtensions.

brdandu (changes requested):

src-electron/db/query-loader.js
```diff
@@ -720,7 +785,15 @@ async function insertClusterExtensions(db, packageId, knownPackages, data) {
       let pCommand = insertCommands(db, packageId, commands)
       let pAttribute = insertAttributes(db, packageId, attributes)
       let pEvent = insertEvents(db, packageId, events)
-      return Promise.all([pCommand, pAttribute, pEvent])
+      return Promise.all([pCommand, pAttribute, pEvent]).catch((err) => {
+        if (err.includes('SQLITE_CONSTRAINT') && err.includes('UNIQUE')) {
+          env.logDebug(
+            `CRC match for file with package id ${packageId}, skipping parsing.`
```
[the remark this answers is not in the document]
dhchandw: This is not an error or a warning. We get to this point if we try to reload a cluster extension for a top-level package that already has it.

test/test-util.js
```diff
@@ -170,12 +170,12 @@ exports.testMatterCustomZap =
 
 exports.totalClusterCount = 111
 exports.totalDomainCount = 20
-exports.totalCommandArgsCount = 1786
```
paulr34, line 173: I am always a little cautious about tests changing...could you explain to me why you did this please?

dhchandw: Some of the unique constraints in the schema were not being hit because for standard clusters, attributes, etc we had manufacturer code set to NULL and NULL does not equal to NULL in SQL. When I added a derived column to circumvent this, I found a duplicate command in `zcl-builtin/silabs/demo.xml`. With my changes this duplicate command wouldn't be loaded in and hence the command count decreased by one.

paulr34 (approved):

test/test-util.js
paulr34: cool, approved

paulr34: assuming CI passes of course

brdandu:
Approved but make sure unify codeine passes before merge

brdandu:
The PR looks good. Please take a look at the unify code gen CI failure.

brdandu (approved):
Assuming codegen failure is due to the fix

PR #1475, opened 2024-10-28
Update the top level template json file based on ZCL extensions

brdandu:
- Add the zcl extension files to part of its crc check.
- If the crc changes based on the inclusion of extensions then reload the top level template json file and turn its IN_SYNC to true and turn the old on to false
- These changes make sure that when a zcl extension file (eg json files in zcl section such as cluster-to-component.json) is updated then everything from the top level json files are loaded again
- JIRA: ZAPP-1092


paulr34 (approved):
looks good

src-electron/util/env.js
```diff
@@ -73,14 +73,25 @@ export function builtinSilabsZclGeneralXmlFile() {
...
 }
 
+/**
+ * Used to retrieve cluster-to-component-dependencies.json by zcl reload test in zcl-loader.test.js
+ *
+ * @returns path to cluster-to-component-dependencies.json file used by zcl-loader.test.js
+ */
+export function builtinZclExtensionClusterToComponentFile() {
+  return locateProjectResource(
+    './test/gen-template/zigbee2/cluster-to-component-dependencies.json'
```
tecimovic, line 91: env.js is not test file. But it refers to a test file here. You should put this in `test-util`, not env.js

Here we keep "zcl-builtin" stuff, but `test` stuff is not even packaged.

tecimovic (changes requested):

src-electron/util/env.js
brdandu: Done

tecimovic (approved):

brdandu:
Unify CI failure is a known failure and is being looked into. Will be disabled for the short term.

PR #1476, opened 2024-10-31
Add helper to determine the atomic identifier for attribute Id given

brdandu:
- as_underlying_atomic_identifier_for_attributeId checks for the attribute type and returns the atomic identifier associated with it.
- JIRA: ZAPP-1385

### DIFF
--- a/src-electron/generator/helper-attribute.js
+++ b/src-electron/generator/helper-attribute.js
@@ -22,7 +22,10 @@
  */
 
 const queryAttribute = require('../db/query-attribute')
+const queryZcl = require('../db/query-zcl')
 const templateUtil = require('./template-util')
+const zclUtil = require('../util/zcl-util')
+const dbEnum = require('../../src-shared/db-enum')
 
 /**
  * Get feature bits from the given context.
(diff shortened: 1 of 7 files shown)


tecimovic (approved):

src-electron/generator/helper-attribute.js
```diff
@@ -76,6 +79,89 @@ async function attributeDefault(options) {
...
+    : null
+  // If attribute type directly points to the atomic type then return that
+  if (atomicInfo) {
+    // All types in the number and string table should be found here.
+    return atomicInfo.atomicId
+  } else {
+    // If attribute type does not point to atomic type
+    let dataType = await queryZcl.selectDataTypeByNameAndClusterId(
+      this.global.db,
+      attributeDetails.type,
```
tecimovic, line 109: This whole block of code is a bit weird.... Mining in all these different places for a name: I can understand where it comes from, but shouldn't there be a better way to do this? Just one table "TYPES" that contains all known types, be it enum values or bitmap values or whatever?

I wonder if having separate tables for these was a mistake...

brdandu: All types are in the data_type table but that only has the common characteristics with the distinct attributes of each type going into their corresponding tables.

src-electron/generator/helper-attribute.js
```diff
@@ -76,6 +79,89 @@ async function attributeDefault(options) {
   return templateUtil.templatePromise(this.global, p)
 }
 
+/**
+ * Given an attribute Id determine its corresponding atomic identifier from the
+ * atomic table.
+ * @param {*} attributeId
+ */
+async function as_underlying_atomic_identifier_for_attribute_id(attributeId) {
```
dhchandw, line 87: I might be wrong but doesn't matter have derived types from integers and floats. What would happen in that case?

brdandu: Don't know of any derived floats, ints or strings. Could you point me to any you ran across them?
if that is the case then need to add additional logic to handle that.

dhchandw: does selectAtomicType() take care of that?

brdandu: I don't think we have any other tags apart from enum, bitmap and struct so I doubt it.

dhchandw: Okay I'm getting confused here, we don't NEED to return baseline atomic types here right?

dhchandw: So returning a derived atomic data type like clusterId would be acceptable here?

brdandu: does selectAtomicType() take care of that? => Yes that takes care of strings, ints, floats

brdandu: This has nothing to do with base types. It is about returning the Atomic_identfier from the atomic table based on attribute type

dhchandw: Got it, sorry for the confusion.

dhchandw (approved):

PR #1481, opened 2024-11-12
Attribute validation defaults to minimum and maximum defined by type if not defined by attribute.

dhchandw:
- Also got rid of `isValidSignedNumberString()` since it was redundant code.

ZAPP-1536

### DIFF
--- a/src-electron/validation/validation.js
+++ b/src-electron/validation/validation.js
@@ -291,22 +291,22 @@ async function getBoundsInteger(attribute, typeSize, isSigned) {
   return {
     min: attribute.min
       ? await getIntegerFromAttribute(attribute.min, typeSize, isSigned)
-      : getTypeBound(typeSize, isSigned, true),
+      : getTypeRange(typeSize, isSigned, true),
     max: attribute.max
       ? await getIntegerFromAttribute(attribute.max, typeSize, isSigned)
-      : getTypeBound(typeSize, isSigned, false)
+      : getTypeRange(typeSize, isSigned, false)
   }
 }
 
 /**
- * Gets the bound of an integer type.
+ * Gets the range of an integer type.
  *
  * @param {*} typeSize
  * @param {*} isSigned
  * @param {*} isMin
  * @returns integer
  */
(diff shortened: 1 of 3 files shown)


src-electron/validation/validation.js
```diff
@@ -315,11 +291,26 @@ async function getBoundsInteger(attribute, typeSize, isSigned) {
...
+
+/**
+ * Gets the bound of an integer type.
+ *
+ * @param {*} typeSize
+ * @param {*} isSigned
+ * @param {*} isMin
+ * @returns integer
+ */
+function getTypeBound(typeSize, isSigned, isMin) {
```
brdandu, line 309: getTypeRange?

src-electron/validation/validation.js
```diff
@@ -315,11 +291,26 @@ async function getBoundsInteger(attribute, typeSize, isSigned) {
   return {
     min: attribute.min
       ? await getIntegerFromAttribute(attribute.min, typeSize, isSigned)
```
brdandu, line 293: Not a big fan of this. Assign this to a variable and then check on that. Is more readable.

src-electron/validation/validation.js
```diff
@@ -128,20 +128,6 @@ async function validateSpecificAttribute(
       //Interpreting float values
       if (!checkAttributeBoundsFloat(attribute, endpointAttribute))
         defaultAttributeIssues.push('Out of range')
-    } else if (await types.isSignedInteger(db, zapSessionId, attribute.type)) {
```
brdandu, line 131: We do not need this if case anymore?

brdandu (changes requested):

src-electron/validation/validation.js
dhchandw: We would need to check attribute.min first and then call getIntegerFromAttribute. Ideally we don't want to call getTypeBound if not required, since, validation is done at every keyboard entry.

dhchandw: How would you suggest breaking this up?

brdandu (approved):

PR #1409, opened 2024-08-22
Feature Page phase 1

ethanzhouyc:
**Phase 1 achievements:**
- parsed feature conformance from XML data and stored it in the database.
- displayed all device type features on a frontend page

**Steps to access the feature page:**
1. create an endpoint using the Matter Dimmable Light device type (currently, this is the only device type with associated features in our XML files).
2. click the "Feature" button on the endpoint card in the sidebar.
3. the feature page, as shown in the image below, should be visible.

**To be discussed:**
- enhance the frontend UI design.
- decide which data should be presented to users.
- plan Phase 2 to enable feature toggling.

image

Issue #1380, opened 2024-07-30
Matter Device Type Features page


paulr34:
**A New Features Page**

**Requirements:**

**Phase 1:**

- Create a new page for displaying features per Device Type.
- Update Matter XML to include relevant feature data
- Design the ZAP SQL database accordingly.
- Translate conformance from the Matter spec to the ZAP SQL database and then to the Vue front end.

**Description:**

When creating Matter device types in ZAP, we propose adding a new page called "Features." This page will display all the features listed in the Element Requirements section of the Device Type Matter spec. It will also include data from the Element Requirements section, such as Conformance. This data must be extracted from the Matter data model XML, translated, stored in the ZAP SQL database, and retrieved by the front end via a REST API.

**Use Case:**

As a ZAP user creating a Matter application, I want to select a device type and activate features. By selecting "Features," I should see all possible features supported by the Device Type. This will provide a convenient reference for setting the featureMap attribute bit when finalizing my application.

**Requirements:**

**Phase 2:**

- Implement the ability to toggle optional features on or off.
- When toggling a feature, update the associated elements to correct conformance, including attributes, commands, events, features, and automatically update the feature map attribute according to the configured features.
- When toggling an element, send a warning if it does not conform
- Update Matter XML to include conformance of attributes and commands in CHIP repo

**Description:**

This page should be configurable like Attributes, Commands, and Events. It will list the features supported by each Device Type. Mandatory features should always be enabled in the GUI. Deprecated, disallowed, or provisional features should always be disabled. Optional features should be toggled on or off based on the user's application needs.

**Use Case:**

As a ZAP user creating a Matter application, I want to choose a device type and activate features. Selecting "Features" should show all possible features for the Device Type and allow me to toggle features. The system should automatically set the corresponding bit in the featureMap attribute, and adjust associated Attributes, Commands, Feature, etc. to ensure conformance, eliminating the need for manual configuration.
If a device type feature has complex conformance and doesn't comply after toggling, I should receive a warning and my changes will be disabled, instead of auto-updating elements for me. Changes to the same feature within a cluster and endpoint should sync across multiple device types. If I modify an attribute, command, or feature that causes incorrect conformance, I should receive warnings.

**Requirements:**

**Phase 3:**

- Feature Page also include features from cluster XMLs that are not present in the device type XML. In phase 2, it only supports features defined in the device type XML.
- Move the feature page UI into a per cluster view in parallel with attribute, command, and event page
- Add Vue tutorial and documentation

[...]

PR #1482, opened 2024-11-13
Feature Page Phase 2

ethanzhouyc:
- Related ZAP PR: #1409 
- Related Issue: #1380 
- JIRA: GHM_ZAP-440 and 10 subtasks under it
- CHIP PR to add conform data to XML: https://github.com/project-chip/connectedhomeip/pull/36233

### New Features:

 **1. Enable toggle of device type features** 
- automatically toggle elements (attributes, commands, and events) to correct conformance
- update `featureMap` attribute for the cluster of toggled feature
- add warning notification if a mandatory feature is disabled or a provisional one is enabled
- disable feature updates and set warnings if conformance is too complex to manage

**2. Add conformance warning to elements**
- display icon with warning messages and notifications for elements with incorrect conformance
- sync warnings across locations and ensure compatibility with existing warnings from device type requirements
- update warnings after feature and element toggles

**3. Other improvements:**
- show Device Type Feature button and element conform warnings only if conformance data exist, ensuring compatibility with older Matter SDK
- disable user from changing `featureMap` attribute
- add unit tests for backend functions
- update svg file

### Example Use Cases:

**1. Enable `HS` feature in an endpoint with `Extended Color Light` device type**

image

**2. After enabling `HS`, disabling attribute `CurrentHue` that conforms to `HS` will cause conformance warnings**

image


src/components/ZclDeviceTypeFeatureManager.vue
```diff
@@ -89,20 +188,218 @@ export default {
...
+        do not set featureMap attribute */
+      let disabled = this.updateElementsAndSetWarnings(
+        featureData,
+        inclusionList
+      )
+      if (!disabled) {
+        this.setFeatureMapAttribute(featureData)
+      }
+    },
+    setFeatureMapAttribute(featureData) {
```
paulr34, line 207: do we want to set the featureMap attribute as the bit or the code?

paulr34 (approved):
Great job @ethanzhouyc !!!!!

paulr34:
We will need to come up with a plan for maintaining Conformance rules in Matter because if Conformance changes and the XML is not updated then this project will do more harm than good.

src-electron/db/query-command.js
```diff
@@ -2046,6 +2048,50 @@ async function selectNonManufacturerSpecificCommandDetailsFromAllEndpointTypesAn
   )
 }
 
+/**
+ * Get all commands in an endpoint type cluster
+ * Non-required commands are not loaded into ENDPOINT_TYPE_COMMAND table,
```
brdandu, line 2053: Are non-required attributes loaded into endpoint_type_attribute table? I don't see deviceTypeClusterId being used in selectAttributesByEndpointTypeClusterId

brdandu (changes requested):
Will need some more time to go through this because of its size.

src-electron/db/query-config.js
```diff
@@ -222,14 +223,16 @@ async function insertOrUpdateAttributeState(
...
       await queryDeviceType.selectDeviceTypeFeaturesByEndpointTypeIdAndClusterId(
         db,
         endpointTypeId,
         clusterRef
       )
-    let featureMapBitsToBeEnabled =
-      mandatoryFeaturesOnEndpointTypeAndCluster.map((f) => f.featureBit)
+    // only set featureMap bit to 1 for mandatory features
+    let featureMapBitsToBeEnabled = featuresOnEndpointTypeAndCluster
+      .filter((f) => f.conformance == 'M')
```
brdandu, line 234: Why do we filter for just Mandatory? Were we doing something wrong before?

src-electron/db/query-feature.js
```diff
@@ -24,53 +24,612 @@ const dbApi = require('./db-api.js')
...
-        dc.INCLUDE_SERVER,
-        dc.INCLUDE_CLIENT,
-        df.DEVICE_TYPE_CLUSTER_CONFORMANCE,
-        f.FEATURE_ID,
-        f.NAME AS FEATURE_NAME,
-        f.CODE,
-        f.BIT,
-        f.DEFAULT_VALUE,
-        f.DESCRIPTION
+			d.DESCRIPTION AS DEVICE_TYPE_NAME,
```
brdandu, line 52: Fix the formatting here.

brdandu (changes requested):

src-electron/db/db-mapping.js
```diff
@@ -719,6 +727,45 @@ exports.map = {
     }
   },
 
+  endpointTypeClusterAttribute: (x) => {
```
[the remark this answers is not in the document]
ethanzhouyc: Will delete all 3 of them

src-electron/db/query-command.js
ethanzhouyc: Yes, if you create an endpoint and check ENDPOINT_TYPE_ATTRIBUTE table, there will be attributes with INCLUDED = 0, so no deviceTypeClusterId is needed.

src-electron/db/query-config.js
ethanzhouyc: Before we only loaded mandatory features to the database. Now features with all types of conformance exist, so we restrict to only enable bits for mandatory features by default

src-electron/db/query-feature.js
ethanzhouyc: The format is in good shape if you click the actual file. Github is displaying it wrong weirdly

ethanzhouyc:
PR #1483 will help fix current fails in the check

src-electron/db/query-config.js
tecimovic, line 234: Hardcoding "M" seems suspicious. This is:
   1.) Matter specific. Zigbee knows nothing about what on earth "M" is, so putting "M" anywhere in the code, is a data point from Matter spec, making way into Zap codebase. And this is in `db` package, which is definitely not Matter specific.

src-electron/db/query-feature.js
```diff
@@ -24,53 +24,612 @@ const dbApi = require('./db-api.js')
...
+) {
+  let arg = []
   let deviceTypeRefsSql = deviceTypeRefs.map(() => '?').join(', ')
+  arg.push(...deviceTypeRefs)
+  arg.push(endpointTypeRef)
   let features = await dbApi.dbAll(
     db,
     `
     SELECT
-        d.DESCRIPTION AS DEVICE_TYPE_NAME,
```
tecimovic, line 39: Why are `d` and `dc` and so lowercase? I thought we generally put uppercase for the temp aliases.

src-electron/db/query-feature.js
```diff
@@ -24,53 +24,612 @@ const dbApi = require('./db-api.js')
...
+ * @param {*} endpointTypeRef
  * @returns All feature information and device type conformance
- * with associated device type and cluster details
+ * with associated device type, cluster, and featureMap attribute details
  */
-async function getFeaturesByDeviceTypeRefs(db, deviceTypeRefs) {
+async function getFeaturesByDeviceTypeRefs(
+  db,
+  deviceTypeRefs,
+  endpointTypeRef
```
tecimovic, line 42: I don't understand why "get feature by device type" needs an "endpoint type" all of suddenly?

src-electron/db/query-feature.js
```diff
@@ -24,53 +24,612 @@ const dbApi = require('./db-api.js')
...
+ * The '[]' indicates optional conformance if the expression inside true.
+ * Expression containing comma means otherwise conformance. See spec for details.
+ * Examples of conformance expression: 'A & (!B | C)', 'A & B, [!C]'
+ *
+ * @export
+ * @param {*} expression
+ * @param {*} elementMap
+ * @returns 'mandatory', 'optional', 'provisional', or 'notSupported'
+ */
+function evaluateConformanceExpression(expression, elementMap) {
```
tecimovic, line 130: All these functions are NOT database query functions. The point of `db/query-xxxx.js` is to contain SQL queries.

The functions related to features, but NOT specifically database queries, should be elsewhere.
Move these functions to some other package.

src-electron/db/query-feature.js
```diff
@@ -24,53 +24,612 @@ const dbApi = require('./db-api.js')
...
+      result.warningMessage =
+        'On endpoint ' +
+        endpointId +
+        ', feature ' +
+        featureName +
+        ' is enabled, but it is still provisional for device type ' +
+        deviceTypeNames
+      result.displayWarning = added
+    }
+    if (conformance == 'mandatory') {
```
tecimovic, line 359: These seem matter specific. I don't know where string "mandatory" is coming from, but it seems specific to Matter spec?

src-electron/rest/user-data.js
```diff
@@ -84,12 +84,123 @@ function httpGetEndpointIds(db) {
  */
 function httpGetDeviceTypeFeatures(db) {
   return async (request, response) => {
-    let deviceTypeRefs = request.query.deviceTypeRefs
-    let deviceTypeFeatures = await queryFeature.getFeaturesByDeviceTypeRefs(
+    let { deviceTypeRefs, endpointTypeRef } = request.query
```
tecimovic, line 87: Why does this need endpointTypeRef?

src-electron/rest/user-data.js
```diff
@@ -84,12 +84,123 @@ function httpGetEndpointIds(db) {
...
+ * in an endpoint type cluster
+ */
+async function getEndpointTypeElements(
+  db,
+  endpointTypeClusterId,
+  deviceTypeClusterId
+) {
+  let elements = {}
+  elements.attributes =
+    await queryAttribute.selectAttributesByEndpointTypeClusterId(
```
tecimovic, line 116: If you have an async function that calls 3 async functions that are unrelated, a much better pattern is to do Promise.all() on them, so you allow them to execute in parallel.

So instead of:
   await promise1
   await promise2
   await promise3
   
You do:
    Promise.all([promise1, promise2, promise3])

src-electron/rest/user-data.js
```diff
@@ -84,12 +84,123 @@ function httpGetEndpointIds(db) {
...
+    // check element conform and return elements that need to be updated
+    let result = queryFeature.checkElementConformance(
+      elements,
+      featureMap,
+      featureData,
+      endpointId
+    )
+
+    // set device type feature warning
+    await querySessionNotification.setNotificationOnFeatureChange(
```
tecimovic, line 161: Same as above: use parallelism.

src-electron/rest/user-data.js
```diff
@@ -84,12 +84,123 @@ function httpGetEndpointIds(db) {
...
+    response.status(StatusCodes.OK).json(result)
+  }
+}
+
+/**
+ * HTTP POST: required and unsupported cluster elements based on conformance
+ * @param {*} db
+ * @returns callback for the express uri registration
+ */
+function httpPostSetRequiredElements(db) {
```
tecimovic, line 189: I'm confused with the name of this function..... Why do you need to POST the required elements? Aren't the required elements calculated on the back-end and at best sent over to the front-end as result of some "POST" that sets feature bits or something like that?

src-electron/rest/user-data.js
```diff
@@ -1008,6 +1119,56 @@ function httpPostDuplicateEndpointType(db) {
...
+function httpPatchUpdateBitOfFeatureMapAttribute(db) {
+  return async (request, response) => {
+    let { featureMapAttributeId, newValue } = request.body
+    let updated = await queryConfig.updateEndpointTypeAttribute(
+      db,
+      featureMapAttributeId,
+      [['defaultValue', newValue]]
+    )
+    response.status(StatusCodes.OK).json({
+      succesesful: updated > 0
```
tecimovic, line 1137: Typo? "succesesful" ?

src-electron/rest/user-data.js
```diff
@@ -1008,6 +1119,56 @@ function httpPostDuplicateEndpointType(db) {
...
+ * Set warning for the required element, and delete its existing warning if any.
+ *
+ * @param {*} db
+ * @returns response of setting the warning notification
+ */
+function httpPostRequiredElementWarning(db) {
+  return async (request, response) => {
+    let resp = await querySessionNotification.setRequiredElementWarning(
+      db,
+      request.body,
```
tecimovic, line 1165: It seems unsafe and suspicious to post the whole request body straight into the query. This feels wrong somehow.

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -2283,18 +2289,36 @@ async function parseFeatureFlags(db, packageId, featureFlags) {
...
+ * @returns The conformance string.
+ */
+function parseConformanceRecursively(operand, parentJoinChar = '') {
+  const baseLevelTerms = ['feature', 'condition', 'attribute', 'command']
   if (operand.mandatoryConform) {
     let insideTerm = operand.mandatoryConform[0]
     // Recurse further if insideTerm is not empty
     if (insideTerm && Object.keys(insideTerm).toString() != '$') {
-      return parseFeatureConformance(operand.mandatoryConform[0])
+      return parseConformanceRecursively(operand.mandatoryConform[0])
```
tecimovic, line 2321: I generally like to pass a "depth" to any recursive function. You start with "depth = 0" and then whenever you call recursively, you call it with "depth+1". When depth gets to like 200, you know you're doing something wrong, you blow up with a clear exception, instead of some stupid stack overflow which is hard to track down.

But I'm possibly old school....

src/components/ZclDeviceTypeFeatureManager.vue
```diff
@@ -89,20 +188,218 @@ export default {
...
+        this.commandsToUpdate.length == 0 &&
+        this.eventsToUpdate.length == 0
+      )
     }
   },
   data() {
     return {
       nodataMessage: 'No device type features available for this endpoint',
+      noElementsToUpdateMessage:
+        'No elements need to be updated after togglging this feature',
```
tecimovic, line 388: Typo: "togglging"

tecimovic (changes requested):
Lost of stuff..... I marked some obvious things, including some typos.

src-electron/db/query-config.js
```diff
@@ -33,6 +33,7 @@ const queryCommand = require('./query-command.js')
 const restApi = require('../../src-shared/rest-api.js')
 const _ = require('lodash')
 const notification = require('../db/query-session-notification.js')
+const queryEndpointType = require('./query-endpoint-type.js')
```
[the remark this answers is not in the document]
ethanzhouyc: all .js suffix in this PR is removed.

ethanzhouyc:
All typos and SQL related comments are addressed

src-electron/db/query-config.js
ethanzhouyc: The entire code section of setting featureMap bits here is too Matter-specific and should be restructured in the future. I created an issue on this: #1488

src-electron/db/query-feature.js
ethanzhouyc: ENDPOINT_TYPE_ATTRIBUTE and ATTRIBUTE table is joined to get info on:
1. featureMap value related to the feature, deciding if the feature is enabled
2. attribute id of featureMap, which is needed when updating the featureMap after toggling a feature

src-electron/zcl/zcl-loader-silabs.js
ethanzhouyc: That's great advice for keeping recursive functions safe, but this function is already safeguarded against infinite recursion levels, as:
1. each recursive call moves one layer deeper into the xml data object
2. the `if/else if/else` structure ensures that all cases have a return value

src-electron/rest/user-data.js
ethanzhouyc: You are right. I changed it to a GET request as it only gets resources from backend and not change them

src-electron/rest/user-data.js
ethanzhouyc: I improved this by only getting relevant variables passed from frontend, creating a new object out of them, and passing the new object to backend

src-electron/rest/user-data.js
ethanzhouyc: Fixed in the new commit

src-electron/rest/user-data.js
ethanzhouyc: There are dependencies or condition among await calls in this function so I keep it unchanged

src-electron/zcl/zcl-loader-silabs.js
tecimovic: You are not guarding against yourself. You are guarding against a future developer, who might add things to this function, that would mistakenly cause this to end up in stack overflow.

Generally, be always aware who you're developing for:
   1.) For you, TODAY.
   2.) For you, TOMORROW.
   3.) For a random guy that has no clue and shows up here TOMORROW.

tecimovic: Not a terribly big deal though, not something to hold this up for.

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -2303,65 +2327,58 @@ function parseFeatureConformance(operand) {
...
+      .map((term) => {
+        let nt = parseConformanceRecursively(term)
+        return nt.includes('&') || nt.includes('|') ? `!(${nt})` : `!${nt}`
+      })
+      .join(` ${parentJoinChar} `)
+  } else if (operand.andTerm || operand.orTerm) {
+    // process andTerm and orTerm in the same logic
+    // when joining multiple orTerms inside andTerms, we need to
+    // surround them with '()', vice versa for andTerms inside orTerms
+    // e.g. A & (B | C) or A | (B & C)
```
tecimovic, line 2352: You should use antlr for these kid of things in the future.

src-electron/zcl/zcl-loader-silabs.js
ethanzhouyc: I added a max depth check on the recursive function as you said so someone in the future would get an error if they blow up

src-electron/zcl/zcl-loader-silabs.js
ethanzhouyc: Yes we could switch to ANTLR if conformance parsing became too complicated to be parsed one day

ethanzhouyc:
There will be future cleanup PRs to separate feature related code from Zigbee

PR #1483, opened 2024-11-14
Restructure getEndpointIds function to pass ui.test.js

ethanzhouyc:
- call API directly without action
- check if $serverGet is defined before calling it
- the change would help #1482 pass unit tests and prevent `ui.test.js` from failing in the future

### DIFF
--- a/src/components/ZclCreateModifyEndpoint.vue
+++ b/src/components/ZclCreateModifyEndpoint.vue
@@ -544,9 +544,11 @@ export default {
       }
     },
     getEndpointIds() {
-      this.$store.dispatch('zap/getEndpointIds').then((res) => {
-        this.endpointIds = res.data
-      })
+      if (this.$serverGet != null) {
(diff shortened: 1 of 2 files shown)


src/components/ZclCreateModifyEndpoint.vue
```diff
@@ -544,9 +544,11 @@ export default {
       }
     },
     getEndpointIds() {
-      this.$store.dispatch('zap/getEndpointIds').then((res) => {
-        this.endpointIds = res.data
-      })
+      if (this.$serverGet != null) {
```
brdandu, line 547: When is `this.$serverGet ` null? What is the reason for this change?

ethanzhouyc: `ui.test.js` call `mount` and `shallowMount` functions to simulate creating frontend components and in this case `$serverGet` is null. This check is very common in frontend files

paulr34 (approved):

tecimovic (approved):

PR #1486, opened 2024-11-20
updating to macos-14

dhchandw:
also using 7zip pre-installed library instead of using node module

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -148,7 +148,7 @@ jobs:
     # Platforms to build on/for
     strategy:
       matrix:
-        os: [macos-12, ubuntu-22.04]
+        os: [macos-14, ubuntu-22.04]
       fail-fast: false
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -272,7 +272,7 @@ jobs:
       - name: Verify zap-cli exists in Windows arm64 .zip package
         if: startsWith(matrix.os, 'macos')
         run: |
-          output=$(./node_modules/7zip-bin/mac/x64/7za l ./dist/zap-win-arm64.zip)
+          output=$(7za l ./dist/zap-win-arm64.zip)
```
tecimovic, line 275: I thought that's what `npx` is for? Shouldn't you do `npx 7za` ? I am confused why are you changing this path.

It is true, that running `./node_modules` like it was is wrong, but `npx` should do that for you.

tecimovic (approved):
I'm ok with it, but to check of this will work, since I thought you need npx there....

.github/workflows/release.yml
dhchandw: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md

7zip comes pre-installed as a utility in the macos-14 environment. I could change it to use the node module though.

tecimovic (approved):

PR #1487, opened 2024-11-21
fix: add a condition to to toolbar about q-electron-drag

tbrkollar:


tecimovic (approved):

tecimovic (approved):

tecimovic (approved):

tbrkollar:
> In my opinion we should leave some space between the border of the application and the device type list. Screenshot 2024-11-21 at 9 43 57 AM

It's from Quasar. I don't have a quick fix for this. Unfortunately Setting margin-top: 15px, for example, is only half of the solution. Sometimes the option displays as a dropdown (depending on the number of options). It is the incorrect method because there will be 15 pixels between the input and the dropdown.

dhchandw:
> > In my opinion we should leave some space between the border of the application and the device type list. Screenshot 2024-11-21 at 9 43 57 AM
> 
> It's from Quasar. I don't have a quick fix for this. Unfortunately Setting margin-top: 15px, for example, is only half of the solution. Sometimes the option displays as a dropdown (depending on the number of options). It is the incorrect method because there will be 15 pixels between the input and the dropdown.

Can we just always use a dropdown since that looks cleaner.

tbrkollar:
> Can we just always use a dropdown since that looks cleaner.

It's also from Quasar. :) it is depending on the number of options and the window height. 
when the space after the input is enough for the options the quasar put the options like a dropdown, if not, put the top of the input. 
Maybe, one solution is that we set the max-height on the dropdown list.

tbrkollar:
> > Can we just always use a dropdown since that looks cleaner.
> 
> It's also from Quasar. :) it is depending on the number of options and the window height. when the space after the input is enough for the options the quasar put the options like a dropdown, if not, put the top of the input. Maybe, one solution is that we set the max-height on the dropdown list.

But this is not this bug; it is an improvement. :)

paulr34 (approved):

dhchandw (approved):

PR #1489, opened 2024-11-21
Enable Multiple Endpoints and Device Types in Device Composition Insertion

paulr34:
Pull Request Summary
Title: Enable Multiple Endpoints and Device Types in Device Composition Insertion

Description:

This update enhances the insertDeviceComposition function to support multiple endpoints and multiple device types (deviceType) per endpoint. Previously, the code could only insert data for a single endpoint and device type.

Key Changes:

Multiple Endpoints: The function now processes all endpoints in deviceType.composition.endpoint, inserting data for each.
Multiple Device Types per Endpoint: Supports inserting multiple deviceType values per endpoint.
Dynamic Conformance and Constraint: Uses endpoint-specific conformance and constraint values when available, with fallbacks to default deviceType values.
Improved Error Handling: Ensures required data is present and correctly formatted, preventing errors.
Benefits:

Scalability: Handles more complex device compositions with multiple endpoints and device types.
Flexibility: Supports a wider range of device configurations.
Data Integrity: Ensures correct insertion of data for each endpoint and device type.
This update improves flexibility and scalability for device composition handling.

### DIFF
--- a/src-electron/db/query-loader.js
+++ b/src-electron/db/query-loader.js
@@ -1167,11 +1167,7 @@ async function getEndpointCompositionIdByCode(db, deviceType) {
  */
 function insertDeviceComposition(db, deviceType, endpointCompositionId) {
   // Ensure that deviceType and its necessary properties are defined
-  if (
-    !deviceType ||
-    !deviceType.composition ||
-    !deviceType.composition.endpoint
-  ) {
+  if (!deviceType?.composition?.endpoint) {
     throw new Error('Invalid deviceType object or endpoint data')
   }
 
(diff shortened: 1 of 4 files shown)


aronrosenberg (approved):

dhchandw (approved):
We could add slightly more detailed test for this

src-electron/db/query-loader.js
```diff
@@ -1151,35 +1151,81 @@ async function getEndpointCompositionIdByCode(db, deviceType) {
...
       deviceType.constraint
-    ])
-  } catch (error) {
-    console.error('Error inserting device composition:', error)
-    throw error // Re-throw the error after logging it
+
+    // Create insert queries for each deviceType in this endpoint and add them to the insertQueries array
+    for (let j = 0; j < deviceTypes.length; j++) {
+      const device = deviceTypes[j]
+
```
tecimovic, line 1208: Why not:
```
   for ( let device of deviceTypes) { 
      ...
   }
```
I don't think you're using the actual index anywhere, and while there is something cozy about the good old fashioned 1960's C-like for loop, it is visually more taxing than modern JS.

tecimovic (approved):
Generally ok, but do please clean up some of those syntactic sugar thingies.... We should write modern code.

tecimovic:
Obviously I'm assuming that you tested this well with both Matter and Zigbee, and that you promise you will write unit tests for it!

PR #1497, opened 2024-12-13
Clean up the duplicate dirty flag mechanism.

tecimovic:
0. Remove the whole process of registering a second socket listener to the same event.
1. Retain only one dirtyFlag notifier.
2. Add the transformation from the default ZAP data format, towards the IDE data format.
3. Remove state transitions in VueX store. These are not needed at all.


paulr34:
Tested, and unfortunately this does not work

paulr34 (approved):

ethanzhouyc (approved):

Issue #1500, opened 2024-12-23
struct_contains_array must calculate transitive dependencies


gmarcosb:
ZAP templates currently decide how to generate `Type` vs `DecodableType` [based on `struct_contains_array`](https://github.com/project-chip/connectedhomeip/blob/4865e5e689807e500bde212cceaf36e9524b6dde/src/app/zap-templates/partials/cluster-objects-struct.zapt#L15)

However, this only works if the array is 1-level deep; if the array is contained > 1 level (a struct referencing a struct with an array field), [the resulting `cluster-objects` generated file is invalid](https://github.com/project-chip/connectedhomeip/actions/runs/12418247992/job/34671039234?pr=36877)

brdandu:
Seems like this has been resolved. @gmarcosb ?

PR #1501, opened 2024-12-23
Take into consideration transitive dependencies, not just root struct fields

gmarcosb:
Fix for https://github.com/project-chip/zap/issues/1500

### DIFF
--- a/src-electron/db/db-mapping.js
+++ b/src-electron/db/db-mapping.js
@@ -378,8 +378,7 @@ exports.map = {
       isNullable: dbApi.fromDbBool(x.IS_NULLABLE),
       isOptional: dbApi.fromDbBool(x.IS_OPTIONAL),
       isFabricSensitive: dbApi.fromDbBool(x.IS_FABRIC_SENSITIVE),
-      dataTypeReference: x.TYPE,
-      dataTypeReferenceName: x.DATA_TYPE_REF_NAME,
+      dataTypeReference: x.DATA_TYPE_REF,
       discriminatorName: x.DISCRIMINATOR_NAME
     }
   },
--- a/src-electron/db/query-zcl.js
+++ b/src-electron/db/query-zcl.js
@@ -342,7 +342,8 @@ async function selectStructsWithItemsImpl(db, packageIds, clusterId) {
       SI.IS_WRITABLE AS ITEM_IS_WRITABLE,
       SI.IS_NULLABLE AS ITEM_IS_NULLABLE,
       SI.IS_OPTIONAL AS ITEM_IS_OPTIONAL,
-      SI.IS_FABRIC_SENSITIVE AS ITEM_IS_FABRIC_SENSITIVE
+      SI.IS_FABRIC_SENSITIVE AS ITEM_IS_FABRIC_SENSITIVE,
+      SI.DATA_TYPE_REF AS ITEM_DATA_TYPE_REF
     FROM
       STRUCT AS S
     INNER JOIN
@@ -374,7 +375,8 @@ async function selectStructsWithItemsImpl(db, packageIds, clusterId) {
       SI.IS_WRITABLE AS ITEM_IS_WRITABLE,
       SI.IS_NULLABLE AS ITEM_IS_NULLABLE,
       SI.IS_OPTIONAL AS ITEM_IS_OPTIONAL,
-      SI.IS_FABRIC_SENSITIVE AS ITEM_IS_FABRIC_SENSITIVE
+      SI.IS_FABRIC_SENSITIVE AS ITEM_IS_FABRIC_SENSITIVE,
+      SI.DATA_TYPE_REF AS ITEM_DATA_TYPE_REF
     FROM
       STRUCT AS S
     INNER JOIN
@@ -428,7 +430,8 @@ async function selectStructsWithItemsImpl(db, packageIds, clusterId) {
       isWritable: dbApi.fromDbBool(value.ITEM_IS_WRITABLE),
       isNullable: dbApi.fromDbBool(value.ITEM_IS_NULLABLE),
       isOptional: dbApi.fromDbBool(value.ITEM_IS_OPTIONAL),
-      isFabricSensitive: dbApi.fromDbBool(value.ITEM_IS_FABRIC_SENSITIVE)
+      isFabricSensitive: dbApi.fromDbBool(value.ITEM_IS_FABRIC_SENSITIVE),
+      dataTypeReference: value.ITEM_DATA_TYPE_REF
     })
     objectToActOn.itemCnt++
     return acc
@@ -449,8 +452,7 @@ async function selectAllStructItemsById(db, id) {
 SELECT
   STRUCT_ITEM.FIELD_IDENTIFIER,
   STRUCT_ITEM.NAME,
-  (SELECT DATA_TYPE.NAME FROM DATA_TYPE WHERE DATA_TYPE.DATA_TYPE_ID = STRUCT_ITEM.DATA_TYPE_REF) AS TYPE,
-  DATA_TYPE.NAME AS DATA_TYPE_REF_NAME,
+  DATA_TYPE.NAME AS TYPE,
   DISCRIMINATOR.NAME AS DISCRIMINATOR_NAME,
   STRUCT_ITEM.STRUCT_REF,
   STRUCT_ITEM.IS_ARRAY,
@@ -460,7 +462,8 @@ SELECT
   STRUCT_ITEM.IS_WRITABLE,
   STRUCT_ITEM.IS_NULLABLE,
   STRUCT_ITEM.IS_OPTIONAL,
-  STRUCT_ITEM.IS_FABRIC_SENSITIVE
+  STRUCT_ITEM.IS_FABRIC_SENSITIVE,
+  STRUCT_ITEM.DATA_TYPE_REF
 FROM
   STRUCT_ITEM
 INNER JOIN
@@ -498,7 +501,7 @@ async function selectAllStructItemsByStructName(
   let clusterJoinQuery = ''
   let clusterWhereQuery = ''
   if (clusterName) {
-    clusterJoinQuery = ` 
+    clusterJoinQuery = `
     INNER JOIN
       DATA_TYPE_CLUSTER
     ON
@@ -506,11 +509,11 @@ async function selectAllStructItemsByStructName(
     INNER JOIN
       CLUSTER
     ON
-      DATA_TYPE_CLUSTER.CLUSTER_REF = CLUSTER.CLUSTER_ID 
+      DATA_TYPE_CLUSTER.CLUSTER_REF = CLUSTER.CLUSTER_ID
       `
-    clusterWhereQuery = ` 
+    clusterWhereQuery = `
     AND
-      CLUSTER.NAME = "${clusterName}" 
+      CLUSTER.NAME = "${clusterName}"
       `
   }
   return dbApi
@@ -531,7 +534,8 @@ SELECT
   SI.IS_WRITABLE,
   SI.IS_NULLABLE,
   SI.IS_OPTIONAL,
-  SI.IS_FABRIC_SENSITIVE
+  SI.IS_FABRIC_SENSITIVE,
+  SI.DATA_TYPE_REF
 FROM
   STRUCT_ITEM AS SI
 INNER JOIN
--- a/src-electron/generator/helper-zcl.js
+++ b/src-electron/generator/helper-zcl.js
@@ -156,9 +156,6 @@ async function zcl_structs(options) {
         if (i.isArray) {
           st.struct_contains_array = true
         }
-        if (i.isFabricSensitive) {
-          st.struct_has_fabric_sensitive_fields = true
-        }
         if (
           !st.struct_contains_array ||
           !st.struct_has_fabric_sensitive_fields
@@ -176,7 +173,11 @@ async function zcl_structs(options) {
         await Promise.all(promises)
       }
     }
-
+    for (const i of st.items) {
+      if (i.isFabricSensitive) {
+        st.struct_has_fabric_sensitive_fields = true
+      }
+    }
     await checkTransitive(st.items)
   }
   if (checkForDoubleNestedArray) {
--- a/src-electron/generator/helper-zcl.js
+++ b/src-electron/generator/helper-zcl.js
@@ -144,21 +144,36 @@ async function zcl_structs(options) {
     )
   }
   structs = await zclUtil.sortStructsByDependency(structs)
-  structs.forEach((st) => {
+  for (const st of structs) {
     st.struct_contains_array = false
     st.struct_has_fabric_sensitive_fields = false
     st.has_no_clusters = st.struct_cluster_count < 1
     st.has_one_cluster = st.struct_cluster_count == 1
     st.has_more_than_one_cluster = st.struct_cluster_count > 1
-    st.items.forEach((i) => {
-      if (i.isArray) {
-        st.struct_contains_array = true
+    const checkTransitive = async (items) => {
+      for (const i of items) {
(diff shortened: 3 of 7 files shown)


src-electron/generator/helper-zcl.js
```diff
@@ -144,21 +144,36 @@ async function zcl_structs(options) {
...
-      }
-      if (i.isFabricSensitive) {
-        st.struct_has_fabric_sensitive_fields = true
+    const checkTransitive = async (items) => {
+      for (const i of items) {
+        if (i.isArray) {
+          st.struct_contains_array = true
+        }
+        if (i.isFabricSensitive) {
+          st.struct_has_fabric_sensitive_fields = true
```
bzbarsky-apple, line 159: This is wrong, given how this boolean actually gets used.  It really is about the specific fields of this struct, not other things inside it.

In general, having non-toplevel structs with fabric-sensitive fields is not really defined in the spec, and not supported in implementation; if we want to support it we need to figure out how it should work.

src-electron/generator/helper-zcl.js
```diff
@@ -144,21 +144,36 @@ async function zcl_structs(options) {
...
+        }
+        if (
+          !st.struct_contains_array ||
+          !st.struct_has_fabric_sensitive_fields
+        ) {
+          let sis = await queryZcl.selectAllStructItemsByStructName(
```
bzbarsky-apple, line 165: Struct name is not unique; you need to know which cluster is involved too, no?

Furthermore, the same name could be a struct or some other type too (especially across clusters).

bzbarsky-apple (changes requested):

bzbarsky-apple:
This PR and changeset need a _much_ clearer explanation of what is actually being changed and why.

src-electron/generator/helper-zcl.js
gmarcosb: >  having non-toplevel structs with fabric-sensitive fields is not really defined in the spec, and not supported in implementation

If this is the case, then this is not wrong? i.e. the behavior is undefined, with or without this change?

src-electron/generator/helper-zcl.js
gmarcosb: Ouch, thanks for catching that, this was copied from the `checkForDoubleNestedArray` clause below, which apparently has this bug

Do you know if there's already a utility which selects the best match for a struct? i.e. within same cluster if it exists, global otherwise? If there isn't one, I'll write one up

gmarcosb:
> This PR and changeset need a _much_ clearer explanation of what is actually being changed and why.

@bzbarsky-apple I added some more this morning after discussion in zap channel - I wasn't sure how to explain this except to point to the existing compilation fail & why it's failing.

Was there something specifically you wanted me to address?

src-electron/generator/helper-zcl.js
brdandu: selectStructByNameAndClusterId

brdandu (changes requested):

brdandu (changes requested):
Please update one of the xml files and add a unit test for this as well.

gmarcosb:
> Please update one of the xml files and add a unit test for this as well.

I added some structs to `zcl-builtin/matter/data-model/chip/test-cluster.xml`, as `partials/cluster-objects-struct.zapt` seemed to be the only test that referenced `struct_contains_array`

src-electron/generator/helper-zcl.js
gmarcosb: `selectAllStructItemsById` did the trick, but unfortunately there was no way to get the ID of the struct from the `STRUCT_ITEM`, so I had to do some deep surgery to expose that field 🤣

But looks a lot better, thanks for the tip!

src-electron/db/query-zcl.js
```diff
@@ -449,8 +452,7 @@ async function selectAllStructItemsById(db, id) {
 SELECT
   STRUCT_ITEM.FIELD_IDENTIFIER,
   STRUCT_ITEM.NAME,
-  (SELECT DATA_TYPE.NAME FROM DATA_TYPE WHERE DATA_TYPE.DATA_TYPE_ID = STRUCT_ITEM.DATA_TYPE_REF) AS TYPE,
```
gmarcosb, line 452: I made this change which is slightly unrelated & more of a cleanup, but makes my changes clearer

i.e. I don't understand why there was a sub-query here to select `DATA_TYPE.NAME`, when we were already joining with `DATA_TYPE` & selecting `DATA_TYPE.NAME` -> `dataTypeReferenceName` (which was completely unused but would have the same value)

src-electron/db/db-mapping.js
```diff
@@ -378,8 +378,7 @@ exports.map = {
       isNullable: dbApi.fromDbBool(x.IS_NULLABLE),
       isOptional: dbApi.fromDbBool(x.IS_OPTIONAL),
       isFabricSensitive: dbApi.fromDbBool(x.IS_FABRIC_SENSITIVE),
-      dataTypeReference: x.TYPE,
```
brdandu, line 381: Do not change any of the dbmappings because that any existing helper could be using these in the .zapt templates and those will get broken. If you need a new key then just add to an existing dbmap with a different key name from the ones present currently. 
Also, some of the dbmaps which are named after the tables actually should have only the table columns in the dbmap. If you intend to add new keys to the dbmap then it is a good idea to create a new db map with the 'extended' postfix in the name where you can add the columns from the joined tables.

src-electron/db/query-zcl.js
```diff
@@ -449,8 +452,7 @@ async function selectAllStructItemsById(db, id) {
...
-  DATA_TYPE.NAME AS DATA_TYPE_REF_NAME,
```
brdandu, line 453: This could break the templates for backwards compatibility.

zcl-builtin/matter/data-model/chip/test-cluster.xml
```diff
@@ -83,6 +83,10 @@ limitations under the License.
     <item name="a" type="NestedStructList" array="true" optional="false"/>
   </struct>
 
+  <struct name="DeepNestedStructList">
+    <cluster code="0xFFF1FC05"/>
+    <item name="a" type="NestedStructList" optional="false"/>
```
brdandu, line 88: Could you add a test for the helper which will be taking this into account? If zcl_structs then add a generation test for that.

brdandu (changes requested):

src-electron/generator/helper-zcl.js
bzbarsky-apple: Looking for fabric-sensitive bits in non-toplevel structs and assigning semantics to it means that if someone does add something like this, no red flags will go up to actually define it.

We should not change the fabric-sensitive behavior here.

src-electron/generator/helper-zcl.js
```diff
@@ -169,10 +189,9 @@ async function zcl_structs(options) {
       for (const i of st.items) {
         if (i.isArray) {
           // Found an array. Now let's check if it points to a struct that also contains an array.
-          let sis = await queryZcl.selectAllStructItemsByStructName(
+          let sis = await queryZcl.selectAllStructItemsById(
             this.global.db,
-            i.type,
-            packageIds
+            i.dataTypeReference
```
bzbarsky-apple, line 194: But `i` is an array, not a struct, no?  Is the dataTypeReference here actually correct?  Where is it coming from, exactly?  @brdandu please check whether the query changes above handle this correctly.

zcl-builtin/matter/data-model/chip/test-cluster.xml
```diff
@@ -83,6 +83,10 @@ limitations under the License.
     <item name="a" type="NestedStructList" array="true" optional="false"/>
   </struct>
 
+  <struct name="DeepNestedStructList">
```
bzbarsky-apple, line 86: OK, but shouldn't we be testing that the boolean for "contains array" actually gets set for this struct?  Where are we testing that?

src-electron/generator/helper-zcl.js
gmarcosb: OK cool, fair enough, easy enough to exclude fabric-sensitive (though it does seem like either way it'll be broken)

src-electron/db/db-mapping.js
gmarcosb: No helper is using these (I checked), and these fields are deceiving as they are calculating an already existing value

See my other comment: https://github.com/project-chip/zap/pull/1501#discussion_r1902229928

src-electron/db/query-zcl.js
gmarcosb: No templates use these values (& rightfully so, they're junk values, see https://github.com/project-chip/zap/pull/1501#discussion_r1902229928)

zcl-builtin/matter/data-model/chip/test-cluster.xml
gmarcosb: The issue is that I could not find any test which was verifying the functionality of `struct_contains_array` - the best I found was 

My thought would be to modify said test to also check for transitive structures - can you point me to said test?

(see also https://github.com/project-chip/zap/pull/1501#issuecomment-2569920792)

src-electron/generator/helper-zcl.js
gmarcosb: Reverted changes to fabric-sensitive bits

`i` is the result of `selectAllStructsWithItems`, i.e. a `structItem` (see `db-mapping.js`)

The changes here must be correct, because all tests pass - unless the tests are no good 😅🤣

In which case that should be addressed separately

zcl-builtin/matter/data-model/chip/test-cluster.xml
gmarcosb: The real question is: where are we testing `struct_contains_array` behavior

Wherever that test is, it should be expanded to also test that it applies transitively

This is the only reference I found to `struct_contains_array` - see https://github.com/project-chip/zap/pull/1501#issuecomment-2569920792

(these zap tests are incredibly hard to grok)

brdandu (changes requested):

brdandu:
A couple of templates using struct_contains_array that I saw were `cluster-objects-struct.zapt` , `struct.zapt` and `type-by-cluster.zapt` . Another place to add the struct contains array test can be to update type-by-cluster.zapt and geo-meta.test.js => `epc = genResult.content['type-by-cluster.h']`

src-electron/db/db-mapping.js
gmarcosb: I have to push back a bit here, because I think this is precisely what leads to unmaintainable code

Yes, I understand that there can always be unforeseen cases - every code change comes with risk.

But code which makes no sense should fall in one of 2 categories:

1) Dead code which should be removed ASAP
2) Unintuitive code that should be well-tested & have good test coverage

Going the route of "cleaning things up could lead to problems, so let's just continue to build debt" is precisely what leads to unmaintainable code

The worst that will happen here is that some template will break & we'll get a compile-time error, correct? If that happens, we rollback & add the respective test coverage?

That seems like a worthwhile risk to move us towards more maintainable & cleaner code

src-electron/db/query-zcl.js
gmarcosb: My general experience is that lots of good deeds result in moving the needle in a positive direction

Yes, sometimes they lead to bugs in unforeseen cases

But the alternative, letting debt build up because it's scary to clean things up, leads to more issues over time

zcl-builtin/matter/data-model/chip/test-cluster.xml
gmarcosb: Thanks for the pointer! Test added

zcl-builtin/matter/data-model/chip/test-cluster.xml
gmarcosb: `struct_contains_array` had no test coverage, that's why I couldn't find it.   Tests added now to both cover transitive as well as at root

src-electron/db/db-mapping.js
paulr34: That sounds fine to me

paulr34: Here is how it could be broken:

- Matter developer has SDK version x.x.x which they were promised works with the chips they bought.
- SDK version x.x.x was created with ZAP version X which is older than ZAP version Y 
- SDK version x.x.x has external helpers referencing this variable
- If Matter developers updates ZAP to version Y their external helper will not return the correct value

That is a legitimate risk but at the same time I agree with you that we should not be maintaining duplicate data. We would need to determine how relevant that risk really is. 

There is a non zero chance that this would break dev environments and people will be angry.

gmarcosb: If they're running SDK version x.x.x, seems they should stay with ZAP version X

ZAP version & SDK version should be lockstep (this is generally true, which comes back to the argument we brought up before about the issues with the repos being separate)

There doesn't seem to be a valid use-case to update ZAP while not updating the SDK version

If someone does this & are upset, send them my way & I can talk to them about how ZAP & SDK should be updated together, as they are inter-dependent

In this case specifically (as there are no refs to these other 2 fields in SDK), I can also inform them that their current use of these fields is probably not having the desired effect, because they have always had the same value as `type` - that should allow them to do some cleanup as well

paulr34 (approved):

src-electron/db/db-mapping.js
paulr34: Approved

brdandu: `If they're running SDK version x.x.x, seems they should stay with ZAP version X` -> When an issue is logged by a user for zap and we fix it then we do not request the user to update the SDK. It will be better out of box experience for users to just get the latest ZAP. From personal experience, I can say this is generally appreciated by users even though it is more work for the ZAP team.
Also, Please note that we also use ZAP internally for our own SDKs(Zigbee apart from Matter for us). I had to go through every release of ours in the last 2-3 years to see if either of the 2 keys were used anywhere. Hopefully I did not miss a release due to branch naming conventions which can change sometimes. Since I could not find an issue in this case yet, I can approve it but I can assure you backwards compatibility promises are not easy to maintain and it is not always feasible to do these checks. If an issue comes because of this in an old SDK application then it is not easy to backtrack and figure this issue out because it can get pretty crazy to backtrack.
It is my request to only add new keys unless we are really doing an all out cleaning. Even before cleaning up we will need to throw a warning that these dead keys will no longer be supported from the next release.

src-electron/generator/helper-zcl.js
```diff
@@ -144,21 +144,42 @@ async function zcl_structs(options) {
...
+  for (const st of structs) {
     st.struct_contains_array = false
     st.struct_has_fabric_sensitive_fields = false
     st.has_no_clusters = st.struct_cluster_count < 1
     st.has_one_cluster = st.struct_cluster_count == 1
     st.has_more_than_one_cluster = st.struct_cluster_count > 1
-    st.items.forEach((i) => {
-      if (i.isArray) {
-        st.struct_contains_array = true
+    const checkTransitive = async (items) => {
```
bzbarsky-apple, line 153:
```suggestion
    const checkForArraysTransitively = async (items) => {
```

bzbarsky-apple (approved):

src-electron/generator/helper-zcl.js
```diff
@@ -144,21 +144,42 @@ async function zcl_structs(options) {
...
+          st.struct_contains_array = true
+        }
+        if (
+          !st.struct_contains_array ||
+          !st.struct_has_fabric_sensitive_fields
+        ) {
+          promises.push(
+            queryZcl
+              .selectAllStructItemsById(this.global.db, i.dataTypeReference)
+              .then((sis) => {
```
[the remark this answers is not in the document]
gmarcosb: Went with the other one (couldn't remember if one could just pass methods as lambdas in JS 😅)

PR #1503, opened 2025-01-09
appending session uuid for parent dom

paulr34:


src/boot/axios.js
```diff
@@ -26,20 +26,34 @@ let zapUpdateExceptions = (payload, statusCode, message) => {}
...
+let query = new URLSearchParams(search)
+
+// Extract stsApplicationId and trim it
+let stsApplicationId = query.get('stsApplicationId')?.trim()
+
+// Get the current session UUID from sessionStorage
+let currentSessionUuid = window.sessionStorage.getItem('session_uuid')
+
+// If stsApplicationId exists, update the session_uuid
+if (stsApplicationId) {
```
tecimovic, line 47: This if will always be true in Studio, right? But if this is true, then the else if on the line 53 will never even get to be tested.

So if you are in Studio, when do you even create the UUID?

src/boot/axios.js
```diff
@@ -26,20 +26,34 @@ let zapUpdateExceptions = (payload, statusCode, message) => {}
...
+  // Concatenate the current session UUID with the stsApplicationId and store it back
+  window.sessionStorage.setItem(
+    'session_uuid',
+    currentSessionUuid + stsApplicationId
```
tecimovic, line 51: If this is an initial hit, it means currentSessionUuid is null.

src/boot/axios.js
```diff
@@ -26,20 +26,34 @@ let zapUpdateExceptions = (payload, statusCode, message) => {}
...
+  )
+} else if (currentSessionUuid == null) {
+  // If session_uuid doesn't exist, generate a new one
   window.sessionStorage.setItem('session_uuid', uuidv4())
```
tecimovic, line 55: Inside Studio, this line is never reached, so the uuidv4() can never even get executed...

tecimovic (changes requested):
This change makes little sense to me..... If you are in a frame with that applicationId, the UUID now never even gets generated. The stsApplicationId doesn't follow the same uniqueness as that, so that is not enough.

THe  UUID should ALWAYS be created, it's just that with appId, that has to be appended.

src/boot/axios.js
paulr34: fixing

PR #1505, opened 2025-01-14
Deal with session handling.

tecimovic:
This should resolve issues related to session handling.

There are two modes of operation:

1.) WITHOUT a passed appId, in which case we simply use a default value, and map the uuid to that key.
2.) WITH a passed appId (which is called "stsApplicationId" on the query string). In this case, this appId is being used to key the uuid inside the session map.

The second case is used when zap is rendered inside a system with multiple iframes, which share the same "sessionStorage", so you have to kepe session separate. That's why you can't simly use session storage directly.

The map ensures two things:
  1.) When an iframe refreshes, with the same appId, it will reattach to the same sessionId.
  2.) When a whole new iframe is created, and a new appId is passed, the whole new uuid will get created.

### DIFF
--- a/src/util/session-id.js
+++ b/src/util/session-id.js
@@ -1,0 +1,121 @@
+/**
+ *
+ *    Copyright (c) 2025 Silicon Labs
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+import { v4 as uuidv4 } from 'uuid'
+import { parse as querystringParse } from 'querystring'
+
+// Session handling constants
+const SESSION_KEY = 'session_uuid'
+const APP_ID_KEY = 'stsApplicationId'
+const DEFAULT_APP_ID = 'defaultAppId'
+
+/**
+ * Loads a session map from the session storage.
+ * Session map is a string=>string key/value map,
+ * which contains "appId" as a key, and the generated UUID as a value.
+ * If appId is not present (as in standalone), then a default value
+ * is used.
+ *
+ * @param {*} window
+ * @returns
+ */
+function loadSessionMap(window) {
+  let json = window.sessionStorage.getItem(SESSION_KEY)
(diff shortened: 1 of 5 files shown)


paulr34 (approved):

ethanzhouyc (approved):

ethanzhouyc:
For learning purposes: Could you explain why a sessionUuid is needed in addition to sessionId? Where does it function, and what is its relationship with sessionId?

src/util/session-id.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/**
+ * Saves a session map into the session storage.
+ * @param {*} window
+ * @param {*} map
+ */
+function saveSessionMap(window, map) {
+  const mapArray = Array.from(map)
+  const json = JSON.stringify(mapArray)
```
brdandu, line 55: Why convert this to json?

src/util/session-id.js
```diff
@@ -0,0 +1,121 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+import { v4 as uuidv4 } from 'uuid'
+import { parse as querystringParse } from 'querystring'
+
+// Session handling constants
+const SESSION_KEY = 'session_uuid'
+const APP_ID_KEY = 'stsApplicationId'
```
brdandu, line 23: What exactly is an appId?
If a user opens 2 Z3Light applications then do we have 2 appIds for it or the same app Id?
Similarly what exactly is an Iframe in terms of a sample application?

brdandu (commented):
Also is the pre-commit hook not being triggered for you to generate the API documentation with all the comments you added?

Is it possible to add unit tests to cover these changes?

src/util/session-id.js
tecimovic: Because the "sessionStorage" only knows how to save "string => string" object types.
It is not designed to save javascript "Map". So you have to serialize/deserialize it.

tecimovic: I stared at it for a while in confusion, until I realized this to make it work......

src/util/session-id.js
tecimovic: "appId" is really a unique identifier for each "iframe". When an IDE opens multiple copies of ZAP in multiple iframes, each one gets one of those "appId" unique keys.

"iframe" in that context, is basically a separate ".zap" file. If you open 3 .zap files, you will have 3 iframes.

tecimovic:
> For learning purposes: Could you explain why a sessionUuid is needed in addition to sessionId? Where does it function, and what is its relationship with sessionId?

@ethanzhouyc : after this change the "sessionUuid" and the "sessionId" are one and the same thing. Before the change, the session ID was the uuid + appId, but now we're no longer propagating the appId into the back-end, we're merely using it as a key to the map between appId and uuid, and only the uuid travels to the back-end as a "sessionId".

PR #1510, opened 2025-01-21
Display warnings for provisional Matter clusters

ethanzhouyc:
JIRA: ZAPP-1571

- Display a warning icon and message for enabled provisional clusters in Matter, ensuring compatibility with existing cluster warnings.
- Add warnings for enabled provisional clusters to the notification system using SQL triggers.


src-electron/db/zap-schema.sql
brdandu, line 1: Add tests to cover all the triggers.

brdandu (changes requested):

paulr34 (approved):

src/components/ZclDomainClusterView.vue
```diff
@@ -112,15 +112,14 @@ limitations under the License.
 
               <div
                 class="row no-wrap"
-                v-show="isRequiredClusterMissingForId(props.row.id)"
+                v-show="getClusterWarningMessage(props.row)"
```
brdandu, line 115: Doesn't v-show need a boolean?

src/components/ZclDomainClusterView.vue
```diff
@@ -348,19 +351,44 @@ export default {
       if (hasServer) tmp.push('server')
       return tmp
     },
-    isAnyRequiredClusterNotEnabled() {
-      let lackingRequiredCluster = false
+    /* A cluster has a warning if:
+       1. It is required but disabled, OR
+       2. It is provisional and enabled */
+    doesAnyClusterHaveWarning() {
```
brdandu, line 357: isAnyClusterNotCompliant  or areAllClustersCompliant(if statements with no negations are better, easy to read)?

src/components/ZclDomainClusterView.vue
```diff
@@ -348,19 +351,44 @@ export default {
       if (hasServer) tmp.push('server')
       return tmp
     },
-    isAnyRequiredClusterNotEnabled() {
-      let lackingRequiredCluster = false
+    /* A cluster has a warning if:
+       1. It is required but disabled, OR
+       2. It is provisional and enabled */
```
brdandu, line 356: We get a warning when a provisional cluster is enabled? Could you give an example?

test/provisional-cluster.test.js
```diff
@@ -0,0 +1,195 @@
...
+
+    let cluster = await queryZcl.selectClusterByCode(
+      db,
+      pkgId,
+      provisionalClusterCode,
+      null
+    )
+    expect(cluster).not.toBeNull()
+
+    clusterId = cluster.id
```
brdandu, line 86: Since this is needed not as a test but for calculation. Do you think this should go into beforeAll ?

brdandu (changes requested):

src/components/ZclDomainClusterView.vue
ethanzhouyc: You are right. I initially used a non-empty versus empty string as a boolean check, but I’ve now created a function that returns a boolean for better clarity.

src/components/ZclDomainClusterView.vue
ethanzhouyc: Yes, warnings appear only when a provisional cluster is enabled. Showing warnings for all provisional clusters by default would create unnecessary distractions for users.

For example, the Chime Cluster is marked as provisional in its XML (`<cluster apiMaturity="provisional">`). Warning icon with messages and notifications will be displayed if user enables the cluster

test/provisional-cluster.test.js
ethanzhouyc: I think beforeAll() should be limited to db setup and package loading. While cluster insertion is for calculation, future ZAP changes might affect it. Placing it in a separate module with expect() assertions ensures failures are easily traceable.

ethanzhouyc:
@brdandu Improvements in new commits:

1. provisional cluster warnings are set in the notification pane and printed to the console on opening a ZAP file
2. add unit tests for setting provisional cluster warnings in the notification table

I added `<cluster apiMaturity="provisional">` to `scene.xml` for testing purpose. (The cluster is marked as provisional in CHIP repo). The added unit tests require adding a new `.zap` file with enabled provisional cluster.

zcl-builtin/matter/data-model/chip/scene.xml
```diff
@@ -35,7 +35,7 @@ limitations under the License.
   </struct>
 
   <domain name="CHIP"/>
-    <cluster>
+    <cluster apiMaturity="provisional">
```
brdandu, line 38: Do we need a chip repo and alchemy repo issue for CSA to start adding this to our existing XML?

brdandu (approved):

ethanzhouyc:
@brdandu  According to my research, Jest runs tests sequentially within a single test file. Since all async functions use await, the first test will always complete before the second runs. The issue you mentioned would only occur if the variable were shared across different test files, as Jest runs test files in parallel.

In this case, both beforeAll and a separate test would work. I prefer keeping it in a test to allow expect() assertions, making potential setup issues easier to trace.

zcl-builtin/matter/data-model/chip/scene.xml
ethanzhouyc: Only 22 clusters are marked as provisional in the CHIP repo. Given that the ZAP XML is too outdated for Alchemy to apply isolated small changes efficiently, a manual update would be more practical. Would you like me to create a PR for this update?

PR #1512, opened 2025-01-22
deleting unify workflow and scripts related to it

dhchandw:


paulr34 (approved):

ethanzhouyc (approved):

tecimovic:
Is there any remaining usages of the dotdot on the system?

tecimovic (approved):
Fine by me to delete that.
But beware: if there are no more usages of dotdot, and if we have in fact killed the Unify support, then there is a lot more to be deleted. We should clean that up!

PR #1513, opened 2025-01-23
Set element conformance notifications on file opening

ethanzhouyc:
JIRA: ZAPP-1565

- Add element conformance warnings to the notification system when opening a ZAP file.
- Print warnings to the console
- Relocate several functions to improve code structure, will do a clean up PR later for further improvement

### DIFF
--- a/src-electron/db/query-feature.js
+++ b/src-electron/db/query-feature.js
@@ -134,7 +134,7 @@ function evaluateConformanceExpression(expression, elementMap) {
    */
   function evaluateBooleanExpression(expr) {
     // Replace terms with their actual values from elementMap
-    expr = expr.replace(/[A-Za-z][A-Za-z0-9]*/g, (term) => {
+    expr = expr.replace(/[A-Za-z][A-Za-z0-9_]*/g, (term) => {
       if (elementMap[term]) {
         return 'true'
       } else {
@@ -167,7 +167,7 @@ function evaluateConformanceExpression(expression, elementMap) {
   let parts = expression.split(',')
   // if any term is desc, the conformance is too complex to parse
   for (let part of parts) {
-    let terms = part.match(/[A-Za-z][A-Za-z0-9]*/g)
+    let terms = part.match(/[A-Za-z][A-Za-z0-9_]*/g)
     if (terms && terms.includes('desc')) {
       return 'desc'
     }
@@ -214,7 +214,7 @@ function evaluateConformanceExpression(expression, elementMap) {
  * @returns all missing terms in an array
  */
 function checkMissingTerms(expression, elementMap) {
-  let terms = expression.match(/[A-Za-z][A-Za-z0-9]*/g)
+  let terms = expression.match(/[A-Za-z][A-Za-z0-9_]*/g)
   let missingTerms = []
   let abbreviations = ['M', 'O', 'P', 'D', 'X']
   for (let term of terms) {
@@ -234,7 +234,7 @@ function checkMissingTerms(expression, elementMap) {
  */
 function filterElementsContainingDesc(elements) {
   return elements.filter((element) => {
-    let terms = element.conformance.match(/[A-Za-z][A-Za-z0-9]*/g)
+    let terms = element.conformance.match(/[A-Za-z][A-Za-z0-9_]*/g)
     return terms && terms.includes('desc')
   })
 }
@@ -248,7 +248,7 @@ function filterElementsContainingDesc(elements) {
  */
 function filterRelatedDescElements(elements, featureCode) {
   return elements.filter((element) => {
-    let terms = element.conformance.match(/[A-Za-z][A-Za-z0-9]*/g)
+    let terms = element.conformance.match(/[A-Za-z][A-Za-z0-9_]*/g)
     return terms && terms.includes('desc') && terms.includes(featureCode)
   })
 }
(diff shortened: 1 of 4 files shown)


src-electron/db/query-attribute.js
```diff
@@ -1248,53 +1248,6 @@ async function selectAttributeMappingsByPackageIds(db, packageIds) {
...
-/**
- * Get all attributes in an endpoint type cluster
- * Disabled attributes are not loaded into ENDPOINT_TYPE_ATTRIBUTE table
- * when opening a ZAP file, so we need to join DEVICE_TYPE_CLUSTER table
- * @param {*} db
- * @param {*} endpointTyeClusterId
- * @param {*} deviceTypeClusterId
- * @returns all attributes in an endpoint type cluster
- */
-async function selectAttributesByEndpointTypeClusterIdAndDeviceTypeClusterId(
```
paulr34, line 1260: Could you please provide an explanation for moving these functions

ethanzhouyc: The `getEndpointTypeElements` function was moved from `user-data.js` because it is unrelated to API calls. Since it queries the attribute, command, and event tables, I could not find a perfect location for it, and `query-zcl.js` seemed to be the most suitable location. The 3 functions called inside `getEndpointTypeElements` were moved to align with it.

I could revert this if you think the location does not fit.

paulr34: I'd say before we merge we should figure out and define for good the difference between query-attribute/event/command etc and query-zcl and clean up accordingly. 

I think what you're doing is a good thing and I think we should take it further and fully flush it out.

What do you think?

ethanzhouyc: I agree. Let's discuss in detail tomorrow

paulr34 (approved):
Approving because we agreed to create a follow up PR to cleanup the files

src-electron/db/query-attribute.js
ethanzhouyc: Will do a cleanup PR that relocates `getEndpointTypeElements` along with other conformance-related functions to a new folder.

brdandu: We are extracting attribute specific details here so it does make sense to keep this in this in the query-attribute.js file. Same for events and commands

src-electron/importexport/import-json.js
```diff
@@ -2057,6 +2068,126 @@ async function jsonDataLoader(
...
+          sessionId,
+          1,
+          0
+        )
+      }
+      return true
+    }
+  }
+  return false
+}
```
brdandu, line 2189: Are these warnings which are populated when the .zap file is opened? Should these also go into the console so that the CSA folks see these warnings when they generate their sample applications?

src-electron/rest/user-data.js
```diff
@@ -98,39 +98,6 @@ function httpGetDeviceTypeFeatures(db) {
...
 
-/**
- * Get all attributes, commands and events in an endpoint type cluster.
- * @param {*} db
- * @param {*} endpointTypeClusterId
- * @param {*} deviceTypeClusterId
- * @returns elements object containing all attributes, commands and events
- * in an endpoint type cluster
- */
-async function getEndpointTypeElements(
```
brdandu, line 109: Was this function never used?

test/query.test.js
brdandu, line 1: Add tests corresponding to your new changes

brdandu (changes requested):

src-electron/db/query-feature.js
```diff
@@ -134,7 +134,7 @@ function evaluateConformanceExpression(expression, elementMap) {
    */
   function evaluateBooleanExpression(expr) {
     // Replace terms with their actual values from elementMap
-    expr = expr.replace(/[A-Za-z][A-Za-z0-9]*/g, (term) => {
+    expr = expr.replace(/[A-Za-z][A-Za-z0-9_]*/g, (term) => {
```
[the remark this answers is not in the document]
ethanzhouyc: The conformance expression could contain '_', for example: `AB_CD & EF`. This change will make sure to capture `AB_CD` as a single term

src-electron/importexport/import-json.js
ethanzhouyc: Yes, will add to warnings to the console too

src-electron/db/query-feature.js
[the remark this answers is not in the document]
brdandu: ok does '_' signify anything in terms of the logical expression here like the &? Didn't see this part of the expression in the spec before so definitely do not understand it. Could you point me to the spec where this is used?

test/query.test.js
ethanzhouyc: Unit tests for "conformance warnings on import" require adding attribute/command/event conformance to related XMLs.
We could:
1. add these tests after aligning ZAP XMLs with CHIP XMLs.
OR
2. manually updates 1-2 related ZAP XMLs

I prefer adding the tests later

src-electron/db/query-feature.js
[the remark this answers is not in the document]
ethanzhouyc: The '_' does not function as a logical operator like & (AND) or | (OR). Instead, it is simply a part of certain term name.
For example: `PA_LF` is a feature from the spec

brdandu (approved):
Create an issue where this issue is linked and you mentioned the remaining items that need to be done before closing this.
Tests for import and for notifications pane both.

ethanzhouyc:
Issue #1515 and #1516 created and will be resolved later.

PR #1520, opened 2025-02-06
return GUI data for selected endpoint and not just the first one in the list

paulr34:


brdandu (changes requested):
Add unit tests for the issue this change is solving.

src-electron/rest/user-data.js
```diff
@@ -438,7 +439,7 @@ function httpPostAttributeUpdate(db) {
     // send latest value to frontend to update UI
     let eptAttr = await queryZcl.selectEndpointTypeAttribute(
       db,
-      endpointTypeIdList[0],
+      selectedEndpoint,
       id,
       clusterRef
     )
```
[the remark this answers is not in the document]
paulr34: What if I enable a cluster in endpoint 3 and that cluster is not enabled in endpoint 1?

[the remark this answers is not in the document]
paulr34: It returns that the attributes are INCLUDED=0 since it's off in endpoint 1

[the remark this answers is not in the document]
brdandu: Don't know what I was saying is does this code also need a change in `endpointTypeIdList[0],`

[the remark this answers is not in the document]
paulr34: I see, good catch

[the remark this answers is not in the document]
paulr34: fixed

cypress/e2e/attribute_validations/reporting.cy.js
```diff
@@ -0,0 +1,68 @@
...
+describe('Add multiple clusters and search', () => {
+  // This will hold the fixture data and the flag for skipping the tests
+  let skipTest = false
+
+  before(function () {
+    // Load the fixture data before running any tests
+    cy.fixture('data').then((data) => {
+      this.data = data
+      // If mode is "matter", set skipTest flag to true
+      if (this.data.mode === 'matter') {
```
brdandu, line 17: Better to run it for zigbee than not matter. If a new protocol does not have attribute reporting then this test will fail.

cypress/e2e/attribute_validations/reporting.cy.js
```diff
@@ -0,0 +1,68 @@
...
+
+    // Continue with the test if the mode is not "matter"
+    cy.fixture('baseurl').then((data) => {
+      cy.visit(data.baseurl)
+    })
+    cy.setZclProperties()
+
+    // Create and validate the first endpoint
+    cy.fixture('data').then((data) => {
+      cy.addEndpoint(data.endpoint4, data.cluster1)
```
brdandu, line 38: I do not understand this as I am not familiar with what is happening here. Comment says first endpoint and `cy.addEndpoint(data.endpoint4, data.cluster1)` says data.endpoint4

cypress/e2e/attribute_validations/reporting.cy.js
```diff
@@ -0,0 +1,68 @@
...
+    })
+    cy.get('.flex  > strong').should('contain', '#1')
+
+    // Select the "Power Configuration" cluster for the second endpoint
```
brdandu, line 42: is another `// Create and validate the second endpoint needed here?`
Just like the addition of the first endpoint.

cypress/e2e/attribute_validations/reporting.cy.js
```diff
@@ -0,0 +1,68 @@
...
+        timeout: 10000
+      })
+        .should('be.visible')
+        .click()
+    })
+    cy.get(
+      '.q-virtual-scroll__content > :nth-child(1) > :nth-child(2) > .q-toggle > .q-toggle__inner > .q-toggle__thumb'
+    ).click()
+    // Check to make sure the button is still visible after clicking
+    cy.get(
```
brdandu, line 64: I hope this test does not just get the next item in the list.

brdandu (approved):
I do not understand cypress that well so take a look at the comments and take appropriate actions.

cypress/e2e/attribute_validations/reporting.cy.js
paulr34: Sure good idea

cypress/e2e/attribute_validations/reporting.cy.js
paulr34: Yeah it's just the method of adding endpoints. It should be cleaned up and I'll make a follow up PR to fix cypress as a whole

cypress/e2e/attribute_validations/reporting.cy.js
paulr34: I'm sorry I'm confused what you mean here

brdandu: I just don't really know how this checks in the UI when an attribute vanishes. If there are 2 attributes in the attribute reporting and one of them vanishes then does the second one just show up in the first one's place and pass this test?
Is it better to test by checking the number of attributes in attribute reporting before and after enablement?

brdandu (approved):

cypress/e2e/attribute_validations/reporting.cy.js
paulr34: I see, it gets a specific one

paulr34: It doesn't just get the first one in the list but I will triple check

paulr34:
> In the following PR to clean up Cypress test, we should rename these endpoints with their actual names. These random numbers are so confusing to read
> 
> ```
> 
> "endpoint1": "Billing Unit (0x0203)",
> 
>  "endpoint2": "CBA Config Tool (0x0005)",
> 
>  "endpoint3": "Chatting Station (0x0601)",
> 
>  "endpoint4": "CBA Thermostat (0x0301)",
> 
> ```

I agree. Good idea. I'll whip something up for next week

cypress/e2e/attribute_validations/reporting.cy.js
paulr34: done, thank you

cypress/e2e/attribute_validations/reporting.cy.js
paulr34: yes, this is just how cypress is setup for all tests but I agree it should be fixed and cleaned up. I will work on a follow up PR to fix Cypress in general

Issue #1413, opened 2024-08-27
Incorrect handling of disallowConform in cluster specifications.


bzbarsky-apple:
Steps to reproduce:

1.  Add the following to `src/app/zap-templates/zcl/data-model/chip/rvc-run-mode-cluster.xml` in Matter:
    ```
    <features>
      <!-- TODO: ZAP still generates feature bits for things with disallowConform! -->
      <feature bit="0" code="DEPONOFF" name="OnOff" summary="Dependency with the OnOff cluster is disallowed for this cluster">
        <disallowConform/>
      </feature>
      <feature bit="16" code="DIRECTMODECH" name="DirectModeChange" summary="Cluster supports changing run modes from non-Idle states">
        <optionalConform/>
      </feature>
    </features>
    ```
3. Remove the `Feature` bitmap from that cluster.
4. Rerun codegen.

This generates a feature bit for the OnOff feature, even though it's explicitly listed as not allowed for this cluster.

PR #1524, opened 2025-02-12
Exclude Features with Disallowed Conformance from Bitmaps

ethanzhouyc:
- Add features to the Feature Bitmap only if their conformance is not disallowed.
- Avoid empty Feature Bitmap by removing them entirely if all contained features have disallowed conformance. This change assumes an empty Feature Bitmap should be removed

- Issue: #1413 
- JIRA: GHM_ZAP-449

```
      <!-- TODO: ZAP still generates feature bits for things with disallowConform! -->
      <feature bit="0" code="DEPONOFF" name="OnOff" summary="Dependency with the OnOff cluster is disallowed for this cluster">
        <disallowConform/>
      </feature>
```
With this PR, the OnOff feature will no longer be added to bitmaps in codegen when the above XML configuration is applied.


ethanzhouyc:
This PR fails CI because it removes disallowed features from bitmaps in code generation, requiring the CHIP repository to be updated by re-running Matter codegen.

ethanzhouyc:
Closing this PR as the new plan is to add warnings for disallowed features instead of removing them from bitmaps

PR #1528, opened 2025-02-19
Bug fix gui data for feature page

paulr34:


ethanzhouyc (approved):
Tested on feature page and the change fixed the bug

brdandu (approved):
Approving. Please log an issue for follow up cypress tests to cover this and further code cleanup here.

Issue #1474, opened 2024-10-28
Defining a list-typed thing in ZAP XML is inconsistent


bzbarsky-apple:
For attributes you have to do:
```
type="array" entryType="int8u"
```
to represent the `list[uint8]` type from the spec.

For event fields, struct fields, command fields, you have to do:
```
array="true" type="int8u"
```
to represent the `list[uint8]` type from the spec.

ethanzhouyc:
Issue fixed in #1530

PR #1530, opened 2025-02-20
Adapt ZAP to Support Both List Type Definition Formats for Attributes

ethanzhouyc:
- This PR adapts ZAP to support parsing list type definition `array="true" type="X"` for attributes, ensuring consistency with command and event definitions in the XML.
- ZAP now converts `array="true" type="X"` to `type="array" entryType="X"` before parsing, allowing support for both formats.
- This update prepares ZAP for upcoming XML changes to standardize list type definitions across elements
- Issue: #1474
- Jira: GHM_ZAP-461

### DIFF
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -630,6 +630,12 @@ function prepareCluster(cluster, context, isExtension = false) {
           attribute.$.reportingPolicy
         )
       }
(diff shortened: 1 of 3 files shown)


paulr34 (approved):

dhchandw:
Change an attribute in the test metadata to use the new format `array="true" type="X"` and a test to verify that the attribute was loaded in correctly.

dhchandw (approved):

ethanzhouyc:
Unit test added

PR #1532, opened 2025-02-25
Adding clusterExtension Support from standard xml file loading like we do from zcl.json

brdandu:
- Updated processClusterExtensions which is part of the delayed promises to load custom data types within clusterExtensions tag.
- Now users can define custom data types within clusterExtensions referring to other data types within standard xml.
- Added Unit tests for the same
- Github: ZAP#1529

### DIFF
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -1968,14 +1968,14 @@ async function processSubItems(
   context,
   collectedStructItems
 ) {
-  let subItemPrmosies = []
+  let subItemPromises = []
   if (dbEnum.zclType.enum in toplevel) {
-    subItemPrmosies.push(
+    subItemPromises.push(
       processEnumItems(db, filePath, packageId, knownPackages, toplevel.enum)
     )
   }
   if (dbEnum.zclType.bitmap in toplevel) {
-    subItemPrmosies.push(
+    subItemPromises.push(
       processBitmapFields(
         db,
         filePath,
@@ -1988,12 +1988,12 @@ async function processSubItems(
   // Treating features in a cluster as a bitmap
   if (featureClusters && featureClusters.length > 0) {
     featureClusters.forEach((fc) => {
-      subItemPrmosies.push(
+      subItemPromises.push(
         processBitmapFields(db, filePath, packageId, knownPackages, fc)
       )
     })
   }
-  // Delaying the loading of struct items into collecedStructItems instead of
+  // Delaying the loading of struct items into collectedStructItems instead of
   // processing them with other subitems. This is because the struct items are
   // dependent on types which could span across multiple xml files.
   if (dbEnum.zclType.struct in toplevel) {
@@ -2005,7 +2005,7 @@ async function processSubItems(
       context
     ])
   }
-  return Promise.all(subItemPrmosies)
+  return Promise.all(subItemPromises)
 }
 
 /**
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -823,6 +823,8 @@ async function processClusterExtensions(
   env.logDebug(
     `${filePath}, ${dataPackageId}: ${data.length} cluster extensions.`
   )
+
+  // Insert cluster extensions
   return queryLoader.insertClusterExtensions(
     db,
     dataPackageId,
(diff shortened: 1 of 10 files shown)


src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -823,6 +823,62 @@ async function processClusterExtensions(
...
+  // </clusterExtension>
+  for (const clusterExtension of data) {
+    // Process data types within cluster extensions
+    await processDataTypes(
+      db,
+      filePath,
+      dataPackageId,
+      knownPackages,
+      clusterExtension,
+      []
```
paulr34, line 847: maybe pass in a variable like "featureClusters = []" so its easier to read?

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1751,6 +1807,252 @@ async function processDeviceTypes(db, filePath, packageId, data, context) {
...
+ * @param {*} context
+ * @returns Promise of processing sub items within a bitmap, enum and structs.
+ */
+async function processSubItems(
+  db,
+  filePath,
+  packageId,
+  knownPackages,
+  toplevel,
+  featureClusters,
```
paulr34, line 2020: alternatively, maybe set "featureClusters = []" by default in the function definition

paulr34 (approved):
Looks good

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1831,187 +2135,40 @@ async function processParsedZclData(
...
-    await Promise.all(batch3)
+    // Batch 3: Load the data type table which lists all data types
+    await processDataTypes(
+      db,
+      filePath,
+      packageId,
+      knownPackages,
+      toplevel,
+      featureClusters
+    )
```
ethanzhouyc, line 2148: Before, all batches were pushed to the same promise and ran in parallel. Why we are turning to use await to run sequentially? Will this affect the performance?

ethanzhouyc (approved):

src-electron/zcl/zcl-loader-silabs.js
ethanzhouyc: Is there any reason we left batch1 and batch2 still in parallel promises and not create separate async functions for them?

brdandu: The batches are still in parallel just made them go into separate function to avoid code redundancy. Which part are you referring to?

test/resource/custom-cluster/custom-xml-in-zcl-json.xml
```diff
@@ -0,0 +1,42 @@
...
+
+  <!--
+    These are test global items (no cluster attached) for testing only.
+    Their usage is defined for UnitTestCluster only.
+  -->
+  <clusterExtension code="0x0201">
+    <struct name="AdditionalInfoStruct">
+      <cluster code="0x0201"/>
+      <item name="SystemMode" type="SystemModeEnum"/>
+    </struct>
```
ethanzhouyc, line 29: The CI failed as it expects only attributes and commands can exist under clusterExtension?

src-electron/zcl/zcl-loader-silabs.js
brdandu: await Promise.all(batch3) became await processDataTypes. Content within is still parallel

ethanzhouyc: I see. You are right!! I thought it was parallel across batches originally but it is not

test/resource/custom-cluster/custom-xml-in-zcl-json.xml
brdandu: Correct fixed that in the xsd. Will need to ask Timotej on how we maintain that file.

test/resource/custom-cluster/custom-xml-in-zcl-json.xml
```diff
@@ -0,0 +1,42 @@
...
+
+<configurator>
+  <domain name="CHIP"/>
+
+  <!--
+    These are test global items (no cluster attached) for testing only.
+    Their usage is defined for UnitTestCluster only.
+  -->
+  <clusterExtension code="0x0201">
+    <struct name="AdditionalInfoStruct">
```
bzbarsky-apple, line 26: We should make this work for structs at toplevel, not inside `<clusterExtension>`, for two reasons:

1. Structs can be referenced by more than one cluster and hence more than one clusterExtension.  Placing the struct inside one of them would then potentially break the other, depending on processing order.
2. Even when not doing clusterExtension and just defining data types, shouldn't we support a struct in file A referencing a data type defined in file B, just like we support a command or an event or an attribute in file A referencing a type defined in file B?

brdandu: Data types need not be within the clusterExtensions.

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1751,19 +1753,278 @@ async function processDeviceTypes(db, filePath, packageId, data, context) {
...
+  db,
+  filePath,
+  packageId,
+  knownPackages,
+  toplevel,
+  featureClusters,
+  context,
+  collectedStructItems
+) {
+  let subItemPrmosies = []
```
bzbarsky-apple, line 1971: Spelling?

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1751,19 +1753,278 @@ async function processDeviceTypes(db, filePath, packageId, data, context) {
...
+  }
+  // Treating features in a cluster as a bitmap
+  if (featureClusters && featureClusters.length > 0) {
+    featureClusters.forEach((fc) => {
+      subItemPrmosies.push(
+        processBitmapFields(db, filePath, packageId, knownPackages, fc)
+      )
+    })
+  }
+  // Delaying the loading of struct items into collecedStructItems instead of
```
bzbarsky-apple, line 1996: What is collecedStructItems?

PR #1537, opened 2025-03-11
Fix: dirty state incorrectly being set to 1 upon .zap file open

dhchandw:
- dirty state being incorrectly set to 1 upon .zap file open when it has 2 or more custom xml packages
- consistent package ordering in .zap file

ZAPP-1462


src-electron/db/query-impexp.js
```diff
@@ -383,7 +383,9 @@ INNER JOIN
   SESSION_PARTITION
 ON
   SESSION_PACKAGE.SESSION_PARTITION_REF= SESSION_PARTITION.SESSION_PARTITION_ID
-WHERE SESSION_PARTITION.SESSION_REF = ? AND SESSION_PACKAGE.ENABLED = 1`,
+WHERE SESSION_PARTITION.SESSION_REF = ? AND SESSION_PACKAGE.ENABLED = 1
```
dhchandw, line 386: This is the only non formatting change in this file

src-electron/db/query-impexp.js
```diff
@@ -383,7 +383,9 @@ INNER JOIN
...
+ORDER BY
+  PACKAGE.PACKAGE_ID`,
```
dhchandw, line 388: This is the only non-formatting change in this file

paulr34:
Can you explain this change to me? Did some sort of ZAP file update/upgrade change the .zap file?

dhchandw:
> Can you explain this change to me? Did some sort of ZAP file update/upgrade change the .zap file?

Fixed the following:
- A verification step we did in the `import.js` file was incorrectly setting the dirty flag to 1 on file open when the `.zap` file had 2 or more Custom XMLs.
- The ordering of packages in `.zap` file were inconsistent when there were 2 or more Custom XMLs (this was fixed by ordering the query)

paulr34 (approved):

src-electron/db/query-impexp.js
```diff
@@ -383,7 +383,15 @@ INNER JOIN
   SESSION_PARTITION
 ON
   SESSION_PACKAGE.SESSION_PARTITION_REF= SESSION_PARTITION.SESSION_PARTITION_ID
-WHERE SESSION_PARTITION.SESSION_REF = ? AND SESSION_PACKAGE.ENABLED = 1`,
+WHERE SESSION_PARTITION.SESSION_REF = ? AND SESSION_PACKAGE.ENABLED = 1
+ORDER BY
+  CASE PACKAGE.TYPE
+    WHEN 'zcl-properties' THEN 1
```
brdandu, line 389: Use dbEnum.packageType

brdandu: This could cause all existing .zap files to update?

test/importexport.test.js
```diff
@@ -523,3 +527,29 @@ test(
...
+      },
+      null,
+      [templatePkgId]
+    )
+    await importJs.importDataFromFile(db, threeCustomXml, {
+      sessionId: sid
+    })
+    let state = await exportJs.createStateFromDatabase(db, sid)
+
+    expect(state.package[2].path).toBe('matter-custom.xml')
```
brdandu, line 550: package[0] and package[1]?
Worth adding a multi-protocol test here where there are 2 templates.json and 2 zcl.json

brdandu (changes requested):

src-electron/db/query-impexp.js
dhchandw: Yes it is possible, but it will only effect files that have 2 or more custom XMLs

dhchandw: As of now whether the zcl-properties file or gen-template-json file is first in the .zap file is random, so this can effect:
.zap files that have gen-templates-json before zcl-properties
.zap files that have 2 or more custom xmls

dhchandw: As of now whether the zcl-properties file or gen-template-json file is first in the .zap file is random, so this can effect:
.zap files that have gen-templates-json before zcl-properties
.zap files that have 2 or more custom xmls

brdandu (approved):

PR #1538, opened 2025-03-11
Allow Features to have cluster codes just like bitmaps and other data types

brdandu:
- Loading the data type and bitmap tables by cluster codes in features
- Adding tests to check that Feature bitmap and bitmap field entries are loaded per cluster mentioned
- Github: ZAP#1306

### DIFF
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -1526,15 +1526,33 @@ async function processBitmapFields(
     data.features.length == 1 &&
     'feature' in data.features[0]
   ) {
-    data.features[0].feature.forEach((item) => {
-      bitmapFields.push({
-        bitmapName: 'Feature',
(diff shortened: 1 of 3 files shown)


tecimovic (approved):

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1526,15 +1526,29 @@ async function processBitmapFields(
...
       })
-    })
+    }
+    clusterCodes.forEach((clCode) =>
+      data.features[0].feature.forEach((item) => {
+        bitmapFields.push({
+          bitmapName: 'Feature',
+          bitmapClusterCode: [clCode],
+          name: item.$.name,
+          mask: 1 << item.$.bit,
```
tecimovic, line 1547: I wonder if this blows up with a very weird error if "bit" in XML is not a nice number.... Might be better to extract "bit" out first, and if it's not a number, throw some clearly specified error, otherwise people might be confused.

PR #1539, opened 2025-03-12
Timeout generation with an error if the template generation is taking forever

brdandu:
- In template-engine.js, produceContent function, timing out the handlebars template handler if it takes more than a certain amount of time for generation because this can lead to ZAP just hanging forever.
- Also throwing errors when there is no content generated out of a generation template file
- Github: ZAP#792

### DIFF
--- a/src-electron/generator/template-engine.js
+++ b/src-electron/generator/template-engine.js
@@ -191,7 +191,36 @@ async function produceContent(
   if (options.initialContext != null) {
     Object.assign(context, options.initialContext)
   }
-  let content = await template(context)
+  let content
+  // Render the template but if it does not render within 30s, then throw an
+  // error instead of just hanging forever.
+  try {
+    // Attempt to render the template
+    content = await Promise.race([
+      template(context),
+      new Promise((_, reject) =>
+        setTimeout(
+          () => reject(new Error('Template rendering timed out')),
+          30000
+        )
+      )
+    ])
+  } catch (error) {
+    // Log the error and throw it
+    notification.setNotification(
+      db,
+      'ERROR',
+      `Error during template rendering of ${singleTemplatePkg.path}: ` +
+        error.message,
+      sessionId,
+      1
+    )
+    console.error(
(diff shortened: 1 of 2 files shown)


src-electron/generator/template-engine.js
```diff
@@ -185,13 +187,47 @@ async function produceContent(
...
+      stats: {},
+      deferredBlocks: []
     }
   }
   if (options.initialContext != null) {
     Object.assign(context, options.initialContext)
   }
-  let content = await template(context)
+  let content
+  // Render the template but if it does not render within 30s, then throw an
```
dhchandw, line 198: Update comment to 90 seconds? or is this a different timeout?

tecimovic (approved):

ethanzhouyc (approved):

test/gen-template/matter/simple-test.zapt
```diff
@@ -42,3 +42,18 @@ Base type for {{name}} : {{baseType}}
...
+{{#if (zcl_command_arguments_count ../index)}}
+zcl command arguments exist for {{../name}} command. It has {{zcl_command_arguments_count ../index}} arguments
+{{else}}
+zcl command arguments do not exist for {{../name}} command.
+{{/if}}
+{{/zcl_command_arguments}}
+{{/zcl_commands_source_server}}
+{{/zcl_clusters}}
+
+
```
ethanzhouyc, line 58: Only 1 empty line should be added here?

src-electron/generator/template-engine.js
```diff
@@ -185,13 +187,48 @@ async function produceContent(
...
+        )
+      )
+    ])
+    // Render deferred blocks
+    for (let deferredBlock of context.global.deferredBlocks) {
+      content += await deferredBlock(context)
+    }
+  } catch (error) {
+    // Log the error and throw it
+    notification.setNotification(
```
ethanzhouyc, line 218: Should we add await here to make sure notifications are set before throwing error?

PR #1541, opened 2025-03-19
Allow user to edit a device type on an existing endpoint as a new endpoint config and not an overwrite on existing endpoint config

brdandu:
- Currently when a user edits the device type on an endpoint configuration then the configuration is overwritten on top of the existing endpoint configuration.
- However the above can lead to non-compliant device type as per the protocol specification if the users are not careful.
- With this change a dialogue pops up when the user changes the device type to either edit on top of existing endpoint config or add as a new endpoint config.
- Added cypress tests to cover the same
- Github: ZAP#1125

### DIFF
--- a/src/components/ZclCreateModifyEndpoint.vue
+++ b/src/components/ZclCreateModifyEndpoint.vue
@@ -262,6 +262,43 @@ limitations under the License.
         />
       </q-card-actions>
     </q-card>
+
+    <!-- Device Type Change Dialog -->
+    <q-dialog v-model="showDeviceTypeChangeDialog" persistent>
+      <q-card>
+        <q-card-section>
+          <div class="text-h6">Device Type Changed</div>
+          <p>
+            The device type configuration for this endpoint has changed. Would
+            you like to apply the changes on top of the existing configuration,
+            which may result in a noncompliant device type (Overwrite), or
+            delete the existing endpoint and create a new one?
+          </p>
+        </q-card-section>
+        <q-card-actions align="right">
+          <q-btn
+            flat
+            label="Cancel"
+            color="primary"
+            @click="cancelDeviceTypeChange"
+          />
+          <q-btn
+            flat
+            label="Overwrite"
+            color="primary"
+            @click="overwriteDeviceType"
+            data-test="overwrite"
+          />
+          <q-btn
+            flat
+            label="Delete and Add"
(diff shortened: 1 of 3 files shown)


src/components/ZclCreateModifyEndpoint.vue
```diff
@@ -588,15 +626,58 @@ export default {
...
+          const currentDeviceType = this.deviceTypeTmp.map(
+            (dt) => dt.deviceTypeRef
+          )
+          const previousDeviceType =
+            this.endpointDeviceTypeRef[
+              this.endpointType[this.endpointReference]
+            ]
+
+          if (currentDeviceType.toString() !== previousDeviceType.toString()) {
+            // Show the dialog
```
ethanzhouyc, line 640: When comparing arrays with toString(), [a, b] and [b, a] will be different. Will this cause any problem?

src/components/ZclCreateModifyEndpoint.vue
```diff
@@ -341,7 +377,9 @@ export default {
       deviceTypeTmp: [], // Temp store for the selected device types
       primaryDeviceTypeTmp: null, // Temp store for the selected primary device type
       endpointIds: [],
-      tmpSelectedOptions: []
+      tmpSelectedOptions: [],
+      showDeviceTypeChangeDialog: false, // Controls the visibility of the dialog
+      pendingAction: null // Stores the action to perform after dialog confirmation
```
ethanzhouyc, line 382: Is the variable `pendingAction` used in anywhere?

ethanzhouyc (approved):

src/components/ZclCreateModifyEndpoint.vue
brdandu: We should change if the order changes because the primary device type gets changed when the order is changed.

src/components/ZclCreateModifyEndpoint.vue
brdandu: Good catch. No, will fix this.

paulr34 (approved):

paulr34:
@brdandu what about Zigbee?

brdandu:
> @brdandu what about Zigbee?

Zigbee will have the same warning when user changes the device type. Both can benefit with this.

PR #1546, opened 2025-03-25
Only Zigbee can't have min/max for attributes larger than 2 bytes

paulr34:


src-electron/generator/helper-endpointconfig.js
```diff
@@ -589,15 +589,16 @@ function endpoint_attribute_min_max_count(options) {
  * @returns All attributes with min max values listed
  */
 function endpoint_attribute_min_max_list(options) {
-  let comment = null
-  let order = options.hash.order
-  if (order == null || order.length == 0) {
-    order = 'def,min,max'
-  }
+  let comment = null;
+  let order = options.hash?.order || 'def,min,max'; // Use optional chaining and default value
```
brdandu, line 593: What if `options.hash?.order` is an empty string? I would keep the old code.

src-electron/generator/helper-endpointconfig.js
```diff
@@ -589,15 +589,16 @@ function endpoint_attribute_min_max_count(options) {
...
+  let ret = '{ \\\n';
+
+  // Extract the category for better readability and safety
+  const category = options.data?.root?.global?.genTemplatePackage?.category
 
-  let ret = '{ \\\n'
   this.minMaxList.forEach((mm, index) => {
-    if (mm.typeSize > 2) {
+    // Check if the category is 'zigbee' and handle safely
+    if (mm.typeSize > 2 && category === 'zigbee') {
```
brdandu, line 601: Change `category === 'zigbee'` to `category !== 'matter'` . If there is no category which is the case in the old GSDK releases then it is better to still have that show up than vanish.

brdandu (changes requested):

src-electron/generator/helper-endpointconfig.js
paulr34: sure, fixing

src-electron/generator/helper-endpointconfig.js
paulr34: will do thanks

ethanzhouyc (approved):

brdandu (approved):

PR #1547, opened 2025-03-25
Fix: Handling package paths in different drives than user home dir or zap file location

dhchandw:
Relative paths don't exist across two different drives. On file import if package is in different drive than home dir or zap file location, package relativity will be changed to absolute for that package.

ZAPP-1087

### DIFF
--- a/src-electron/importexport/export.js
+++ b/src-electron/importexport/export.js
@@ -136,6 +136,10 @@ async function exportSessionPackages(db, sessionId, zapProjectFileLocation) {
         pathRelativity = dbEnum.pathRelativity.relativeToZap
       }
     }
+    if (path.isAbsolute(relativePath) && /^[A-Z]:\\/.test(relativePath)) {
+      // Handling Windows path when package is on different drive than zapProjectFile or user home
+      pathRelativity = dbEnum.pathRelativity.absolute
+    }
     let ret = {
       pathRelativity: pathRelativity,
       path: relativePath,
--- a/src-electron/importexport/export.js
+++ b/src-electron/importexport/export.js
@@ -136,7 +136,7 @@ async function exportSessionPackages(db, sessionId, zapProjectFileLocation) {
         pathRelativity = dbEnum.pathRelativity.relativeToZap
       }
     }
-    if (path.isAbsolute(relativePath) && /^[A-Z]:\\/.test(relativePath)) {
+    if (path.isAbsolute(relativePath) && /^[A-Z]:\\/i.test(relativePath)) {
       // Handling Windows path when package is on different drive than zapProjectFile or user home
       pathRelativity = dbEnum.pathRelativity.absolute
     }
--- a/src-electron/importexport/import-json.js
+++ b/src-electron/importexport/import-json.js
@@ -737,6 +737,7 @@ async function importEvents(
  * @param {*} db
  * @param {*} packages
  * @param {*} zapFilePath
+ * @param {*} sessionId
  */
 async function validatePackagePathDrive(db, packages, zapFilePath, sessionId) {
   // If zap file path is not a windows path, skip check
@@ -756,7 +757,7 @@ async function validatePackagePathDrive(db, packages, zapFilePath, sessionId) {
         "Package path '" +
         packagePath +
         "' is on a different drive than the .zap file path. " +
-        'It is suggested to have all packages on the same drive as the .zap file.'
+        'It is recommended that all packages reside on the same drive as the .zap file.'
       querySessionNotice.setNotification(
         db,
         'WARNING',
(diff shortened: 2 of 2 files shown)


src-electron/importexport/export.js
```diff
@@ -136,6 +136,10 @@ async function exportSessionPackages(db, sessionId, zapProjectFileLocation) {
         pathRelativity = dbEnum.pathRelativity.relativeToZap
       }
     }
+    if (path.isAbsolute(relativePath) && /^[A-Z]:/.test(relativePath)) {
```
ethanzhouyc, line 139: Is regex `/^[A-Z]:/` checking patterns like `C:` ? Checking for `C:/` would be better here

ethanzhouyc (approved):

src-electron/importexport/export.js
```diff
@@ -136,6 +136,10 @@ async function exportSessionPackages(db, sessionId, zapProjectFileLocation) {
         pathRelativity = dbEnum.pathRelativity.relativeToZap
       }
     }
+    if (path.isAbsolute(relativePath) && /^[A-Z]:\\/.test(relativePath)) {
```
brdandu, line 139: What happens when a user opens a sample app in Studio where the project work space is in a different drive than the GSDK? Would that use case even hit the export.js#exportSessionPackages?

brdandu:
Could you also provide some content on how you tested this manually?

src-electron/importexport/export.js
brdandu: Could you test the customer flow:
- GSDK in C:/, 
- Make sure SS workspace exists in D:/
- create sample app which has a relative path for templates and zcl json files which do not exist.
- Open the .zap file of this project
- Save it
- Reopen it
 
Make sure generation passes in each of the above use cases of open, save and reopen

dhchandw:
> Could you also provide some content on how you tested this manually?

For the following situations:
- ZCL File, Gen Templates Json File and Custom XML all in the same drive and generate.

- ZCL File, Gen Templates Json File in the same drive, Custom XML in a different drive. Generate and also save the ZAP file.

- Open ZAP file from above and generate

- ZCL File, Gen Templates Json File and Custom XML all in the same drive and save ZAP file in different drive. Open ZAP file and generate.

In all cases generated content was the same as expected and ZAP files stored paths across different drives as expected.

brdandu (approved):

src-electron/importexport/export.js
brdandu: Add a warning that goes into zap.log and notifications UI in zap

brdandu (approved):

src-electron/importexport/export.js
dhchandw: > Could you test the customer flow:
> 
> * GSDK in C:/,
> * Make sure SS workspace exists in D:/
> * create sample app which has a relative path for templates and zcl json files which do not exist.
> * Open the .zap file of this project
> * Save it
> * Reopen it
> 
> Make sure generation passes in each of the above use cases of open, save and reopen

Manually tested this - generation does pass and paths are corrected upon first file save.

src-electron/importexport/import-json.js
```diff
@@ -730,6 +731,45 @@ async function importEvents(
   }
 }
 
+/**
+ * Checks if any package path is on a different drive than the zap file path (Windows only)
+ * and sets appropriate warnings
+ * @param {*} db
+ * @param {*} packages
+ * @param {*} zapFilePath
```
brdandu, line 739: sessionId?

src-electron/importexport/import-json.js
```diff
@@ -730,6 +731,45 @@ async function importEvents(
...
+    if (
+      packagePath &&
+      packagePath.split(':')[0] !== zapDrive &&
+      path.isAbsolute(packagePath)
+    ) {
+      let message =
+        "Package path '" +
+        packagePath +
+        "' is on a different drive than the .zap file path. " +
+        'It is suggested to have all packages on the same drive as the .zap file.'
```
brdandu, line 759: It is recommended that all packages reside on the same drive as the .zap file.

src-electron/importexport/import-json.js
```diff
@@ -730,6 +731,45 @@ async function importEvents(
...
+ * @param {*} packages
+ * @param {*} zapFilePath
+ */
+async function validatePackagePathDrive(db, packages, zapFilePath, sessionId) {
+  // If zap file path is not a windows path, skip check
+  if (!/^[A-Z]:/i.test(zapFilePath)) {
+    return
+  }
+
+  let zapDrive = zapFilePath.split(':')[0]
```
brdandu, line 747: Can : be part of a directory or file name?

brdandu (approved):

src-electron/importexport/import-json.js
dhchandw: Unlikely but even if it happens the first line in the function makes sure it is a Windows path starting with the drive. So it shouldn't matter

PR #1554, opened 2025-03-31
Add Confirmation Dialog on Device Type Feature Toggle and Disable Toggling when Related Cluster is Disabled

ethanzhouyc:
- Added a confirmation dialog when user toggles a device type feature. After viewing the list of elements to be updated, user can choose to proceed with the change or cancel it.
- Disabled toggling of a device type feature if its associated cluster is disabled, and displayed warnings to inform user which cluster is missing.

ZAPP-1549, ZAPP-1552

### DIFF
--- a/src/components/ZclDeviceTypeFeatureManager.vue
+++ b/src/components/ZclDeviceTypeFeatureManager.vue
@@ -88,16 +88,11 @@ limitations under the License.
             </q-tr>
           </template>
         </q-table>
-        <q-dialog v-model="showDialog">
+        <q-dialog v-model="showDialog" :persistent="!noElementsToUpdate">
           <q-card>
             <q-card-section>
               <div class="row items-center">
-                <div class="text-h6 col">Updated Elements</div>
-                <div class="col-1 text-right">
-                  <q-btn dense flat icon="close" v-close-popup>
-                    <q-tooltip>Close</q-tooltip>
-                  </q-btn>
-                </div>
+                <div class="text-h6 col">Elements to be updated</div>
               </div>
               <div v-if="attributesToUpdate.length > 0">
                 <div
@@ -108,7 +103,9 @@ limitations under the License.
                 </div>
                 <ul>
                   <li
-                    v-for="(attribute, index) in attributesToUpdate"
+                    v-for="(attribute, index) in processElementsForDialog(
+                      attributesToUpdate
+                    )"
                     :key="'attribute' + index"
                     style="margin-bottom: 10px"
                   >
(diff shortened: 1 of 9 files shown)


paulr34:
It would be smooth if this happened once and not every single time, what do you think @ethanzhouyc @tecimovic @brdandu ?

ethanzhouyc (commented):
Confirmation dialog on feature toggle:
image

Disable toggling and display warnings for features with disabled cluster.
image

paulr34:
> Confirmation dialog on feature toggle:
> 
> image
> 
> 
> 
> Disable toggling and display warnings for features with disabled cluster.
> 
> image
> 
> 
> 
> 

The user could just disable the feature and get the same results right?

ethanzhouyc:
@paulr34  I think this confirmation should happen on every feature toggle to ensure ZAP doesn’t notify users about changes they didn’t want but were already committed. I am planning to set a meeting for getting feedback from the team after completing all features and we could discuss this then.

paulr34:
> @paulr34 When user enables and then disables a feature, ZAP might "correct" some elements to align with the expected conformance. While this is technically correct, it could still modify user configurations that they didn't intend to change. By allowing the user to cancel the change, we ensure ZAP doesn't modify anything without explicit confirmation.

Sure that makes sense. It would be great if I could opt out by clicking "don't show me this again" but otherwise it's great good job

paulr34:
> > @paulr34 When user enables and then disables a feature, ZAP might "correct" some elements to align with the expected conformance. While this is technically correct, it could still modify user configurations that they didn't intend to change. By allowing the user to cancel the change, we ensure ZAP doesn't modify anything without explicit confirmation.
> 
> Sure that makes sense. It would be great if I could opt out by clicking "don't show me this again" but otherwise it's great good job

Correction: "don't ask me this again" if you want to show what happened

ethanzhouyc:
That's a cool idea! I take it down and will bring it up when discussing feedback with team.

src/components/ZclDeviceTypeFeatureManager.vue
```diff
@@ -371,6 +441,34 @@ export default {
...
+      return this.getMissingClusterSide(feature).length > 0
+    },
+    getMissingClusterSide(feature) {
+      let sides = []
+      let clusterId = feature.clusterRef
+      if (
+        !this.selectionClients.includes(clusterId) &&
+        feature.includeClient == 1
+      ) {
+        sides.push('client')
```
ethanzhouyc, line 455: switch to enums

ethanzhouyc:
Issue https://github.com/project-chip/zap/issues/1563 created to add Cypress tests covering the feature page related code in future PR

brdandu (approved):
Just fix the comments. Looks good!

ethanzhouyc:
@paulr34 #1568 Issue created on the 'Don't shown me again' option and will discuss before merge

ethanzhouyc:
Discussed with team and the confirmation dialog should be displayed on each toggle.

paulr34 (approved):

PR #1557, opened 2025-04-01
Handling struct items properly when we have a global and cluster specific struct with the same name

brdandu:
- Fixed the zcl_struct_items_by_struct_and_cluster_name helper for this global vs cluster specific struct items
- Added tests for it
- Github: ZAP#1551

### DIFF
--- a/src-electron/db/query-zcl.js
+++ b/src-electron/db/query-zcl.js
@@ -498,66 +498,73 @@ async function selectAllStructItemsByStructName(
   packageIds,
   clusterName = null
 ) {
-  let clusterJoinQuery = ''
-  let clusterWhereQuery = ''
+  let clusterCondition = clusterName
+    ? `CLUSTER.NAME = "${clusterName}"`
+    : `CLUSTER.NAME IS NULL`
+
+  let structItemQuery = `
+  SELECT
+    SI.FIELD_IDENTIFIER,
+    SI.NAME,
+    (SELECT DATA_TYPE.NAME FROM DATA_TYPE WHERE DATA_TYPE.DATA_TYPE_ID = SI.DATA_TYPE_REF) AS TYPE,
+    DT.NAME AS DATA_TYPE_REF_NAME,
+    DISCRIMINATOR.NAME AS DISCRIMINATOR_NAME,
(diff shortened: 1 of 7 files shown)


src-electron/db/query-zcl.js
```diff
@@ -498,23 +498,17 @@ async function selectAllStructItemsByStructName(
   packageIds,
   clusterName = null
 ) {
-  let clusterJoinQuery = ''
   let clusterWhereQuery = ''
   if (clusterName) {
```
bzbarsky-apple, line 502: With these changes, for a global struct none of the items are being shown.

To be clear, for a global struct what will come through here is some name for the struct _and some name for the cluster where the struct is being used_.  And when the lookup fails for that struct in that cluster, a lookup for a global struct with that name should happen....

bzbarsky-apple (changes requested):

test/gen-template/matter3/miscellaneous_helper_tests.zapt
```diff
@@ -30,6 +30,14 @@ TargetStruct item {{index}} from Access Control cluster: {{name}}
 TargetStruct item {{index}} from Binding cluster: {{name}}
 {{/zcl_struct_items_by_struct_and_cluster_name}}
 
+{{#zcl_struct_items_by_struct_and_cluster_name "SemanticTagStruct" "Mode Select"}}
+SemanticTagStruct item {{index}} from Mode Select cluster: {{name}}
+{{/zcl_struct_items_by_struct_and_cluster_name}}
+
+{{#zcl_struct_items_by_struct_and_cluster_name "SemanticTagStruct" ""}}
```
bzbarsky-apple, line 37: This test should be passing a nonempty string that is not "Mode Select" (e.g. "Descriptor" to really match what Matter SDK does) to test the desired behavior of this helper.

src-electron/db/query-zcl.js
brdandu: Cleaned this up. Thanks

bzbarsky-apple (approved):

ethanzhouyc (approved):

tecimovic (approved):

PR #1558, opened 2025-04-02
Cleaning up loadPartial to make sure they are registered properly before generation

brdandu:
- The partialFile loading await fsPromise.readFile(singlePkg.path, 'utf8'); is now done before registering the partial.
- We are hitting a use case intermittently where the partial is not loaded and generation is triggered. Believe that this was happening due to partial file loading and registering happening asynchronously(Not always before generation).
- JIRA: ZAPP-1623

### DIFF
--- a/src-electron/generator/generation-engine.js
+++ b/src-electron/generator/generation-engine.js
@@ -742,10 +742,15 @@ async function generateAllTemplates(
   })
 
   // Next load the partials
-  packages.forEach((singlePkg) => {
-    if (singlePkg.type == dbEnum.packageType.genPartial) {
-      templateEngine.loadPartial(hb, singlePkg.category, singlePkg.path)
-    }
+  const partialPackages = packages.filter(
+    (pkg) => pkg.type === dbEnum.packageType.genPartial
+  )
+  const partialDataArray = await Promise.all(
+    partialPackages.map((pkg) => fsPromise.readFile(pkg.path, 'utf8'))
+  )
+  partialDataArray.forEach((data, index) => {
+    const singlePkg = partialPackages[index]
+    templateEngine.loadPartial(hb, singlePkg.category, data)
   })
 
   // Let's collect the required list of helpers.
--- a/docs/api.md
+++ b/docs/api.md
@@ -7770,9 +7770,9 @@ Get endpoint type events from the given endpoint type ID.
     * [~produceContent(hb, metaInfo, db, sessionId, singlePkg, overridePath:)](#module_JS API_ generator logic..produceContent) ⇒
     * [~wrapOverridable(originalFn, overrideFn)](#module_JS API_ generator logic..wrapOverridable) ⇒
     * [~loadOverridable(path)](#module_JS API_ generator logic..loadOverridable)
-    * [~loadPartial(path)](#module_JS API_ generator logic..loadPartial)
+    * [~loadPartial(hb, name, data)](#module_JS API_ generator logic..loadPartial)
     * [~helperWrapper(wrappedHelper)](#module_JS API_ generator logic..helperWrapper) ⇒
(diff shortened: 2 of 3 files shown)


src-electron/generator/generation-engine.js
```diff
@@ -742,11 +742,12 @@ async function generateAllTemplates(
   })
 
   // Next load the partials
-  packages.forEach((singlePkg) => {
+  for (const singlePkg of packages) {
     if (singlePkg.type == dbEnum.packageType.genPartial) {
-      templateEngine.loadPartial(hb, singlePkg.category, singlePkg.path)
+      const data = await fsPromise.readFile(singlePkg.path, 'utf8')
```
ethanzhouyc, line 747: Why we have to separate the line `const data = await fsPromise.readFile(singlePkg.path, 'utf8')` out of `loadPartial`?
Will the following code just do the same thing?
```
for (const singlePkg of packages) {
  await templateEngine.loadPartial(hb, singlePkg.category, singlePkg.path)
}
```

src-electron/generator/generation-engine.js
```diff
@@ -742,11 +742,12 @@ async function generateAllTemplates(
   })
 
   // Next load the partials
-  packages.forEach((singlePkg) => {
+  for (const singlePkg of packages) {
```
tecimovic, line 745: Wouldn't this be easier handled by simply doing Promise.all() instead of making the load partial be non-async function?

src-electron/generator/generation-engine.js
tecimovic, line 747: I would keep loadPartial as an async function with path as an argument. We just can't run them with simple forEach.

tecimovic (changes requested):

ethanzhouyc (approved):

tecimovic:
I don't like making loadPartial() a non async function. In the long run, it's proper for that to be async function as it could be a "long duration" kind of a thing, whether we read a file or not.

But the problem here was that we were not awaiting for them properly.

We can either do Promise.all() or we can do one of those "run  promises one-by-one until they're done" that we have somewhere. But yeah, we can't do a simple forEach like we did there.

src-electron/generator/generation-engine.js
brdandu: loadHelper is actually not an async function. The file read function is sometimes not finished before the template generation happens and partials are not loaded when this happens

brdandu:
> I don't like making loadPartial() a non async function. In the long run, it's proper for that to be async function as it could be a "long duration" kind of a thing, whether we read a file or not.
> 
> But the problem here was that we were not awaiting for them properly.
> 
> We can either do Promise.all() or we can do one of those "run promises one-by-one until they're done" that we have somewhere. But yeah, we can't do a simple forEach like we did there.

The partials basically have to be loaded and registered before the templates are generated or else we hit partial not found errors

src-electron/generator/generation-engine.js
brdandu: So that partials are surely registered before template generation happens. Due to that async file read call generation sometimes starts before templates are registered.

ethanzhouyc:
```
await loadPartial()

async loadPartial() {
  await readFile()
  registerPartial()
}
```
```
await readFile()
loadPartial()

loadPartial() {
   registerPartial()
}
```

Correct me if I am wrong. These 2 are doing the same thing right? Why we have to move `readFile()` out?

brdandu: yup those are the same, we could have done that too. With the current changes readFile is done as a promise array in parallel. We wait for it and then partials are registered.

brdandu:
Have no idea why this change is causing regen diff on Matter. @bzbarsky-apple any ideas on changes to Matter codegen?

PR #1562, opened 2025-04-03
Clean up helper when structs with same name exist as a global struct and a cluster specific one

brdandu:
- Add selectStructByNameAndClusterName because selectStructByNameAndClusterId could not be used for some of our existing helpers such as asObjectiveCClass.
- Update query-util.js to form the correct data type retrieval using data type name and cluster name: sqlQueryForDataTypeByNameAndClusterName
- Update asObjectiveCClass with selectStructByNameAndClusterName instead of selectStructByName
- Github: ZAP#1559

### DIFF
--- a/src-electron/db/query-struct.js
+++ b/src-electron/db/query-struct.js
@@ -181,6 +181,10 @@ async function selectStructByNameAndClusterName(
     .then((rows) => rows.map(dbMapping.map.struct))
   if (res && res.length == 1) {
     return res[0]
+  } else if (res && res.length > 1) {
+    throw new Error(
+      `More than one struct ${name} exists with same name for ${clusterName} cluster.`
+    )
   } else {
     queryWithClusterName = queryUtil.sqlQueryForDataTypeByNameAndClusterName(
       dbEnum.zclType.struct,
--- a/src-electron/db/query-struct.js
+++ b/src-electron/db/query-struct.js
@@ -25,6 +25,7 @@ const dbApi = require('./db-api')
 const dbCache = require('./db-cache')
 const dbMapping = require('./db-mapping')
 const queryUtil = require('./query-util')
+const dbEnum = require('../../src-shared/db-enum.js')
 
 /**
  * Get all structs from a given package ID.
@@ -132,12 +133,12 @@ ORDER BY
  */
 async function selectStructByNameAndClusterId(db, name, clusterId, packageIds) {
   let queryWithoutClusterId = queryUtil.sqlQueryForDataTypeByNameAndClusterId(
-    'struct',
+    dbEnum.zclType.struct,
     null,
     packageIds
   )
   let queryWithClusterId = queryUtil.sqlQueryForDataTypeByNameAndClusterId(
-    'struct',
+    dbEnum.zclType.struct,
     clusterId,
     packageIds
   )
@@ -151,6 +152,44 @@ async function selectStructByNameAndClusterId(db, name, clusterId, packageIds) {
     return dbApi
       .dbGet(db, queryWithClusterId, [name, clusterId])
       .then(dbMapping.map.struct)
+  }
+}
+
+/**
+ * Select a struct matched by name and cluster name
+ * Note: Use selectStructByNameAndClusterName but this was needed for backwards compatibility.
+ * @param {*} db
+ * @param {*} name
+ * @param {*} clusterName
+ * @param {*} packageIds
+ * @returns struct information or undefined
+ */
+async function selectStructByNameAndClusterName(
+  db,
+  name,
+  clusterName,
+  packageIds
+) {
+  let queryWithClusterName = queryUtil.sqlQueryForDataTypeByNameAndClusterName(
+    dbEnum.zclType.struct,
+    name,
+    clusterName,
+    packageIds
+  )
+  let res = await dbApi
+    .dbAll(db, queryWithClusterName)
+    .then((rows) => rows.map(dbMapping.map.struct))
+  if (res && res.length == 1) {
+    return res[0]
+  } else {
+    queryWithClusterName = queryUtil.sqlQueryForDataTypeByNameAndClusterName(
+      dbEnum.zclType.struct,
+      name,
+      null, // Retrieving global data types since cluster specific ones were not found.
+      packageIds
+    )
+    res = await dbApi.dbGet(db, queryWithClusterName).then(dbMapping.map.struct)
+    return res
   }
 }
 
@@ -208,5 +247,6 @@ exports.selectStructByName = dbCache.cacheQuery(selectStructByName)
 exports.selectStructByNameAndClusterId = dbCache.cacheQuery(
   selectStructByNameAndClusterId
 )
+exports.selectStructByNameAndClusterName = selectStructByNameAndClusterName
 exports.selectStructsWithClusterAssociation =
   selectStructsWithClusterAssociation
--- a/src-electron/db/query-util.js
+++ b/src-electron/db/query-util.js
@@ -117,8 +117,7 @@ function sqlQueryForDataTypeByNameAndClusterName(
     DATA_TYPE.NAME AS NAME,
     ${numberExtensionString}
     ${typeTableName}.SIZE AS SIZE,
-    CLUSTER.NAME AS CLUSTER_NAME,
-    REPLACE(CLUSTER.NAME, ' ', '') AS CLUSTER_NAME_WITHOUT_SPACES
+    CLUSTER.NAME AS CLUSTER_NAME
   FROM ${typeTableName}
   INNER JOIN
     DATA_TYPE
@@ -139,7 +138,7 @@ function sqlQueryForDataTypeByNameAndClusterName(
 
   let whereClause = !options.ignoreClusterWhereClause
     ? clusterName
-      ? `AND (CLUSTER.NAME = '${clusterName}' OR CLUSTER_NAME_WITHOUT_SPACES = '${clusterName}')`
+      ? `AND CLUSTER.NAME = '${clusterName}'`
       : `AND CLUSTER.NAME IS NULL`
     : ``
 
--- a/src-electron/db/query-util.js
+++ b/src-electron/db/query-util.js
@@ -83,5 +83,71 @@ function sqlQueryForDataTypeByNameAndClusterId(
   return clusterId ? queryWithClusterId : queryWithoutClusterId
 }
 
+/**
+ * Formulate a sqlite query string for a data type from the given cluster name and package IDs.
+ * @param {*} typeDiscriminator
+ * @param {*} name data type name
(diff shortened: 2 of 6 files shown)


src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -240,9 +240,10 @@ async function asObjectiveCClass(type, cluster, options) {
...
 
   if (isStruct) {
-    const structObj = await zclQuery.selectStructByName(
+    const structObj = await zclQuery.selectStructByNameAndClusterName(
       this.global.db,
       type,
+      cluster,
       pkgIds
     );
     if (structObj.structClusterCount == 0) {
```
bzbarsky-apple, line 249: So when I try this I get:
```
TypeError: Cannot read properties of undefined (reading 'structClusterCount')
    at Object.asObjectiveCClass (/Users/bzbarsky/zap/dist/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js:203:23)
    at Object.asObjectiveCType (/Users/bzbarsky/zap/dist/src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js:238:19)
```

because `selectStructByName` populates that by the "by name and ..." APIs do not.

bzbarsky-apple (changes requested):

paulr34 (approved):

src-electron/db/query-util.js
```diff
@@ -83,5 +83,72 @@ function sqlQueryForDataTypeByNameAndClusterId(
...
+ * @returns SQLite query string
+ */
+function sqlQueryForDataTypeByNameAndClusterName(
+  typeDiscriminator,
+  name,
+  clusterName,
+  packageIds,
+  options = {}
+) {
+  //let clusterNameWithNoSpaces = clusterName ? clusterName.replace(/\s+/g, '') : clusterName;
```
ethanzhouyc, line 102: delete the commented line?

src-electron/db/query-struct.js
```diff
@@ -154,6 +155,60 @@ async function selectStructByNameAndClusterId(db, name, clusterId, packageIds) {
...
+    return res[0]
+  } else {
+    queryWithClusterName = queryUtil.sqlQueryForDataTypeByNameAndClusterName(
+      dbEnum.zclType.struct,
+      name,
+      null, // Retrieving global data types since cluster specific ones were not found.
+      packageIds
+    )
+    res = await dbApi.dbGet(db, queryWithClusterName).then(dbMapping.map.struct)
+    if (!res) {
```
ethanzhouyc, line 192: should we check whether res.data is empty instead of res?

src-electron/db/query-struct.js
```diff
@@ -154,6 +155,60 @@ async function selectStructByNameAndClusterId(db, name, clusterId, packageIds) {
...
+      // Still no struct is found. Ignore where clause
+      // Note: Seems like there is a use case where the cluster xml file name
+      // is passed instead of cluster name in certain helpers arguments.
+      // Believe we should break here than actually satisfy this if condition.
```
ethanzhouyc, line 196: The clarity of this comment could be improved somehow

ethanzhouyc (approved):

src-electron/db/query-struct.js
bzbarsky-apple, line 196: Ah, this is for the "backward compat names" bit.  This should be handled over in the Darwin bits, since it's specific to those.  I'll make a suggestion on that code.

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
```diff
@@ -240,12 +240,14 @@ async function asObjectiveCClass(type, cluster, options) {
...
+    const structObj = await zclQuery.selectStructByNameAndClusterName(
       this.global.db,
       type,
+      cluster,
       pkgIds
     );
-    if (structObj.structClusterCount == 0) {
+
+    if (!structObj.clusterName) {
       // This is a global struct.
```
bzbarsky-apple, line 251: Here's where I think we should handle the cases when "cluster" is not an actual cluster name.  Something like:
```suggestion
    let isGlobalStruct;
    if (!structObj) {
      // This can fail when our "cluster name" is a backwards compat name.  Just
      // fetch by struct name only in that case.
      structObj = await zclQuery.selectStructByName(
        this.global.db,
        type,
        pkgIds
      );

      isGlobalStruct = (structObj.structClusterCount == 0);
    } else {
      isGlobalStruct = !structObj.clusterName;
    }

    if (isGlobalStruct) {
      // This is a global struct.
```

though the fact that the two APIs have totally different ways of representing global structs is kind of annoying.

src-electron/db/query-util.js
```diff
@@ -83,5 +83,72 @@ function sqlQueryForDataTypeByNameAndClusterId(
...
+      : ''
+  let selectQueryString = `
+  SELECT
+    ${typeTableName}.${typeTableName}_ID,
+    ${structExtensionString}
+    DATA_TYPE.NAME AS NAME,
+    ${numberExtensionString}
+    ${typeTableName}.SIZE AS SIZE,
+    CLUSTER.NAME AS CLUSTER_NAME,
+    REPLACE(CLUSTER.NAME, ' ', '') AS CLUSTER_NAME_WITHOUT_SPACES
```
bzbarsky-apple, line 122: I see why you ended up here, but sadly it's not just a "without spaces", though maybe in practice it is: what comes into the Darwin code there is `(asUpperCamelCase clusterName preserveAcronymes=true)` or so.

I don't have any better ideas than this, though this seems extremely unexpected as a caller.  :(  Maybe it really would be better to change all consumers of asObjectiveCType/Class to pass in a cluster id instead...  That's 60-ish callsites, but maybe we can manage that....  

Or of course we could leave this bit, which is obviously less work for me, but might come back to bite us sometime.

src-electron/db/query-struct.js
brdandu: res is just null or undefined for this when a struct is not found.

src-electron/db/query-util.js
brdandu: We should still keep this just to satisfy an old chip release and the latest zap but eventually changing to cluster IDs is safer.

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
brdandu: Had to change it because of the extended use case of same named data types for global vs cluster specific ones. Is there anything you want back? count logic instead?

src-electron/db/query-struct.js
brdandu: removed with suggested change from Boris. What was the confusing part?

src-electron/generator/matter/darwin/Framework/CHIP/templates/helper.js
brdandu: added the suggestion

src-electron/db/query-struct.js
ethanzhouyc: I think 'break here' is not clear enough, but nvm

src-electron/db/query-util.js
```diff
@@ -83,5 +83,71 @@ function sqlQueryForDataTypeByNameAndClusterId(
...
+  ON
+    DATA_TYPE_CLUSTER.CLUSTER_REF = CLUSTER.CLUSTER_ID
+  WHERE
+    (DATA_TYPE.NAME = '${name}' ${checkLowerCaseString})
+  AND
+    DATA_TYPE.PACKAGE_REF IN (${dbApi.toInClause(packageIds)}) `
+
+  let whereClause = !options.ignoreClusterWhereClause
+    ? clusterName
+      ? `AND (CLUSTER.NAME = '${clusterName}' OR CLUSTER_NAME_WITHOUT_SPACES = '${clusterName}')`
```
bzbarsky-apple, line 142: After https://github.com/project-chip/connectedhomeip/pull/38267 lands I don't think you need this bit at all.

src-electron/db/query-struct.js
```diff
@@ -154,6 +155,44 @@ async function selectStructByNameAndClusterId(db, name, clusterId, packageIds) {
...
+  let queryWithClusterName = queryUtil.sqlQueryForDataTypeByNameAndClusterName(
+    dbEnum.zclType.struct,
+    name,
+    clusterName,
+    packageIds
+  )
+  let res = await dbApi
+    .dbAll(db, queryWithClusterName)
+    .then((rows) => rows.map(dbMapping.map.struct))
+  if (res && res.length == 1) {
```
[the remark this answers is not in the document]
brdandu: Added that to cover the empty array returned from dbAll if not found. I can add the error conditioning for safety

src-electron/db/query-util.js
brdandu: My suggestion is to still keep it if someone loads the chip repo before your change is added with latest ZAP it still gives a reasonable output.

brdandu: I guess it may not happen because the corresponding xml changes don't exist yet. However, should we still keep it?
I can remove it if we definitely don't need that safety because the use case won't exist.

bzbarsky-apple: That use case won't exist, I expect, so let's just take this bit out.

src-electron/db/query-struct.js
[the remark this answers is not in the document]
bzbarsky-apple: Ah, so there are three cases: length 0 (do the query without cluster), length 1 (found it), length > 1 (error)?

And "res" is never falsy?  Or sometimes res is falsy and sometimes length 0 depending on ... what?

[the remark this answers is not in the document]
brdandu: I believe dbGet could have returned undefined/null and dbAll returns an empty array if no entries are found on the query. Based on that the cases can differ.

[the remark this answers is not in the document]
bzbarsky-apple: Where is dbGet involved here?  `res` is coming from `dbAll`, and either we get an exception or an array there, as far as I can tell, since that's what the last `then` clause does.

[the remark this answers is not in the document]
bzbarsky-apple: Then why are we checking whether "res" is falsy?  It can't be: it's an array.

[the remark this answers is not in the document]
brdandu: You are right. A lot of the times just end up applying defensive coding because it does not hurt having to check for nulls.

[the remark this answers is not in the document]
bzbarsky-apple: I guess my point is that defending against things that obviously can't happen just leads to confusion as here...  Anyway, approving, either way.

bzbarsky-apple (approved):

PR #1564, opened 2025-04-03
Increasing timeout because for now it is not reasonable for all machines

brdandu:


andy31415 (approved):

dhchandw (approved):
We should create an issue to integrate a watchdog.

brdandu:
> We should create an issue to integrate a watchdog.

https://github.com/project-chip/zap/issues/1567

PR #1569, opened 2025-04-04
Update iteration keyword replacement to have `tokens to upper camel case`

andy31415:
When trying latest zap after #1566 I found out that generation in matter differs from "general string utilites". Specifically for cluster "PM2.5 Concentration Measurement" the names I need are `Pm25ConcentrationMeasurement` (to match class names so I can use `{{asUpperCamelCase name}}` in templates for consistency and compatiblity) where as string utilities gives me `Pm2.5ConcentrationMeasurement`.

This PR updates the naming for things and:
  - moves (and shares) code from matter asCamelCase into a strings.tokensIntoCamelCase
  - adds unit tests
  - updates iterator methods

### DIFF
--- a/src-electron/util/string.js
+++ b/src-electron/util/string.js
@@ -69,7 +69,7 @@ function toSpacedLowercase(str) {
  * "Foo 2.5" becomes "Foo25" rather than "Foo2.5").
  *
  */
-function asCamelCase(label, firstLower, preserveAcronyms) {
+function tokensIntoCamelCase(label, firstLower, preserveAcronyms) {
   const wordSplitRegexp = / |_|-|\//
   let tokens = label.replace(/[+()&]/g, '').split(wordSplitRegexp)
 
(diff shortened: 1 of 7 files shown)


src-electron/util/string.js
```diff
@@ -60,6 +60,74 @@ function toSpacedLowercase(str) {
   return res.toLowerCase()
 }
 
+/**
+ * Converts a string into `camelCase` or `CamelCase` by splitting
+ * the string into tokens.
+ *
+ * Differs from `toCamelCase` by different handling of "." inside names,
+ * so that results are always usable as a code-compatibe name (e.g.
```
bzbarsky-apple, line 68:
```suggestion
 * so that results are always usable as a code-compatible name (e.g.
```

paulr34 (approved):

brdandu (approved):

brdandu:
This PR was rebased and fixed using #1616

PR #1572, opened 2025-04-10
Fix: error in postCluster

dhchandw:


src-electron/rest/user-data.js
```diff
@@ -336,7 +336,6 @@ function httpPostCluster(db) {
           pkg.type == dbEnum.packageType.zclProperties ||
           pkg.type == dbEnum.packageType.zclXmlStandalone
       )
-      pkgs = zclPropertiesPkgs.concat(zclXmlStandalonePkgs)
```
brdandu, line 339: Log bug for cypress tests to check cluster enablement.

brdandu (approved):

paulr34 (approved):

PR #1573, opened 2025-04-11
Fix the Legal Clusters view where users will only see clusters associated with a device type and not just enabled clusters which is wrong

brdandu:
- Use the device types on the selected endpoint to figure out the device type clusters and then only show those clusters in the UI.
- The Legal clusters UI view will also not allow users to enable/disable non compliant clusters and also not show non compliant clusters.
- Github: ZAP#316


brdandu:
@mlepage-google @bzbarsky-apple , Added you to just for FYI and if you guys wanted to try the feature before it is merged. Feel free to ignore the PR review process. However, all reviews are appreciated.

tecimovic (approved):
Cursory sanity check seems ok if the builds pass.

src/store/zap/actions.js
```diff
@@ -709,33 +712,35 @@ export async function updateSelectedEndpointType(
...
-        })
+      ).then((responses) => {
+        const allClusters = responses.flatMap((res) => res.data)
+        setRecommendedClusterList(context, allClusters)
+      })
     )
     p.push(
-      axiosRequests
-        .$serverGet(
-          `${restApi.uri.deviceTypeAttributes}${endpointTypeDeviceTypeRefPair.deviceTypeRef}`
```
paulr34, line 725: Why did you do this? I'm lost

brdandu: This was all broken when multiple device type refs are passed(Which happens when user selects more than one device type on an endpoints). The URIs were like `Identify/1,2` which would return nothing. Now it does `Identify/1 and `Identify/2` and collates the 2 results.

Am surprised no one has complained about this yet.

PR #1577, opened 2025-04-18
Bug/update gen templates json file for nested files/zapp 1297

brdandu:
When json files referenced from within the templates.json file are updated then checksum is calculated for those as well and updated accordingly.


brdandu:
Tests already covered in zcl-loader.test.js `test changing the zcl extension file in a top level templates json file and make sure it is re-loaded again`

paulr34 (approved):

Issue #1526, opened 2025-02-18
Remove redundancy in conformance definitions in ZAP XMLs


ethanzhouyc:
In ZAP XML, both formats like `optional="true"` and `<mandatoryConform>` are used to define compliance, creating redundancy and confusion:

```
<attribute side="server" code="0x0004" name="CurrentY" define="COLOR_CONTROL_CURRENT_Y" type="int16u" max="0xFEFF" default="0x607D" reportable="true" optional="true">
      <mandatoryConform>
        <feature name="XY"/>
      </mandatoryConform>
</attribute>
```

Users don’t know which takes precedence and both must be updated manually. Changing one without updating the other can cause inconsistencies.

Proposed solution:

1. For pure mandatory attributes, no conformance listed at all.  optional="false" is disallowed.
2. For pure optional (O) attributes, allow only optional="true" so we can use the shorthand attribute representation in the XML.
3. For anything else, disallow having the optional attribute entirely, list longhand conformance.

PR #1578, opened 2025-04-18
Resolve element optionality via both optional attribute and conformance

ethanzhouyc:
ZAP XMLs currently declare optionality redundantly—via optional attribute and conformance tag.
This PR makes ZAP interpret either source and ready for removing optional attribute in the future.

Issue: #1526 
GHM_ZAP-477

- Extend optionality logic for attributes, commands, and events: an element is now treated as optional when either `optional="true"` or its conformance ≠ `'M'` (mandatory).
- Add unit tests covering all combinations of optional attribute and conformance definitions
- Update ZAP XMLs for testing purpose. The conformance data are obtained from CHIP ZAP XMLs

Let optional attribute takes precedence over conformance for backwards compatibility like the element below:
```
    <!-- ***NOTE***: UpTime attribute is mandatory starting at Rev >= 2, but because of backwards compatibility, has to be optional here.
         The device type config and cert tests (TC-DGGEN-1.1/2.1) ensure that it is present. -->
    <attribute side="server" code="0x02" name="UpTime" define="UP_TIME" type="int64u" optional="true">
      <mandatoryConform/>
    </attribute>
```

### DIFF
--- a/src-electron/validation/conformance-xml-parser.js
+++ b/src-electron/validation/conformance-xml-parser.js
@@ -21,6 +21,9 @@
  * @module Validation API: Parse conformance data from XML
  */
 
+const dbEnum = require('../../src-shared/db-enum')
+const conformEvaluator = require('./conformance-expression-evaluator')
+
 /**
  * Parses conformance from XML data.
  * The conformance could come from features, attributes, commands, or events
@@ -155,4 +158,30 @@ function parseConformanceRecursively(operand, depth = 0, parentJoinChar = '') {
   }
 }
 
+/**
+ * if optional attribute is defined, return its value
+ * if optional attribute is undefined, check if the element conformance is mandatory
+ * if both optional attribute and conformance are undefined, return false
+ * Optional attribute takes precedence over conformance for backward compatibility on certain elements
+ *
+ * @param {*} element
+ * @returns true if the element is optional, false if the element is mandatory
+ */
+function getOptionalAttributeFromXML(element) {
+  if (element.$.optional) {
+    return element.$.optional == 'true'
+  } else {
+    let conformance = parseConformanceFromXML(element)
+    if (conformance) {
+      return !conformEvaluator.checkIfExpressionHasTerm(
+        conformance,
+        dbEnum.conformance.mandatory
+      )
+    } else {
+      return false
+    }
+  }
+}
+
 exports.parseConformanceFromXML = parseConformanceFromXML
+exports.getOptionalAttributeFromXML = getOptionalAttributeFromXML
--- a/src-electron/validation/conformance-xml-parser.js
+++ b/src-electron/validation/conformance-xml-parser.js
@@ -23,6 +23,7 @@
 
 const dbEnum = require('../../src-shared/db-enum')
 const conformEvaluator = require('./conformance-expression-evaluator')
+const env = require('../util/env')
 
 /**
  * Parses conformance from XML data.
@@ -163,15 +164,23 @@ function parseConformanceRecursively(operand, depth = 0, parentJoinChar = '') {
  * if optional attribute is undefined, check if the element conformance is mandatory
  * if both optional attribute and conformance are undefined, return false
  * Optional attribute takes precedence over conformance for backward compatibility on certain elements
+ * Log warnings to zap.log if both optional attribute and conformance are defined
  *
  * @param {*} element
+ * @param {*} elementType
  * @returns true if the element is optional, false if the element is mandatory
  */
-function getOptionalAttributeFromXML(element) {
+function getOptionalAttributeFromXML(element, elementType) {
+  let conformance = parseConformanceFromXML(element)
   if (element.$.optional) {
+    if (conformance) {
+      env.logWarningToFile(
+        `Redundant 'optional' attribute and 'conformance' tag defined for ${elementType}: ${element.$.name}.` +
+          " 'optional' takes precedence, but consider removing it as 'conformance' is the recommended format."
+      )
+    }
     return element.$.optional == 'true'
   } else {
-    let conformance = parseConformanceFromXML(element)
     if (conformance) {
       return !conformEvaluator.checkIfExpressionHasTerm(
         conformance,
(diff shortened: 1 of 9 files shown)


paulr34 (approved):

zcl-builtin/matter/data-model/chip/door-lock-cluster.xml
```diff
@@ -67,16 +67,27 @@ limitations under the License.
...
-        <attribute side="server" code="0" define="LOCK_STATE" type="DlLockState" min="0" max="2" isNullable="true" reportable="true" writable="false">LockState</attribute>
-        <attribute side="server" code="1" define="LOCK_TYPE" type="DlLockType" min="0" max="10" writable="false">LockType</attribute>
+        <attribute side="server" code="0" define="LOCK_STATE" type="DlLockState" min="0" max="2" isNullable="true" reportable="true" writable="false">
+            <description>LockState</description>
+            <mandatoryConform/>
+        </attribute>
+        <attribute side="server" code="1" define="LOCK_TYPE" type="DlLockType" min="0" max="10" writable="false" optional="false">LockType</attribute>
         <attribute side="server" code="2" define="ACTUATOR_ENABLED" type="boolean" writable="false">ActuatorEnabled</attribute>
         <!-- Conformance feature DPS - for now optional -->
-        <attribute side="server" code="3" define="DOOR_STATE" type="DoorStateEnum" min="0" max="5" isNullable="true" reportable="true" optional="true">DoorState</attribute>
```
brdandu, line 74: In order to make sure we are not breaking backwards compatibility I would let optional="true" stay in the xml files within zap because the purpose of these is to just test. If optional is present and so is conformance then pick conformance instead of optional.

src-electron/validation/conformance-xml-parser.js
```diff
@@ -155,4 +155,22 @@ function parseConformanceRecursively(operand, depth = 0, parentJoinChar = '') {
...
+ * if both optional attribute and conformance are undefined, return false
+ *
+ * @param {*} element
+ * @returns true if the element is optional, false if the element is mandatory
+ */
+function getOptionalAttributeFromXML(element) {
+  if (element.$.optional) {
+    return element.$.optional == 'true'
+  } else {
+    let conformance = parseConformanceFromXML(element)
```
brdandu, line 170: Should conformance get precedence over optional since it is the newer format?

src-electron/validation/conformance-xml-parser.js
```diff
@@ -155,4 +155,22 @@ function parseConformanceRecursively(operand, depth = 0, parentJoinChar = '') {
...
+    return conformance ? conformance != 'M' : false
```
brdandu, line 171: Can 'M' be part of conformance enums?

test/zcl-loader.test.js
```diff
@@ -594,14 +595,14 @@ test(
       let ctx = await zclLoader.loadZcl(db, env.builtinMatterZclMetafile())
       let packageId = ctx.packageId
 
-      let zclCluster = await queryZcl.selectClusterByCode(db, packageId, 0x001f)
+      let aclCluster = await queryZcl.selectClusterByCode(db, packageId, 0x001f)
```
brdandu, line 598: what is acl? accessControl?

test/zcl-loader.test.js
```diff
@@ -623,3 +624,107 @@ test(
   },
   testUtil.timeout.long()
 )
+
+test(
+  'test loading IS_OPTIONAL column for Matter attributes, commands, and events',
+  async () => {
+    let db = await dbApi.initRamDatabase()
```
brdandu, line 631: If both formats are present then add a test to pick up the write precedence order if possible for optional vs conformance.

brdandu (changes requested):

zcl-builtin/matter/data-model/chip/door-lock-cluster.xml
ethanzhouyc: I updated with no optional attribute removed. But also some of them with wrong value in conflict with conformance to test whether the conformance takes precedence.

src-electron/validation/conformance-xml-parser.js
ethanzhouyc: Updated in the new commit, the if branch checks conformance first and then optional attribute

test/zcl-loader.test.js
ethanzhouyc: Yes. This is a typo fix to a previous PR, unrelated to this PR.

test/zcl-loader.test.js
ethanzhouyc: Tests added for both conformance and optional exist and in conflict with each other

ethanzhouyc:
https://github.com/project-chip/connectedhomeip/pull/38644
Created a CHIP PR to fix the failed Matter code gen CI

src-electron/validation/conformance-xml-parser.js
```diff
@@ -155,4 +158,30 @@ function parseConformanceRecursively(operand, depth = 0, parentJoinChar = '') {
...
+function getOptionalAttributeFromXML(element) {
+  if (element.$.optional) {
+    return element.$.optional == 'true'
+  } else {
+    let conformance = parseConformanceFromXML(element)
+    if (conformance) {
+      return !conformEvaluator.checkIfExpressionHasTerm(
+        conformance,
+        dbEnum.conformance.mandatory
+      )
```
bzbarsky-apple, line 179: "conformance" here is a string, right?  And it's the sort of conformance string that appears in the spec?

It's a little weird to convert a data structure into a string and then start doing string matching on it, instead of just doing matching on the data structure directly.  It's making some interesting assumptions about the structure of feature names, for example.  Could a feature be named "M-9"?  Or have non-ASCII characters in its name?

It would really be better if we just walked the data structure here instead of doing this string stuff.

ethanzhouyc: Thanks for review! Will discuss with team on the best way to store the conformance data structure directly in ZAP database.

PR #1580, opened 2025-04-22
Adding default value column for command args, event fields and struct items

dhchandw:
- Default values can be defined for command args, event fields and struct items. These default values can be accessed through zcl static helpers
- Added tests


paulr34 (approved):

test/gen-matter-1.test.js
```diff
@@ -293,6 +293,19 @@ test(
...
+    // Testing default values for command arguments, event fields and struct items
+    expect(sdkExt).toContain('ProductID - int16u - default_value=0x01')
+    expect(sdkExt).toContain(
+      'RequestorCanConsent - boolean - default_value=false'
+    )
+    expect(sdkExt).toContain(
+      'Struct name: ChannelInfoStruct, Struct Item Name: MajorNumber, Struct Item Type: int16u, Struct Default Value: 0xFFFF'
+    )
+    let eventOut = genResult.content['events.out']
+    expect(eventOut).toContain(
```
brdandu, line 306: Add a not contain test too just to cover negative test cases

zcl-builtin/matter/data-model/chip/channel-cluster.xml
```diff
@@ -54,7 +54,7 @@ limitations under the License.
 
   <struct name="ChannelInfoStruct">
     <cluster code="0x0504"/>
-    <item name="MajorNumber"       type="INT16U"/>
+    <item name="MajorNumber"       type="INT16U" default="0xFFFF"/>
```
brdandu, line 57: what happens if `default=null` ?

brdandu (approved):

PR #1584, opened 2025-04-26
Fix various helpers that were not handling name collisions between global and cluster-specific structs.

bzbarsky-apple:


bzbarsky-apple:
NOTE: the failure on Matter codegen is expected; see comment about the topological sort used not being stable.  So sorting the two arrays of global and non-global structs separately does not produce the same output as sorting them together, and in general adding any struct at all can change the sorting always....

ReneJosefsen (approved):

ethanzhouyc (approved):

tecimovic (approved):

PR #1588, opened 2025-05-02
wait for server

paulr34:


dhchandw (approved):
Assuming it's tested

src-electron/ide-integration/studio-rest-api.js
```diff
@@ -136,6 +136,31 @@ function wsApiUrl(api, path) {
...
 }
 
+/**
+ * Retry connecting to the Studio server until it is ready.
+ * @param {string} url - The URL to connect to.
+ * @param {number} retries - Number of retry attempts.
+ * @param {number} delay - Delay between retries in milliseconds.
+ * @returns {Promise} - Resolves when the connection is successful, rejects if it fails.
+ */
+async function waitForServer(url, retries = 5, delay = 2000) {
```
brdandu, line 146: Not a big deal but passing those numbers as arguments from the call is cleaner than defining defaults here(5 and 2000).

brdandu (approved):
Was this issue reproducible? When does this happen? If so then see if unit tests can be added.

ethanzhouyc (approved):

PR #1589, opened 2025-05-02
Adding upgrade rule mechanism in ZAP for SDKs

brdandu:
- Users can now include an upgrade rules json file in the zcl.json file which has a list of all the upgrade rules that may be needed when a user updates an existing app from one version of the GSDK to another.
- The upgrade rules json file lists a set of upgrade rules scripts with their priority as can be seen in upgrade-rules.json file. The scripts are run in order of priority with lower number signifying higher priority
- The upgrade rules return an object including a message and status which helps determine if an upgrade rules was actually executed or not. These results can be output into a yaml file if required for post analysis on which upgrade rules ran on a certain .zap file once the GSDK was updated.
- Github: ZAP#1020


brdandu (commented):
- Add documentation
- Add unit tests

docs/zap-file-upgrade.md
```diff
@@ -0,0 +1,91 @@
...
+    },
+    {
+      "path": "../../test/resource/test-attribute-default-value-update.js",
+      "priority": 100
+    }
+  ]
+}
+```
+
+**category**: Determines that these upgrade rules need to run for matter.
```
ethanzhouyc, line 37: What else can user put in category?

ethanzhouyc: And the user can upgrade for matter on zigbee .zap files?

src-electron/db/query-upgrade.js
```diff
@@ -0,0 +1,98 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ * This module provides queries for upgrade rules.
+ *
+ * @module DB API: endpoint configuration queries against the database.
+ */
```
ethanzhouyc, line 22: Do we want to create a new file for only 1 function? It's good if more functions are coming to here in the future.

src-electron/util/post-import-api.js
```diff
@@ -288,6 +293,30 @@ async function modifyAttribute(
...
+    endpoint.endpointTypeRef,
+    cluster.id,
+    attribute.side,
+    attribute.id,
+    params,
+    attribute.reportMinInterval,
+    attribute.reportMaxInterval,
+    attribute.reportableChange
+  )
+}
```
ethanzhouyc, line 318: Should this function be async?

docs/zap-file-upgrade.md
brdandu: Yes if more than 2 zcl json files are not passed then users could pass a any zcl json with upgrade rules in it and they may get executed.

src-electron/db/query-upgrade.js
brdandu: This is the part I did not like either for just one file but we may have to extend some other queries for attributes and commands here(was mostly needed for categorization). Will see and move this out of this file.

docs/zap-file-upgrade.md
```diff
@@ -0,0 +1,91 @@
...
+
+**upgradeRuleScripts**: List of upgrade rules to run on .zap files. Includes the relative path from upgrade-rules.json to the upgrade scripts written in Javascript. Priority determines the order of execution for the upgrade rules.
+
```
ethanzhouyc, line 40: `The scripts are run in order of priority with lower number signifying higher priority`
Add this from your PR description in the document too?

ethanzhouyc: Would be good if we can add tests for if priority defines execution order correctly

tecimovic (approved):
In principle this is ok, but I'd lie if I said I reviewed every line of code....

PR #1593, opened 2025-05-02
Fix Matter helpers to do the right thing with global enums/bitmaps.

bzbarsky-apple:
NOTE: the Matter generation changes here are somewhat expected, because some of the new template changes in Matter are buggy until https://github.com/project-chip/connectedhomeip/pull/38733 merges.  There should be no compat impact here for Matter bits from older Matter releases.

### DIFF
--- a/src-electron/generator/matter/app/zap-templates/templates/app/helper.js
+++ b/src-electron/generator/matter/app/zap-templates/templates/app/helper.js
@@ -648,6 +648,52 @@ function nsValueToPythonNamespace(ns, clusterCount) {
   return ns;
 }
 
+// Not to be exported.
+//
+// dataType can be "Enum", "Struct", or "Bitmap".
+async function getClusterCountForType(db, pkgId, type, dataType, options) {
+  if (options.hash.cluster === '') {
+    // This is a non-global data type that is associated with multiple
+    // clusters: in that case our caller can pass in cluster="", and
+    // we know that clusterCount > 1, so just set it to 2.
+    return 2;
+  }
+
+  const cluster = options.hash.cluster || options.hash.ns;
+  const typeObj = await zclQuery[`select${dataType}ByNameAndClusterName`](
+    db,
+    type,
+    cluster,
+    pkgId
+  );
+  if (typeObj) {
+    return typeObj[`${dataType.toLowerCase()}ClusterCount`];
+  }
+
+  if (options.hash.cluster === undefined) {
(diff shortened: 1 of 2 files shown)


paulr34 (approved):

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
```diff
@@ -717,57 +765,28 @@ async function zapTypeToClusterObjectType(type, isDecodable, options) {
...
+      const clusterCount = await getClusterCountForType(
         this.global.db,
         pkgId,
-        type
-      );
-      const ns = nsValueToNamespace(
-        options.hash.ns,
-        bitmapObj.bitmapClusterCount
+        type,
+        'Bitmap',
```
brdandu, line 772: Minor cleanup: Switch to `db-enum.zclType.bitmap` ? same for enums and structs

brdandu (commented):
There is a regen check failure on the Matter SDK. Is that expected?

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
bzbarsky-apple: The problem is the casing @brdandu .  In the function I need the strings "bitmap" and "Bitmap".

db-enum.zclType.bitmap is "bitmap", so to get "Bitmap" you need to do some complicated "uppercase the first letter, append the rest" thing.  To get "bitmap" from "Bitmap" you just lowercase the whole thing.  So passing in Bitmap here makes the callee code a lot simpler.

But I can make this change if desired.

bzbarsky-apple:
> There is a regen check failure on the Matter SDK. Is that expected?

Yes.  See PR description.

brdandu (approved):

src-electron/generator/matter/app/zap-templates/templates/app/helper.js
brdandu: Should a new set of enums be created?

bzbarsky-apple: I don't think it's worth creating a separate set of enums.  If you think enum should be used here, I can just do the uppercase first letter thing.

bzbarsky-apple: Per Slack discussion, keeping this as-is for now.

PR #1594, opened 2025-05-05
show save button in ide when dirtyflag is sent to parent

paulr34:


ethanzhouyc (approved):

src/components/ZCLToolbar.vue
```diff
@@ -245,6 +243,12 @@ export default {
   name: 'ZCLToolbar',
   mixins: [CommonMixin],
   computed: {
+    query() {
+      return this.$store.state.zap.query // Access the query string from Vuex
+    },
+    showSaveButton() {
+      return !!this.query?.stsApplication // Use optional chaining and double negation for clarity
```
brdandu, line 250: What is stsApplication?

src/components/ZCLToolbar.vue
```diff
@@ -61,9 +61,7 @@
       flat
       no-caps
       @click="saveChanges"
-      v-if="
-        this.$store.state.zap.saveButtonVisible && this.$store.state.zap.isDirty
-      "
```
ethanzhouyc, line 66: If `saveButtonVisible` and `isDirty` is not used in other places, after they are deleted, their related code to define or set their value should be removed too

src/components/ZCLToolbar.vue
paulr34: the passed from studio to indicate which version it is..from the long discussions with Balint and @tecimovic

src/components/ZCLToolbar.vue
paulr34: I agree, thanks

brdandu (approved):

PR #1606, opened 2025-06-09
sdk root is always there so check sdkProvidedProperties first 

paulr34:
… if it is present it should be used


brdandu (approved):

brdandu:
Might need a rerun on gen-test failure

ethanzhouyc (approved):

PR #1613, opened 2025-06-25
When closing ZAP sometimes a system error shows up

brdandu:
- Believe that this was being caused by asynchronous reporting that we do for dirty flag and session notification counts on a periodic interval. When we close ZAP some of these reporting events are running queries on a closed database and thus throw a bunch of errors. The changes here stop the asynchronous reporting events from running when the database has been closed.
- JIRA: ZAPP-1607


ethanzhouyc (approved):

ethanzhouyc (commented):
Not related to this PR, but * @param {*} intervalMs should be deleted in the function header comment.

tecimovic (approved):

Issue #1615, opened 2025-07-01
Incorrect min/max defaults for temperature


hasty:
When only min or max is defined for an attribute, the generated value in GENERATED_MIN_MAX_DEFAULTS for the omitted limit does not seem to account for whether or not the underlying value is signed.

e.g.:

```diff
-    <attribute side="server" code="0x0011" name="OccupiedCoolingSetpoint" define="OCCUPIED_COOLING_SETPOINT" type="temperature" min="-27315" max="0x7FFF" writable="true" default="2600" optional="true">
+    <attribute side="server" code="0x0011" name="OccupiedCoolingSetpoint" define="OCCUPIED_COOLING_SETPOINT" type="temperature" min="-27315" writable="true" default="2600">
```

yields

```diff
-        { (uint16_t) 0xA28, (uint16_t) -0x6AB3, (uint16_t) 0x7FFF }, /* OccupiedCoolingSetpoint */                                 \
+        { (uint16_t) 0xA28, (uint16_t) -0x6AB3, (uint16_t) 0xFFFF }, /* OccupiedCoolingSetpoint */     
```

bzbarsky-apple:
The XML relevant here is:

```
<type id="0xD8" description="Temperature"              name="temperature"       size="2"  analog="true"     signed="true" />
```

I think the issue is in `endpoint_attribute_min_max_list` which has this logic:

```
    let def = parseInt(mm.default)
    let min = parseInt(mm.min)
    let max = parseInt(mm.max)

    if (isNaN(def)) def = 0
    if (isNaN(min)) min = 0
    if (isNaN(max)) max = 0xffff
```

Compare `determineAttributeDefaultValue` which does `isSignedInteger` checks and acts accordingly (and ends up as the `mm.default` there).  But for min/max we seem to be just grabbing the values from the XML directly without accounting for signedness anywhere.

I expect the min would be wrong too, if it were left out but max were provided, ending up as 0, not the min value for the type...

brdandu:
This issue seems to have been resolved. Forgot to close the issue after merging the PR. Please re-open if still an issue.

PR #1617, opened 2025-07-03
Fixing the generation for Attribute Min Max used in endpoint_config.h

brdandu:
- Fixing the generation of min and max values to be auto assumed and generated correctly when either min or max are mentioned instead of both min and max
- Also making sure we do not always typecase everything to (uint16_t) but instead typecast the values based on size and sign
- Github: ZAP #1615

### DIFF
--- a/src-electron/generator/helper-endpointconfig.js
+++ b/src-electron/generator/helper-endpointconfig.js
@@ -613,7 +613,8 @@ function endpoint_attribute_min_max_list(options) {
 
     if (isNaN(def)) def = 0
     if (isNaN(min)) min = 0
-    if (isNaN(max)) max = 0xffff
+    if (isNaN(max)) max = '0x' + 'FF'.repeat(mm.typeSize)
+
     let defS =
       (def >= 0 ? '' : '-') + '0x' + Math.abs(def).toString(16).toUpperCase()
     let minS =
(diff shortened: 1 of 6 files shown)


src-electron/generator/helper-endpointconfig.js
```diff
@@ -613,7 +613,8 @@ function endpoint_attribute_min_max_list(options) {
 
     if (isNaN(def)) def = 0
     if (isNaN(min)) min = 0
-    if (isNaN(max)) max = 0xffff
+    if (isNaN(max)) max = '0x' + 'FF'.repeat(mm.typeSize * 2)
```
dhchandw: if (isNaN(max)) max = '0x' + 'FF'.repeat(mm.typeSize);

dhchandw (approved):

tecimovic (approved):

src-electron/generator/helper-endpointconfig.js
```diff
@@ -627,13 +628,19 @@ function endpoint_attribute_min_max_list(options) {
       .forEach((tok) => {
         switch (tok) {
           case 'def':
-            defMinMaxItems.push(`(uint16_t)${defS}`)
+            defMinMaxItems.push(
+              `(${mm.isTypeSigned ? '' : 'u'}int${mm.typeSize * 8}_t)${defS}`
```
bzbarsky-apple, line 632: I suspect this will completely break things, since the actual type those structs store is in fact uint16_t, no?

src-electron/generator/helper-endpointconfig.js
```diff
@@ -613,7 +613,8 @@ function endpoint_attribute_min_max_list(options) {
 
     if (isNaN(def)) def = 0
     if (isNaN(min)) min = 0
-    if (isNaN(max)) max = 0xffff
+    if (isNaN(max)) max = '0x' + 'FF'.repeat(mm.typeSize)
```
bzbarsky-apple, line 616: This is not the correct max for a signed type.

src-electron/generator/helper-endpointconfig.js
```diff
@@ -613,7 +613,8 @@ function endpoint_attribute_min_max_list(options) {
 
     if (isNaN(def)) def = 0
     if (isNaN(min)) min = 0
```
bzbarsky-apple, line 615: This is not the correct min for a signed type.  Should at least have a comment explaining why this is here in this form and why it's OK.

If the expectation is that this case never happens, shouldn't this error out instead of silently doing something wrong?

src-electron/generator/helper-endpointconfig.js
brdandu: Did not cross check that. Just fixed this based on the typecasting being wrong

src-electron/generator/helper-endpointconfig.js
brdandu: Created https://github.com/project-chip/zap/pull/1618/files

src-electron/generator/helper-endpointconfig.js
brdandu: Created https://github.com/project-chip/zap/pull/1618/files

PR #1642, opened 2025-09-09
added template helper to get mandatory attribute count

dhchandw:
ZAPP-1659

### DIFF
--- a/src-electron/generator/helper-attribute.js
+++ b/src-electron/generator/helper-attribute.js
@@ -29,13 +29,13 @@ const dbEnum = require('../../src-shared/db-enum')
 
 /**
  * Counts the number of mandatory attributes (not optional, non-global) for the current cluster context.
- * Usage: {{count_mandatory_attributes side="server"}}
+ * Usage: {{count_mandatory_matter_attributes side="server"}}
  * If side is not provided or invalid, counts all attributes.
  */
-async function count_mandatory_attributes(options) {
+async function count_mandatory_matter_attributes(options) {
   if (!('id' in this))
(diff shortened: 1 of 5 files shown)


brdandu (changes requested):
Add unit tests for the helper.

src-electron/generator/helper-attribute.js
```diff
@@ -172,3 +206,4 @@ exports.global_attribute_default = attributeDefault
 exports.feature_bits = featureBits
 exports.as_underlying_atomic_identifier_for_attribute_id =
   as_underlying_atomic_identifier_for_attribute_id
+exports.count_mandatory_attributes = count_mandatory_attributes
```
brdandu: If this is Matter only then switch this to have matter in the name. count_mandatory_matter_attributes

src-electron/generator/helper-attribute.js
```diff
@@ -27,6 +27,40 @@ const templateUtil = require('./template-util')
...
 
+/**
+ * Counts the number of mandatory attributes (not optional, non-global) for the current cluster context.
+ * Usage: {{count_mandatory_attributes side="server"}}
+ * If side is not provided or invalid, counts all attributes.
+ */
+async function count_mandatory_attributes(options) {
+  if (!('id' in this))
+    throw new Error(
+      'count_mandatory_attributes requires an id inside the context.'
```
brdandu: Also add that this should be used under a cluster based block helper. The id has to be a reference to a cluster id?

brdandu (approved):

PR #1646, opened 2025-09-11
updating the xsd schema for zap

brdandu:
Github: ZAP#1609

### DIFF
--- a/zcl-builtin/shared/schema/zcl.xsd
+++ b/zcl-builtin/shared/schema/zcl.xsd
@@ -34,6 +34,13 @@ This schema describes the format of the XML files, that describe the ZCL specifi
     <xs:attribute name="type" use="required" type="xs:string"/>
     <xs:attribute name="array" use="optional" type="xs:boolean"/>
     <xs:attribute name="isNullable" use="optional" type="xs:boolean"/>
+    <xs:attribute name="length" use="optional" type="xs:string"/>
+    <xs:attribute name="optional" use="optional" type="xs:boolean"/>
+    <xs:attribute name="min" use="optional" type="xs:string"/>
+    <xs:attribute name="max" use="optional" type="xs:string"/>
+    <xs:attribute name="default" use="optional" type="xs:string"/>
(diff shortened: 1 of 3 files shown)


paulr34 (approved):

zcl-builtin/shared/schema/zcl.xsd
```diff
@@ -50,15 +57,24 @@ This schema describes the format of the XML files, that describe the ZCL specifi
...
+    <xs:attribute name="min" type="xs:string" use="optional"/>
+    <xs:attribute name="max" type="xs:string" use="optional"/>
+    <xs:attribute name="minLength" type="xs:integer" use="optional"/>
+    <xs:attribute name="apiMaturity" type="zclApiMaturity" use="optional"/>
   </xs:complexType>
   <xs:complexType name="clusterRef">
     <xs:attribute name="code" type="xs:string"/>
   </xs:complexType>
   <xs:complexType name="item">
+    <xs:sequence>
```
dhchandw: I think you can use a choice construct here since both tags can't exist at the same time

```
<xs:choice minOccurs="0">
  <xs:element ref="mandatoryConform"/>
  <xs:element ref="optionalConform"/>
</xs:choice>
```

zcl-builtin/shared/schema/zcl.xsd
```diff
@@ -536,4 +673,133 @@ This schema describes the format of the XML files, that describe the ZCL specifi
...
+      <xs:attribute name="more" type="xs:string" use="optional"/>
+      <xs:attribute name="min" type="xs:string" use="optional"/>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="features">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element name="feature" maxOccurs="unbounded">
+          <xs:complexType mixed="true">
+            <xs:sequence>
```
dhchandw: Not sure if sequence is the right construct here, can multiple of these tags be present at the same time.

Sequence also means that the validation will break if they're not in the order defined.

PR #1647, opened 2025-09-16
fix download-artifact script

dhchandw:
ZAPP-1662


.eslintrc.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   root: true,
 
+  ignorePatterns: ['src-script/download-artifact.ts'],
```
tecimovic: ?
Why are you ignoring this?

paulr34:
@dhchandw it says "fix download artifact script" but the script is deleted in your PR

dhchandw:
> I see there are two of the same script?

@paulr34 The typescript file is the script that my PR is talking about, the javascript file you see is just the transpiled file.

package.json
```diff
@@ -20,7 +20,7 @@
   "bin": "dist/src-electron/main-process/main.js",
   "scripts": {
     "download-artifact": "npx ts-node src-script/download-artifact.ts",
-    "pkg:download-artifact": "npx pkg src-script/download-artifact.js  -t node18-linux-x64,node18-linux-arm64,node18-macos-x64,node18-win-x64 --output dist/download-artifact --compress GZip",
+    "pkg:download-artifact": "npx pkg src-script/download-artifact.js  -t node16-linux-x64,node16-linux-arm64,node16-macos-x64,node16-win-x64 --output dist/download-artifact --compress GZip",
```
brdandu, line 23: Why is the node versions being downgraded? we should be upgrading if anything.

.gitignore
```diff
@@ -86,6 +86,9 @@ reports
 # clinic.js performance profiling data
 .clinic
 
+# Compiled JavaScript output from TypeScript source
+src-script/download-artifact.js
```
brdandu, line 90: Does this exist?

src-script/download-artifact.ts
```diff
@@ -389,19 +431,21 @@ async function artifactoryDownloadArtifacts(
...
   let refCommit: string | undefined = ''
   let refWorkflowRunId: number | undefined = 0
 
   const res = await octokit.request(
     'GET /repos/{owner}/{repo}/actions/artifacts',
     {
       owner,
-      repo
+      repo,
+      headers: { authorization: `token ${githubToken}` }
```
brdandu, line 448: Why was this change needed?

src-script/download-artifact.ts
```diff
@@ -324,7 +360,10 @@ async function getExistingGithubBranches(
 
   try {
     if (DEBUG) console.log(`GET: ${url}`)
-    let resp = await axios.get(url)
+    const headers = options.githubToken
+      ? { Authorization: `token ${options.githubToken}` }
+      : undefined
+    let resp = await axios.get(url, { headers })
```
brdandu, line 366: Why is this change needed?

dhchandw: This sends the Github token for the API if available, this call is made even if the Github token is not passed in (downloading from Artifactory)

src-script/download-artifact.ts
dhchandw: This sends the Github token as the Github API expects it

.gitignore
dhchandw: It will if somebody compiles the type script code, we don't want them to commit that.

package.json
dhchandw: Packaging with node18 is slowing down the performance for this particular artifact, maybe because of the packages used. We are phasing this out so didn't bother to dig deeper into this.

brdandu: Not really sue what slowing down performance means here. Educate me later on this.

src-script/download-artifact.ts
brdandu: Did the api change with no backwards compatibility?

package.json
ethanzhouyc: Give Node 20 a try too here

ethanzhouyc (approved):

PR #1648, opened 2025-09-17
fix: correct hex padding for as_generated_default_macro to fix 24-bit attribute defaults

brdandu:
- Ensure proper hex padding based on attribute size (attributeSize * 2 digits)
- Fix issue where 0x186A0 for 24-bit attributes generated incorrect byte sequence
- Handle both hex (0x prefixed) and decimal input values correctly
- Maintain backward compatibility with existing test expectations
- Fixes customer-reported issue where 24-bit values were 16x larger than expected
- The padding logic now ensures that hex values like 0x186A0 are properly padded to 0x0186A0 for 3-byte attributes, generating the correct byte sequence 0x01, 0x86, 0xA0 instead of 0x18, 0x6A, 0x0.
- JIRA: ZAPP-1658

### DIFF
--- a/test/gen-zigbee-3.test.js
+++ b/test/gen-zigbee-3.test.js
@@ -133,7 +133,7 @@ test(
     let cfgVer2 = genResult.content['zap-config-version-2.h']
     // Test GENERATED_DEFAULTS
     expect(cfgVer2).toContain(
-      '0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,  /* 0,DEFAULT value for cluster: Over the Air Bootloading, attribute: OTA Upgrade Server ID, side: client*/'
+      '0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, /* 0,DEFAULT value for cluster: Over the Air Bootloading, attribute: OTA Upgrade Server ID, side: client*/'
     )
     // Test GENERATED_ATTRIBUTE_COUNT
     expect(cfgVer2).toContain('#define GENERATED_ATTRIBUTE_COUNT 81')
--- a/test/helpers.test.js
+++ b/test/helpers.test.js
@@ -241,7 +241,7 @@ test(
     return zclHelper
       .as_generated_default_macro('0x003840', 3, options)
       .then((res) => {
-        return expect(res).toBe('0x40, 0x38, 0x00, ')
+        return expect(res).toBe('0x40, 0x38, 0x00,')
       })
   },
   testUtil.timeout.short()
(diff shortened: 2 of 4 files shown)


test/gen-zigbee-3.test.js
```diff
@@ -133,7 +133,7 @@ test(
     let cfgVer2 = genResult.content['zap-config-version-2.h']
     // Test GENERATED_DEFAULTS
     expect(cfgVer2).toContain(
-      '0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,  /* 0,DEFAULT value for cluster: Over the Air Bootloading, attribute: OTA Upgrade Server ID, side: client*/'
+      '0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, /* 0,DEFAULT value for cluster: Over the Air Bootloading, attribute: OTA Upgrade Server ID, side: client*/'
     )
```
ethanzhouyc, line 137: Can you help me understand why we need to add paddings here in general?

test/helpers.test.js
```diff
@@ -376,19 +376,41 @@ test(
...
+test(
+  'Generated Macro for 3 byte integer big endian with padding',
+  () => {
+    let options = { hash: { endian: 'big' } }
+    return zclHelper
+      .as_generated_default_macro('0x0186A0', 3, options)
+      .then((res) => expect(res).toBe('0x01, 0x86, 0xA0,'))
+  },
+  testUtil.timeout.short()
+)
```
ethanzhouyc, line 416: Add test for little endian too if possible?

test/helpers.test.js
```diff
@@ -459,7 +481,7 @@ test(
     return zclHelper
       .as_generated_default_macro('140737488355327', 8, options)
       .then((res) =>
-        expect(res).toBe('0x00, 0x00,  0x7F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,')
+        expect(res).toBe('0x00, 0x00, 0x7F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,')
```
ethanzhouyc, line 495: Why we are having paddings in the middle of the byte list before? Will remove this causing any issues?

test/gen-zigbee-3.test.js
brdandu: padding is for making sure the bytes are read appropriately. It is adding 0s to make sure the hex values are even numbered for eg if it is a 3 byte data type then it 0x0000A should be converted to 0x00000A. The padding is not for the spaces in the change here. This is just a result of the trim() function applied in the changes

test/helpers.test.js
brdandu: There were some additional spaces added after which was causing this. Now those went away.

brdandu: Should not cause any issues that I know off.

test/helpers.test.js
brdandu: added

tecimovic (approved):

Issue #1204, opened 2023-11-08
Make XML errors easier to debug during codegen


andy31415:
We had a type error recently in two separate reasons: `uint16` used instead of `int16u` and once a type being `""` (not sure why ... maybe an erase that was never updated).

The code generator failed with very hard to debug errors like:

```
TypeError: Cannot read properties of undefined (reading 'startsWith')
    at Object.atomicType (/snapshot/zap/dist/src-electron/generator/overridable.js:63:14)
    at Object.atomicType (/snapshot/zap/dist/src-electron/generator/template-engine.js:175:31)
    at /snapshot/zap/dist/src-electron/util/zcl-util.js:545:47
    at async Promise.all (index 0)
```

Instead we should have some context on various things, in this case for example atomicType should at least tell me what type it was trying to process, so I could have seen `uint16` being at fault.

Ideally though, we should also have XML context inside types. I would have liked the trace to tell me "in foo.xml, type defined at line 123 failed to be processed" and then human review would have been much easier. As it stands we had to bisect all types one by one in XML which is very slow.

paulr34:
@andy31415 we have XML type warnings like this 
<img width="1003" alt="Screenshot 2023-11-09 at 12 49 01 PM" src="https://github.com/project-chip/zap/assets/64710345/9902ed03-5c14-4f2f-8a45-5b22df6d0ad9">

paulr34:
also if you click extensions you can see 
<img width="798" alt="Screenshot 2023-11-09 at 12 50 56 PM" src="https://github.com/project-chip/zap/assets/64710345/96235e7e-157b-4e58-9aee-a0e71ced9cd0">

andy31415:
Checked offline, the existing warnings did not catch the errors in the XML I had (neither empty type nor using `uint16` instead of `int16u`)

PR #1651, opened 2025-09-23
Adding error handling for the atomic types under overridable

brdandu:
Github: ZAP #1204

### DIFF
--- a/src-electron/generator/overridable.js
+++ b/src-electron/generator/overridable.js
@@ -21,6 +21,8 @@
  * @module Templating API: Overridable functions.
  */
 
+const env = require('../util/env')
+
 // Local utility function
 
 /**
@@ -69,6 +71,15 @@ function atomicType(arg = { name: 'unknown', size: 0, no_warning: 0 }) {
   let name = arg.name
   let size = arg.size
   let no_warning = arg.no_warning
+  if (!name || typeof name !== 'string') {
+    env.logError(`atomicType failed for type: Invalid or empty type name.`)
+  }
+  // Regex to catch invalid Zigbee type names like uint16, int16, etc.
+  if (/^(u?int\d+)$/i.test(name)) {
+    env.logError(
+      `atomicType failed for type: Invalid Zigbee type name ${name}.`
+    )
+  }
   if (name.startsWith('int')) {
     let signed
     if (name.endsWith('s')) signed = true


src-electron/generator/overridable.js
```diff
@@ -66,49 +66,62 @@ function nonAtomicType(arg = { name: 'unknown', isStruct: false }) {
...
-  if (name.startsWith('int')) {
-    let signed
-    if (name.endsWith('s')) signed = true
-    else signed = false
+  try {
+    let name = arg.name
+    let size = arg.size
+    let no_warning = arg.no_warning
+    if (!name || typeof name !== 'string') {
+      throw new Error(`Invalid or empty type name ${name}.`)
```
ethanzhouyc: Will throwing error here block user from using ZAP normally? Should we put the error into notification system?

src-electron/generator/overridable.js
```diff
@@ -66,49 +66,62 @@ function nonAtomicType(arg = { name: 'unknown', isStruct: false }) {
...
+    }
+    // Regex to catch invalid Zigbee type names like uint16, int16, etc.
+    if (/^(u?int\d+)$/i.test(name)) {
+      throw new Error(
+        `Invalid Zigbee type name ${name}. Did you mean "int16u", "int16s", "int32u", etc.?`
+      )
```
ethanzhouyc: If the only types that trigger error are 'unit8', 'uint16', 'uint32', could use a mapping function instead of regex. Also the error message could be more specific. If it catches 'uint16', then the error tells user that it should be 'int16u'

test/helpers.test.js
```diff
@@ -634,3 +635,47 @@ test(
...
+      type: '',
+      side: 'server',
+      min: '0x0000',
+      max: '0xFFFF',
+      writable: false,
+      optional: true,
+      xmlFile: 'zcl-builtin/matter/data-model/chip/color-control-cluster.xml'
+    }
+    expect(() => atomicType(attribute)).toThrow(
+      /atomicType failed for type : Invalid Zigbee type name uint16. Did you mean "int16u", "int16s", "int32u", etc.?/
```
ethanzhouyc: atomicType failed for: Invalid Zigbee type name unit16
atomicType failed for: invalid or empty type name

Does this sound more clear?

brdandu: Sounds good

src-electron/generator/overridable.js
brdandu: This is generation time so throwing errors into notification system is very different. Would not stretch it that much here. It may not be a bad idea to get another feature to show all generation time warnings somewhere

brdandu: For now put the errors into log file

PR #1658, opened 2025-10-21
Support device type revisions

andy31415:
This adds support for a `deviceRevision` element in `matter-devices.xml` and the ability to access it as `{{revision}}` inside zap templates.

### DIFF
--- a/zcl-builtin/shared/schema/zcl.xsd
+++ b/zcl-builtin/shared/schema/zcl.xsd
@@ -161,7 +161,7 @@ This schema describes the format of the XML files, that describe the ZCL specifi
         <xs:element ref="deviceId"/>
         <xs:element minOccurs="0" ref="channels"/>
         <xs:element ref="clusters"/>
-        <xs:element minOccurs="0" ref="deviceRevision"/>
+        <xs:element minOccurs="0" ref="revision"/>
       </xs:sequence>
     </xs:complexType>
   </xs:element>
@@ -206,7 +206,7 @@ This schema describes the format of the XML files, that describe the ZCL specifi
       </xs:restriction>
     </xs:simpleType>
   </xs:element>
-  <xs:element name="deviceRevision">
+  <xs:element name="revision">
     <xs:simpleType>
       <xs:restriction base="xs:integer">
         <xs:minInclusive value="1"/>
--- a/src-electron/db/query-loader.js
+++ b/src-electron/db/query-loader.js
@@ -1268,7 +1268,7 @@ async function insertDeviceTypes(db, packageId, data) {
           dt.class,
           dt.scope,
           dt.superset,
-          dt.deviceRevision
+          dt.revision
         ]
       })
     )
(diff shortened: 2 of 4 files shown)


paulr34 (approved):

src-electron/db/zap-schema.sql
```diff
@@ -368,6 +368,7 @@ CREATE TABLE IF NOT EXISTS "GLOBAL_ATTRIBUTE_BIT" (
 DROP TABLE IF EXISTS "DEVICE_TYPE";
 CREATE TABLE IF NOT EXISTS "DEVICE_TYPE" (
   "DEVICE_TYPE_ID" integer primary key autoincrement,
+  "DEVICE_REVISION" integer,
```
brdandu: See that the cluster table just says "REVISION". Would be good to keep it consistent for this table as well.

zcl-builtin/shared/schema/zcl.xsd
```diff
@@ -161,6 +161,7 @@ This schema describes the format of the XML files, that describe the ZCL specifi
         <xs:element ref="deviceId"/>
         <xs:element minOccurs="0" ref="channels"/>
         <xs:element ref="clusters"/>
+        <xs:element minOccurs="0" ref="deviceRevision"/>
```
brdandu: Will this be revision or deviceRevision?

andy31415: I used deviceRevision because we use `deviceId`. I have no preference ... is there some standard?

andy31415: Alchemy seems to have used `revision` in https://github.com/project-chip/alchemy/commit/e28398114d7bee09b7a4e2d69a036985c04e6126 .... if that is the case we probably want `revision`

andy31415: Never mind, code is:

```
xml.SetOrCreateSimpleElement(dte, "deviceRevision", mostRecentRevision.Number.IntString(), "name", "domain", "typeName", "profileId", "deviceId").CreateAttr("editable", "false")
```

So it creates deviceRevision. In that case I prefer to keep as is.

andy31415: Updated to `revision` and alchemy also matched now.

test/gen-template/matter/device-types.zapt
```diff
@@ -15,4 +15,3 @@ property="testDeviceTypeExtension"}}'
 
 
 {{/zcl_device_types}}
```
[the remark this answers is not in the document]
andy31415: This seems nonsense. I believe this was auto-formatter doing it, howevery in any case there is a trailing newline so removing the blank line does not seem wrong.

zcl-builtin/shared/schema/zcl.xsd
```diff
@@ -205,6 +206,15 @@ This schema describes the format of the XML files, that describe the ZCL specifi
...
   </xs:element>
+  <xs:element name="revision">
+    <xs:simpleType>
+      <xs:restriction base="xs:integer">
+        <xs:minInclusive value="1"/>
+        <xs:maxInclusive value="255"/>
+        <xs:whiteSpace value="collapse"/>
+      </xs:restriction>
+    </xs:simpleType>
+  </xs:element>
```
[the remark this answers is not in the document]
andy31415: I am doing the "when in rome do like the romans do" ... none of the existing declarations have xs:annotation here.

PR #1671, opened 2025-11-17
Show package warnings when incorrect attribute default values are present in xml files

brdandu:
- validation is called along with insertAttributes call now
- JIRA: ZAPP-1648

### DIFF
--- a/test/validation.test.js
+++ b/test/validation.test.js
@@ -32,6 +32,7 @@ const testQuery = require('./test-query')
 const env = require('../src-electron/util/env')
 const types = require('../src-electron/util/types')
 const { timeout } = require('./test-util')
+const queryPackageNotification = require('../src-electron/db/query-package-notification')
 
 let db
 let sid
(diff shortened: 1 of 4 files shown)


test/validation.test.js
```diff
@@ -551,6 +552,23 @@ test(
...
 
+test(
+  'validateXmlAttributeDefault - integer out of range',
+  async () => {
+    // Check that notification was created
+    let notifications =
+      await queryPackageNotification.getNotificationByPackageId(db, pkgId)
+    let xmlValidationNotifs = notifications.filter((n) =>
+      n.message.includes(
+        'XML validation issues for attribute "active power max phase b" (type: int16s, defaultvalue: 0xffff): Out of range (min: -32768, max: 32767)'
```
dhchandw, line 563: Would it possible to tell the user which endpoint has this issue?

brdandu: These are package errors. When editing the endpoint attributes they will see those in the UI itself for incorrect values

PR #1677, opened 2026-01-08
Add aarch64 paths to `apack.json` for Linux and Windows

puddly:
`zap-linux-arm64.zip` and `zap-windows-arm64.zip` are created as release artifacts but they aren't usable by tooling reading `apack.json` (e.g. `slc-cli`) on ARM64 platforms. This modification allows them to be picked up automatically.

### DIFF
--- a/apack.json
+++ b/apack.json
@@ -12,6 +12,10 @@
       "exe": "zap.exe",
       "optional": true
     },
+    "zap:win32.aarch64": {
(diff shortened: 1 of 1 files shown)


brdandu:
> I can't test ARM windows but I assume it'll work too. Want me to roll those into this PR?

That would be great. Thank you!

dhchandw (approved):

PR #1685, opened 2026-02-11
ZAPP-1695: Enable toggling desc features to fix the TagList feature issue

ethanzhouyc:
Before, ZAP disabled toggling desc features as the complex descriptive conformance might lead to wrong cluster state.

But we should not restrict users. Since we have issues on unable to toggle the TagList feature, now we enable toggling desc features and will display a warning instead.

See ZAPP-1695 for issue details.

### DIFF
--- a/test/feature.test.js
+++ b/test/feature.test.js
@@ -768,6 +768,56 @@ test(
 )
 
 test(
+  'Check feature with desc conformance allows toggling but shows warning',
+  () => {
+    // Test enabling a feature with 'desc' conformance
+    let elements = {
+      attributes: [],
+      commands: [],
+      events: []
+    }
+    let featureMap = {
+      DESCFEATURE: false
+    }
+    let featureWithDesc = {
+      cluster: 'Test Cluster',
+      name: 'Feature With Desc',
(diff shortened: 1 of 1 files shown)


src-electron/validation/conformance-checker.js
```diff
@@ -203,6 +190,17 @@ function generateWarningMessage(
...
     }
+    // if the feature conformance contains the operand 'desc', do not disable toggling, but show warning message
+    let featureContainsDesc = conformEvaluator.checkIfExpressionHasOperand(
+      featureData.conformance,
+      dbEnum.conformanceTag.described
+    )
+    if (featureContainsDesc) {
+      result.warningMessage =
+        warningPrefix +
+        ` is being ${added ? 'enabled' : 'disabled'}, but it has descriptive conformance and is too complex for ZAP to process.`
```
brdandu: but it has descriptive conformance and is too complex for ZAP to process. -> but it has descriptive conformance and requires manual validation from the feature specification to enable/disable the right dependencies in ZAP.

brdandu (approved):

PR #1687, opened 2026-02-17
parallelize generation and notification paths with Promise.all

brdandu:
- Template engine: run iterative iterations (e.g. per-cluster) and deferred blocks in parallel; use per-iteration options to avoid races.
- Generation engine: load template JSON files and package options in parallel; run generateAndWriteFiles for multiple template packages in parallel per .zap file.
- ClustersHelper: fetch cluster details per endpoint in parallel in loadEndpoints().
- helper-endpointconfig: resolve attribute default values in parallel per cluster, then process in order.
- query-session-notification: run deleteNotification and setWarningIfMessageNotExists in parallel where multiple IDs/messages are handled.
- Github ZAP#1659

### DIFF
--- a/src-electron/main-process/startup.js
+++ b/src-electron/main-process/startup.js
@@ -907,22 +907,20 @@ async function generateSingleFile(
   if (usedTemplatePackageIds.length === 0) {
     usedTemplatePackageIds = [templatePackageId]
   }
-  const genResults = await Promise.all(
-    usedTemplatePackageIds.map((templatePackageId) =>
-      generatorEngine.generateAndWriteFiles(
-        db,
-        sessionId,
-        templatePackageId,
-        output,
-        options
-      )
+  let genResults = []
+  for (let i = 0; i < usedTemplatePackageIds.length; i++) {
+    let genResult = await generatorEngine.generateAndWriteFiles(
+      db,
+      sessionId,
+      usedTemplatePackageIds[i],
+      output,
+      options
     )
-  )
-  for (const genResult of genResults) {
     if (genResult.hasErrors) {
       console.log(JSON.stringify(genResult.errors))
       throw new Error(`Generation failed: ${zapFile}`)
     }
+    genResults.push(genResult)
   }
 
   if (options.upgradeZapFile && isZapFileUpgradeNeeded) {


src-electron/generator/generation-engine.js
```diff
@@ -596,9 +596,13 @@ async function loadTemplates(
...
     if (genTemplatesJsonArray != null && genTemplatesJsonArray.length > 0) {
-      for (let jsonFile of genTemplatesJsonArray) {
-        if (jsonFile == null || jsonFile == '') continue
-        let ctx = await loadGenTemplatesJsonFile(db, jsonFile)
+      const filesToLoad = genTemplatesJsonArray.filter(
+        (f) => f != null && f !== ''
+      )
+      const loadResults = await Promise.all(
+        filesToLoad.map((jsonFile) => loadGenTemplatesJsonFile(db, jsonFile))
+      )
```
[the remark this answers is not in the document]
brdandu: Reverting

src-electron/generator/helper-endpointconfig.js
```diff
@@ -953,22 +953,30 @@ async function collectAttributes(
...
+          determineAttributeDefaultValue(
+            a.defaultValue,
+            a.type,
+            a.typeSize,
+            a.isNullable,
+            db,
+            sessionId
+          )
+        )
+      )
```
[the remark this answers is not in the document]
brdandu: Not doing this.

src-electron/db/query-session-notification.js
```diff
@@ -288,13 +283,8 @@ async function deleteNotificationWithPatterns(db, sessionId, patterns) {
...
   if (rows && rows.length > 0) {
-    let ids = rows.map((row) => row.NOTICE_ID)
-    let deleteResponses = []
-    for (let id of ids) {
-      let response = await deleteNotification(db, id)
-      deleteResponses.push(response)
-    }
-    return deleteResponses
+    const ids = rows.map((row) => row.NOTICE_ID)
+    return Promise.all(ids.map((id) => deleteNotification(db, id)))
```
ethanzhouyc, line 287: These changes are good practices, but I don't think they will improve performance a lot as the functions are only used in rare feature change cases

src-electron/generator/generation-engine.js
```diff
@@ -23,16 +23,87 @@ const _ = require('lodash')
...
+const workerpool = require('workerpool')
 const dbApi = require('../db/db-api.js')
 const dbCache = require('../db/db-cache.js')
 const queryNotification = require('../db/query-package-notification.js')
 
+const ITERATION_WORKER_SCRIPT = path.join(
+  __dirname,
+  'template-iteration-worker.js'
+)
+const MAX_WORKERS = 8 //cap at 8 workers
```
ethanzhouyc, line 42: Will fine-tuning the number of workers here improve the performance?

brdandu: Is system and use case dependent. No fixed amount can give optimized performance for all use cases. Depends on how big the task is for the worker performance and how much time it takes to allocate work for the workers

src-electron/db/query-session-notification.js
brdandu: Correct, just good practice. We are not getting benefits with this. The main benefit here is the multithreading for handlebar template rendering

brdandu:
> package-lock changes?

Those came from workerpool package of npm which is used to do multi-threading

Issue #1683, opened 2026-02-10
Event args in XML should use "fieldId" for its identifier


hasty:
The rest of the ZAP XML uses "fieldId" to specify the identifier for each field, aside from events.

PR #1688, opened 2026-02-17
Use fieldId for event fields; keep loader backwards compatible

dhchandw:
- Rename event <field> attribute from id to fieldId in all ZCL/Matter XML
  to match zcl.xsd (eventField requires fieldId).
- In zcl-loader-silabs.js, read fieldId with fallback to id so existing
  XML using id still loads.

### DIFF
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -587,7 +587,8 @@ function prepareCluster(cluster, context, isExtension = false) {
         let lastFieldId = -1
         event.field.forEach((field) => {
           let defaultFieldId = lastFieldId + 1
-          lastFieldId = field.$.id ? parseInt(field.$.id) : defaultFieldId
+          const rawFieldId = field.$.fieldId ?? field.$.id
+          lastFieldId = rawFieldId ? defaultFieldId : parseInt(rawFieldId)
           if (field.$.removedIn == null) {
             ev.fields.push({
               name: field.$.name,
--- a/test/resource/old-matter/access-control-cluster.xml
+++ b/test/resource/old-matter/access-control-cluster.xml
@@ -105,20 +105,20 @@ limitations under the License.
 
     <event side="server" code="0x0000" name="AccessControlEntryChanged" priority="info" optional="false">
       <description>The cluster SHALL send AccessControlEntryChanged events whenever its ACL attribute data is changed by an Administrator.</description>
-      <field id="1" name="AdminNodeID" type="node_id" isNullable="true"/>
-      <field id="2" name="AdminPasscodeID" type="INT16U" isNullable="true"/>
-      <field id="3" name="ChangeType" type="ChangeTypeEnum"/>
-      <field id="4" name="LatestValue" type="AccessControlEntry" isNullable="true"/>
-      <field id="0xFE" name="AdminFabricIndex" type="fabric_idx"/>
+      <field fieldId="1" name="AdminNodeID" type="node_id" isNullable="true"/>
+      <field fieldId="2" name="AdminPasscodeID" type="INT16U" isNullable="true"/>
+      <field fieldId="3" name="ChangeType" type="ChangeTypeEnum"/>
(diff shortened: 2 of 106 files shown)


test/resource/old-matter/access-control-cluster.xml
```diff
@@ -105,20 +105,20 @@ limitations under the License.
 
     <event side="server" code="0x0000" name="AccessControlEntryChanged" priority="info" optional="false">
       <description>The cluster SHALL send AccessControlEntryChanged events whenever its ACL attribute data is changed by an Administrator.</description>
-      <field id="1" name="AdminNodeID" type="node_id" isNullable="true"/>
-      <field id="2" name="AdminPasscodeID" type="INT16U" isNullable="true"/>
-      <field id="3" name="ChangeType" type="ChangeTypeEnum"/>
-      <field id="4" name="LatestValue" type="AccessControlEntry" isNullable="true"/>
-      <field id="0xFE" name="AdminFabricIndex" type="fabric_idx"/>
+      <field fieldId="1" name="AdminNodeID" type="node_id" isNullable="true"/>
```
paulr34, line 108: this looks redundant at first glance

dhchandw: These are updated so it adheres to the updated schema

test/resource/old-matter/basic-information-cluster.xml
```diff
@@ -68,7 +68,7 @@ limitations under the License.
 
     <event side="server" code="0x00" name="StartUp" priority="critical" optional="false">
       <description>The StartUp event SHALL be emitted by a Node as soon as reasonable after completing a boot or reboot process.</description>
-      <field id="0" name="SoftwareVersion" type="INT32U"/>
+      <field fieldId="0" name="SoftwareVersion" type="INT32U"/>
```
paulr34, line 71: Since this is already a <field> element, is there a specific reason for renaming id to fieldId? It seems like id was already clearly referring to the field identifier.

dhchandw: This is to adhere to the Matter naming, Alchemy uses the XSD file to generate the ZAP file. Look at #1683

test/resource/old-matter/access-control-cluster.xml
paulr34: I see, its a little awkward but everything you did looks correct based on the ask so ill approve

test/resource/old-matter/basic-information-cluster.xml
paulr34: yeah, again, I see what you did. I don't really understand why it needs to be so explicit instead of obviously derived but everything you were asked to do looks correct so ill approve

test/resource/old-matter/access-control-cluster.xml
brdandu: Don't think we need to update all the XML. Just update one XML and make sure it works for both kinds of changes. The XMLs in the ZAP repo are just used for texting so we need not have this big diff on XML

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -587,7 +587,9 @@ function prepareCluster(cluster, context, isExtension = false) {
         let lastFieldId = -1
         event.field.forEach((field) => {
           let defaultFieldId = lastFieldId + 1
-          lastFieldId = field.$.id ? parseInt(field.$.id) : defaultFieldId
+          const rawFieldId = field.$.fieldId ?? field.$.id
+          lastFieldId =
+            rawFieldId == null ? defaultFieldId : parseInt(rawFieldId)
```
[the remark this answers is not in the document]
brdandu: Should you be taking this in?

brdandu (changes requested):
Add tests for the loader which loads either of those XML variations. Get rid of all the xml changes.

dhchandw:
@brdandu @paulr34 The reason I changed all the internal XML files is so they follow the `zcl.xsd` schema. 

Alchemy uses this schema for the ZAP XML generation and I don't want to leave it vague (i.e. allowing both id and fieldId for eventField)

I can revert the internal XML changes if that's preferred.

PR #1694, opened 2026-03-03
Fix some ui issues

tbrkollar:


brdandu (changes requested):
add open ALL and close all cypress tests while changing the different filter options(enabled, all clusters, etc) and make sure nothing breaks

cypress/e2e/clusters/cluster-filter.cy.js
```diff
@@ -33,6 +33,50 @@ describe('Testing cluster filters', () => {
...
+        cy.get('tbody').children().contains(data.cluster10).should('exist')
+      })
+    }
+  )
+  it(
+    'filter all clusters and check close all and open all buttons',
+    { retries: { runMode: 2, openMode: 2 } },
+    () => {
+      cy.get('[data-test="filter-input"]').click()
+      // Selecting All clusters
```
brdandu, line 62: Selecting all clusters filter option

cypress/e2e/clusters/cluster-filter.cy.js
```diff
@@ -33,6 +33,50 @@ describe('Testing cluster filters', () => {
...
+      cy.fixture('data').then((data) => {
+        cy.get('tbody').children().contains(data.cluster9).should('exist')
+        cy.get('tbody').children().contains(data.cluster10).should('exist')
+      })
+
+      // Close All: doActionFilter sets allDomainsCollapsed = true,
+      // hiding domains and swapping to Open All button
+      cy.dataCy('cluster-btn-closeall').click()
+      cy.dataCy('cluster-btn-closeall').should('not.exist')
+      cy.dataCy('cluster-btn-openall').should('exist')
```
brdandu, line 75: Check if the clusters do not exist after close all? (data.cluster10 and data.cluster9)

cypress/e2e/clusters/cluster-filter.cy.js
```diff
@@ -33,6 +33,50 @@ describe('Testing cluster filters', () => {
...
+
+      // Open All: doActionFilter sets allDomainsCollapsed = false,
+      // restoring domains and swapping back to Close All button
+      cy.dataCy('cluster-btn-openall').click()
+      cy.dataCy('cluster-btn-openall').should('not.exist')
+      cy.dataCy('cluster-btn-closeall').should('exist')
+
+      cy.fixture('data').then((data) => {
+        cy.get('tbody').children().contains(data.cluster9).should('exist')
+        cy.get('tbody').children().contains(data.cluster10).should('exist')
```
brdandu: What about the not.exist when close all button is clicked?

cypress/e2e/clusters/cluster-filter.cy.js
```diff
@@ -33,6 +33,50 @@ describe('Testing cluster filters', () => {
...
         cy.get('tbody').children().contains(data.cluster10).should('exist')
       })
+
+      cy.dataCy('cluster-btn-closeall').click()
+      cy.dataCy('cluster-btn-openall').click()
+
+      cy.fixture('data').then((data) => {
+        // Power Configurator for Zigbee is disabled and should not show up
+        // Occupancy Sensing for Matter is disabled and should not exist
+        cy.get('tbody').children().contains(data.cluster9).should('not.exist')
```
brdandu, line 50: not exist?

tbrkollar: Please open the file and check it again. Github ui is not so good. 
you can see above in the code (before open and close) that this value does not exist.

cypress/e2e/clusters/cluster-filter.cy.js
tbrkollar: Github ui is not so good. line 56, 
 cy.get('.q-virtual-scroll__content > :nth-child(2)').click({
        force: true
      })
It is the "All Clusters" option.

cypress/e2e/clusters/cluster-filter.cy.js
tbrkollar: you are right. i add a not-visible check.

cypress/e2e/clusters/cluster-filter.cy.js
tbrkollar: you are right. i add a not-visible check.

brdandu (approved):

cypress/e2e/clusters/cluster-filter.cy.js
paulr34: @tbrkollar I am also confused by this line of code `should('not.exist')`

paulr34: nvm I see the comment clearly explains it `        // Power Configurator for Zigbee is disabled and should not show up
        // Occupancy Sensing for Matter is disabled and should not exist`

paulr34 (approved):

PR #1696, opened 2026-03-10
Disable feature toggle when featureMap attribute has external storage

ethanzhouyc:
When the featureMap attribute of a cluster has External storage, ZAP has no control over its value. Toggling the feature will cause mismatch between UI and database.

To solve this, this PR:
* Disable feature toggle when featureMap attribute has external storage, and display warning to user
* Added warning if user toggles a feature with external attribute depending on it
* Added warning if user updates an attribute to external, and removes related warning if they switch back
* Add unit tests


paulr34 (approved):

src-electron/validation/conformance-checker.js
```diff
@@ -72,6 +73,15 @@ function generateWarningMessage(
 
   let updateDisabledString = `cannot be ${added ? 'enabled' : 'disabled'} as`
 
+  // Check 0: if the featureMap attribute storage is external, ZAP cannot modify it
+  if (featureMapStorageOption === dbEnum.storageOption.external) {
+    result.warningMessage.push(
+      warningPrefix +
+        ` ${updateDisabledString} the featureMap attribute in the cluster is external and ZAP does not have control over it.`
+    )
```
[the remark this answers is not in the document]
paulr34: @ethanzhouyc we need to look into this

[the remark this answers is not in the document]
ethanzhouyc: Fixed by removing the `html: true` attribute here: https://github.com/project-chip/zap/pull/1696/changes#diff-58bf86e4d537d00e6d61f4e651879caa5d9adfe22853891f8bb7b87bbb188c5dL297

All warnings returned to Notify in this case are plain text so we do not need html here.

test/feature.test.js
```diff
@@ -763,6 +763,53 @@ test(
...
+      featureMap,
+      featureHS,
+      endpointId,
+      clusterFeatures,
+      dbEnum.storageOption.external
+    )
+    expectedWarning =
+      warningPrefix +
+      `feature: ${featureHS.name} (${featureHS.code}) ${featureBitMessage} cannot be enabled as ` +
+      `the featureMap attribute in the cluster is external and ZAP does not have control over it.`
```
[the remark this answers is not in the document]
ethanzhouyc: Fixed

ethanzhouyc:
@brdandu pointed out that if user toggles a feature with external attributes conforming to it, in this case, ZAP cannot handle updating the value of the attributes, and we should throw a warning.

Addressed this issue in the latest ocmmits

src-electron/db/query-session-notification.js
```diff
@@ -212,10 +212,12 @@ async function setNotificationOnFeatureChange(db, sessionId, result) {
     }
   } else {
     await deleteNotificationWithPatterns(db, sessionId, outdatedWarningPatterns)
-    if (displayWarning) {
-      await setNotification(db, 'WARNING', warningMessage, sessionId, 2, 0)
-    } else {
-      await searchNotificationByMessageAndDelete(db, sessionId, warningMessage)
+    for (const message of warningMessage) {
```
brdandu, line 216: Why do we need a for loop addition here and did not need it before?

src-electron/rest/user-data.js
```diff
@@ -115,6 +115,7 @@ function httpPostCheckConformOnFeatureUpdate(db) {
       clusterFeatures,
       endpointId,
       endpointTypeId,
+      featureMapStorageOption,
```
brdandu: assuming this is the storage option: NVM, RAM, External. Naming it `storageOption` makes it more readable.

src-electron/validation/conformance-checker.js
```diff
@@ -50,7 +50,9 @@ function generateWarningMessage(
   elementMap = {},
   descElements = {},
   featuresToUpdate = {},
-  changedConformFeatures = []
+  changedConformFeatures = [],
+  featureMapStorageOption = null,
```
brdandu: storageOption

src-electron/validation/conformance-checker.js
```diff
@@ -72,6 +74,15 @@ function generateWarningMessage(
 
   let updateDisabledString = `cannot be ${added ? 'enabled' : 'disabled'} as`
 
+  // Check 0: if the featureMap attribute storage is external, ZAP cannot modify it
+  if (featureMapStorageOption === dbEnum.storageOption.external) {
+    result.warningMessage.push(
+      warningPrefix +
```
brdandu, line 80: Could you give a user use case on when this would show up in the notifications pane?

src/util/feature-mixin.js
```diff
@@ -293,8 +302,7 @@ export default {
           message: warning,
           type: 'warning',
           classes: 'custom-notification notification-warning',
-          position: 'top',
-          html: true
```
brdandu, line 297: What was html: true doing?

brdandu (changes requested):

src-electron/db/query-session-notification.js
ethanzhouyc: Previously there was always exactly one warning per feature toggle, if the toggle is enabled, so a loop wasn't needed. The external attribute check introduced the possibility of multiple warnings, so the loop was added to handle them all uniformly.

ethanzhouyc: For example, if a mandatory feature with external attributes associated is disabled, there will be 2 warnings

src-electron/rest/user-data.js
ethanzhouyc: Renamed

src/util/feature-mixin.js
ethanzhouyc: It allows the message string to be rendered as HTML instead of plain text. 
So we can put `<b>bold</b>`, `<br>` in the warning message. But all messages we display are plain text so it's not needed.

I removed this only because the AI review said adding it would cause security risk (possibly leading to a SonarQube issue)

src-electron/validation/conformance-checker.js
ethanzhouyc: Toggling a feature with external attributes conforming to it, would trigger the warning.

Here is the real example I use:
1. create endpoint on Extended Color Light
2. go to color control cluster
3. enable currentHue attribute, change storage from RAM to External, and then disable it
4. toggle the HS feature

As currentHue has mandatory conformance on HS, this would trigger the warning:

> ⚠️ Check Feature Compliance on endpoint: 1, cluster: Color Control, attribute CurrentHue, required by feature: Hue And Saturation (HS), has external storage and ZAP does not have control over it.

brdandu: What about existing applications which have this problem?

PR #1699, opened 2026-04-22
Add getApiMaturityFromXML() in conformance-xml-parser:

### LINKED ISSUE #1664: Have zap resolve conflict resolution between conformance and apiMaturity
Currently apiMaturity in Clusters
Data types (structs, enums, bitmaps)
Commands
Attributes
Events
Command fields
Event fields
Struct fields
Enum values
Bitmap values can be contradictory to conformance in the xml. Figure out a way to use conformance if present and if not then use apiMaturity in zap code.

Talk to @andy31415 , @bzbarsky-apple , @hasty

brdandu:
if apiMaturity is set on a data model entity, use it; otherwise treat conformance as provisional when the parsed expression includes provisional 
Use it from zcl-loader-silabs for clusters, commands, args, attributes, events, event fields, enums, enum items, bitmaps, bitmap fields, features, structs, and struct items.
Extend test/resource/meta/api_maturity.xml and explicit apiMaturity winning over conflicting conformance.
- Github: ZAP #1664

### DIFF
--- a/src-electron/validation/conformance-xml-parser.js
+++ b/src-electron/validation/conformance-xml-parser.js
@@ -194,5 +194,34 @@ function getOptionalAttributeFromXML(element, elementType) {
   }
 }
 
+/**
+ * Determine the effective apiMaturity for an Data Model XML element.
+ *
+ * If the element has an explicit 'apiMaturity' attribute, return it as-is.
+ * Otherwise, if the element's conformance contains a <provisionalConform/>
+ * (including nested conformance tags), return 'provisional'.
+ * Return null when neither source specifies maturity.
+ *
+ * @param {*} element XML element from xml2js
+ * @returns {string|null}
+ */
+function getApiMaturityFromXML(element) {
+  if (element && element.$ && element.$.apiMaturity) {
+    return element.$.apiMaturity
+  }
+  let conformance = parseConformanceFromXML(element)
+  if (
+    conformance &&
+    conformEvaluator.checkIfExpressionHasOperand(
+      conformance,
+      dbEnum.conformanceTag.provisional
+    )
+  ) {
+    return dbEnum.conformanceVal.provisional
+  }
+  return null
+}
+
 exports.parseConformanceFromXML = parseConformanceFromXML
 exports.getOptionalAttributeFromXML = getOptionalAttributeFromXML
+exports.getApiMaturityFromXML = getApiMaturityFromXML
--- a/src-electron/zcl/zcl-loader-silabs.js
+++ b/src-electron/zcl/zcl-loader-silabs.js
@@ -502,7 +502,7 @@ function prepareCluster(cluster, context, isExtension = false) {
       }
       ret.introducedIn = cluster.$.introducedIn
       ret.removedIn = cluster.$.removedIn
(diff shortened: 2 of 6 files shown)


src-electron/validation/conformance-xml-parser.js
```diff
@@ -194,5 +194,34 @@ function getOptionalAttributeFromXML(element, elementType) {
...
+    conformance &&
+    conformEvaluator.checkIfExpressionHasOperand(
+      conformance,
+      dbEnum.conformanceTag.provisional
+    )
+  ) {
+    return dbEnum.conformanceVal.provisional
+  }
+  return null
+}
```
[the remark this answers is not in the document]
brdandu: Believe the issue is being caught by if (element) and if (conformance) checks so this is not really needed and an overkill

brdandu:
The codegen diff failures in the Matter CI builds are as expected and the correct behavior. Validated it with @andreilitvin on this.

src-electron/validation/conformance-xml-parser.js
```diff
@@ -194,5 +194,34 @@ function getOptionalAttributeFromXML(element, elementType) {
...
+ * (including nested conformance tags), return 'provisional'.
+ * Return null when neither source specifies maturity.
+ *
+ * @param {*} element XML element from xml2js
+ * @returns {string|null}
+ */
+function getApiMaturityFromXML(element) {
+  if (element && element.$ && element.$.apiMaturity) {
+    return element.$.apiMaturity
+  }
```
ethanzhouyc, line 211: According to Andrei, apiMaturity should be marked as provisional when either it or the conformance is provisional. I think this logic could be applied before we return apiMaturity as it is if it exists.

Now it's fine as apiMaturity only has deprecated and provisional, but in the future if it can be in conflict with conformance, might cause issues

ethanzhouyc (approved):

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -1587,7 +1587,7 @@ async function processBitmapFields(
           name: item.$.name,
           mask: 1 << itemBit,
           fieldIdentifier: itemBit,
-          apiMaturity: item.$.apiMaturity
+          apiMaturity: conformParser.getApiMaturityFromXML(item)
```
paulr34, line 1608: why deviate from the normal pattern?

paulr34 (approved):

src-electron/validation/conformance-xml-parser.js
brdandu: Still trying to understand if I am missing something. Could you give an example? If apiMaturity is explicitly mentioned then we need not look for conformance, that takes precedence, only if it is not mentioned then we look for conformance.

src-electron/zcl/zcl-loader-silabs.js
brdandu: XML is not clean. Conformance is mentioned properly in all xml elements but apiMaturity is not mentioned the same way for all the xml elements yet. So this is like a fallback to get apiMaturity from conformance if not mentioned explicitly.

PR #1701, opened 2026-04-22
JSDoc db-mapping keys for block helpers

### LINKED ISSUE #915: Update each helper function to include the return map in helper comments
Update the helper function comments with the map of returned elements. This will improve readability and usage of existing helpers. Also, developer will not have to go through more code to understand what all is returned by the helper.

brdandu:
- Attribute keys available for each block helper added to the the block helper function definition

- Github: ZAP #915

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -9072,15 +9072,15 @@ For each element, context contains role, operation, accessModifier.
 Additionally it creates booleans hasRole, hasOperation and hasAccessModifier
 and hasAtLeastOneAccessElement and hasAllAccessElements
 From `exports.map.access` in `src-electron/db/db-mapping.js`:
+- accessModifier
 - operation
 - role
-- accessModifier
 Helper-added:
-- hasRole
-- hasOperation
 - hasAccessModifier
 - hasAllAccessElements
 - hasAtLeastOneAccessElement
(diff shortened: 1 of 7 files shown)


docs/api.md
```diff
@@ -9071,6 +9071,16 @@ Access helper iterates across all the access triplets associated with the elemen
 For each element, context contains role, operation, accessModifier.
 Additionally it creates booleans hasRole, hasOperation and hasAccessModifier
 and hasAtLeastOneAccessElement and hasAllAccessElements
+From `exports.map.access` in `src-electron/db/db-mapping.js`:
```
ethanzhouyc, line 9074: If all the keys are coming from `db-mapping.js`, then adding 'in `src-electron/db/db-mapping.js`' in each comment looks redundant somehow to me

ethanzhouyc (approved):

docs/api.md
brdandu: Agree that it is redundant but in case something does not work then users will know how to back track or where to look for failures in the code. This file is actually auto-generated as part of our commits so is being brought automatically from the function descriptions.

paulr34 (approved):
looks good

PR #1703, opened 2026-04-27
Align Matter/CSA encoding with events and struct fields: capture fabric sensitivity for attributes when <access fabricSensitive=true/> is present.

### LINKED ISSUE #1644: Should attributes support isFabricSensitive?
Attributes [do not currently have an isFabricSensitive property](https://github.com/project-chip/zap/blob/d6ee8315f10bbbb20b3edf94763ae647c65e7270/src-electron/db/db-mapping.js#L90

[Events do, though](https://github.com/project-chip/zap/blob/d6ee8315f10bbbb20b3edf94763ae647c65e7270/src-electron/db/db-mapping.js#L189) - should attributes support this as the spec uses it in attributes? (Access = S)

brdandu:
- Add ATTRIBUTE.IS_FABRIC_SENSITIVE in zap-schema; insert and map in query-loader / db-mapping; include column in attribute SELECTs (query-zcl, query-attribute, query-session-zcl, query-endpoint).
- zcl-loader-silabs: set isFabricSensitive via isAccessFabricSensitive() on parsed attribute access children.
- Extend meta fixture (test1.xml) and gen-meta tests for at2 and HelloEvent isFabricSensitive where applicable.
- Github: ZAP #1644


paulr34 (approved):

dhchandw (approved):

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -445,6 +445,23 @@ function extractAccessIntoArray(xmlElement) {
...
+ * form (not isFabricSensitive on the attribute start tag), e.g.
+ *   <access read="true" readPrivilege="manage" fabricSensitive="true"/>
+ *
+ * @param {*} xmlElement xml2js-parsed element that may contain an `access` array
+ * @returns Whether any access child has fabricSensitive="true"
+ */
+function isAccessFabricSensitive(xmlElement) {
+  if (!('access' in xmlElement)) return false
+  for (const ac of xmlElement.access) {
+    if (ac && ac.$ && ac.$.fabricSensitive == 'true') return true
```
bzbarsky-apple, line 460: Will fabric sensitive attributes always be flagged as fabric scoped in the XML?  In the spec, fabric sensitive implies fabric scoped; that needs to happen in ZAP as well, either on the level of the XML or during import.

brdandu: Do see the xml define `<access fabricScoped="true"/>` separately. Do you want me to add that into the helpers as a separate offering?

bzbarsky-apple: I think as long as the overall pipeline works (e.g. XML has fabricScoped whenever it has fabricSensitive) I don't have strong feelings about the importer.

brdandu: Will merge this PR for now and if we need fabricScoped to be imported/added here as well then I can create another PR for it.

PR #1704, opened 2026-05-04
Add headless .zap validation, UI panel, and CLI improvements

### LINKED ISSUE #1678: Add a mode in whih ZAP enforces that default values in XML are within the min/max range
Right now if the default value of an attribute in XML is outside the min/max range, ZAP warns but does not fail. It would be good to have a way to make it fail (e.g. command-line flag?  SDK config?) that Matter could use for its scripted codegen.

Note that the checking here needs to be careful to not treat the null value as "out of range" if the attribute is nullable.

brdandu:
- Add validate-all session orchestration and GET /validate REST support
- Wire main-process startValidate: merged report, optional file or stdout
- CLI: register logToStdout; support multiple .zap paths (positionals, repeat -i, comma-separated for validate)
- Conformance: optional skip writing SESSION_NOTICE during bulk validate so that it does not go into session notifications on every run
- Validation: Matter root endpoint 0, session-wide duplicate check fix, external storage attribute skip; validate-all null endpoint pre-check
- UI: Validate in ZCL toolbar, side panel (ValidationPanelPage), layout/store
- Docs: add docs/validating-zap-files.md; trim README to link developer docs
- Tests: extend arg and validation tests; fix duplicate-endpoint test data
- Github: ZAP #1678

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ This software is licensed under [Apache 2.0 license](LICENSE.txt).
 
 ## Detailed Developer Documentation
 
+- [Validating `.zap` files](docs/validating-zap-files.md)
 - [ZAP Template Helpers](docs/helpers.md)
 - [ZAP External Template Helpers](docs/external-helpers.md)
 - [ZAP file Extensions](docs/zap-file-extensions.md)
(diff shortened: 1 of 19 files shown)


paulr34 (approved):

dhchandw (approved):
Looks good to me

ethanzhouyc (approved):

PR #1705, opened 2026-05-05
Move cross-endpoint block helpers into helper-shared-config.js and keep per-endpoint / singleton-hybrid helpers in helper-session.js.

### LINKED ISSUE #987: Feedback and cleanup for better developer experience
From @andy31415 in https://csamembers.slack.com/archives/C01B01GH64E/p1680542354402279
Feedback for zap: this is not a good developer experience:
why do seemingly similar loops have different properties (isNullable available in one but not the other)
how does one determine when things filter or not (e.g. by endpoint id)
how does one determine if included/not-included filters are applied
Generally I would expect naming to make it somewhat obvious by convention. I am missing this, so we seem to end up writing templates by trial & error (or at least I am) ... if it seems to work, I assume it works.

Some action items to start with according to me:
- Have all queries use the mapping constructs from db-mapping. Queries should not have individual maps. This creates confusion on the keys present for block helpers associated to attributes, commands, clusters and endpoints.
- Updating the block helper documentation to include all the attributes that are supported for easy reference.
- Need to classify block helpers based on shared configuration across endpoints and unshared configuration across endpoints. We should include this in documentation and also separate the helper files based on this classification
- Need to clean up helper naming. We use user, enabled, included in various ways across our helper naming. Need to come towards a common naming scheme and add that to our documentation.

brdandu:
- Register helper-shared-config in template-engine includedHelpers.
- Classify using SQL GROUP BY: shared queries omit ENDPOINT_IDENTIFIER;  per-endpoint (or singleton-aware hybrid) include it.
- Extend @module JSDoc on both files; regenerate docs/helpers.md.
- Github: ZAP #987

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -136,7 +136,66 @@ across different query files.</p>
 <dd><p>This module contains the API for accessing SDK extensions.</p>
 </dd>
 <dt><a href="#module_Templating API_ user-data specific helpers">Templating API: user-data specific helpers</a></dt>
-<dd><p>This module contains the API for templating. For more detailed instructions, read {@tutorial template-tutorial}</p>
+<dd><p>This module contains the API for templating helpers that operate on the
+<strong>user-data session</strong> with a <strong>per-endpoint (unshared) configuration</strong>
+lens.</p>
+<p>Classification criterion (objective, based on the underlying SQL):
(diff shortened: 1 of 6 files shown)


paulr34 (approved):
looks good

ethanzhouyc (approved):

PR #1713, opened 2026-05-18
fix: clear UC cluster warnings when Studio reports components installed

paulr34:
Before: Comparing apples to oranges. Selected IDs got run through one (broken) extractor; required IDs got passed through untouched. So they were two different string formats and .includes() failed.

After: Run both sides through the same extractUcClusterCode helper before comparing. Both sides reduce to the same canonical short code (e.g. zigbee_basic), so the existing .includes() check now actually matches what it should match.

### DIFF
--- a/src/util/common-mixin.js
+++ b/src/util/common-mixin.js
@@ -298,22 +298,11 @@ export default {
         )
       }
 
-      // If ZAP just installed components for this cluster, skip the check.
-      const installRequested =
-        this.$store.state.zap.studio.installRequestedClusterIds || []
-      if (cluster && installRequested.includes(cluster.id)) {
-        return []
-      }
-
-      const rawSelected =
-        this.$store.state.zap.studio.selectedUcComponents || []
-      const selectedIds = Util.getClusterIdsByUcComponents(rawSelected)
-      const installedNorm = new Set(
-        selectedIds.map((id) => Util.normalizeUcDependencyId(id))
+      const selectedIds = Util.getClusterIdsByUcComponents(
+        this.$store.state.zap.studio.selectedUcComponents
       )
-
       return requiredComponentIdList.filter(
-        (id) => !installedNorm.has(Util.normalizeUcDependencyId(id))
+        (id) => !selectedIds.includes(Util.extractUcClusterCode(id))
       )
     },
 
--- a/src/util/common-mixin.js
+++ b/src/util/common-mixin.js
@@ -298,23 +298,15 @@ export default {
         )
       }
 
-      const studio = this.$store.state.zap.studio
-      const fromTree = Util.getClusterIdsByUcComponents(
-        (studio.ucComponents || []).filter((x) =>
-          Util.isUcComponentEffectivelyInstalled(x)
-        )
+      const selectedIds = Util.getClusterIdsByUcComponents(
+        this.$store.state.zap.studio.selectedUcComponents || []
       )
-      const fromSelected = Util.getClusterIdsByUcComponents(
-        studio.selectedUcComponents || []
-      )
-      const installedNorm = new Set(
-        [...fromTree, ...fromSelected].map((id) =>
-          Util.normalizeUcDependencyId(id)
-        )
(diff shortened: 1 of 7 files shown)


src/util/util.js
```diff
@@ -153,24 +153,31 @@ export function getUcComponents(ucComponentTreeResponse) {
...
+ * Extract a list of cluster ids (in short cluster-code form) from a list of
+ * UC component objects.
+ *
+ * @param {*} ucComponents - an array of UC component objects with `id`
  */
 export function getClusterIdsByUcComponents(ucComponents) {
-  return ucComponents
-    .map((component) => component.id)
-    .map((id) => {
-      if (id.includes('zigbee')) {
```
brdandu, line 168: Getting rid of this worries me. There seems to be a distinction between how we handle cluster to component enablement/disablement between matter and zigbee. Removing this may break that. We do not have unit tests for checking this behavior. You will need to get the ZAP from this PR, add it to Studio and make sure the cluster to component enablement/disablement still works

src/util/common-mixin.js
```diff
@@ -298,12 +298,11 @@ export default {
         )
       }
 
-      let selectedUcComponentIds = Util.getClusterIdsByUcComponents(
```
brdandu, line 301: selectedUcComponentIds name seems more precise and will avoid the additional diff here

brdandu (changes requested):

src/util/util.js
paulr34: Yeah that distinction was actually the bug. Old code reduced zigbee ids to zigbee_basic but kept matter as %extension-matter%matter_level_control, and the required ids on the other side never got normalized at all. So zigbee never matched, matter only matched when both sides happened to keep the prefix, and matter in studiocomproot form returned undefined.

New code runs both sides through the same helper, so they collapse to the same short code. Matter still works, zigbee now works, and the broken matter case is fixed too.

It can't break enable/disable. getClusterIdsByUcComponents is only called from missingUcComponentDependencies. The actual install and uninstall calls go through updateSelectedComponent and a POST to Studio, none of which touches this function.

You're right about the tests though. Added one in a1b8abca that covers all the id formats and asserts matter and zigbee both match against the same comparison the warning uses.

src/util/common-mixin.js
paulr34: I agree

src/util/util.js
paulr34: @brdandu I tested it 
Screenshot 2026-05-20 at 5 12 14 PM
manually

src/util/common-mixin.js
paulr34: made your suggested change

ethanzhouyc (approved):

src/util/util.js
```diff
@@ -153,24 +153,31 @@ export function getUcComponents(ucComponentTreeResponse) {
...
- * Extract a list of cluster id from a list of Uc component id
+ * Returns the short cluster code from a UC component id, regardless of
+ * which prefix format the id uses. Use this when you need to compare ids
+ * that came from different sources.
  *
- * return: a list of ids
- *   id value example:
- *     "zigbee_basic" from "studiocomproot-Zigbee-Cluster_Library-Common-zigbee_basic"
-       "%extension-matter%matter_level_control" from "matter:1.0.0-Matter-Clusters-%extension-matter%matter_level_control"
- * @param {*} ucComponentIds - an array of ids
```
brdandu, line 162: Add an example to the description on how the new functionality works so it is easy to understand

brdandu (approved):
Approving this based on manual tests for clusters -> component toggling passing for both Zigbee and Matter with a ZAP from this PR in Studio V6. Make sure to verify that there was no collision between old and new ZAP during the testing. Remove the existing zap-cli and zap binary in Studio to be sure.

src/util/util.js
paulr34: will do

PR #1717, opened 2026-06-12
move upgraded .zap file into zcl/ subdirectory

paulr34:
During the upgrade flow, write the upgraded .zap file into a zcl/ subdirectory relative to where the original lives, so the path changes from /config/app.zap to /config/zcl/app.zap. The subdirectory is created automatically if it doesn't exist.

findZapFiles now accepts an optional skip-subdirectory parameter so that re-runs don't recurse into the output directory and pick up files that were already upgraded.

TODO: make the subdirectory name configurable (e.g. via apack.json) instead of hardcoded.


src-electron/main-process/startup.js
```diff
@@ -196,27 +196,30 @@ async function noopConvert(resultsFile, logger) {
...
+ * re-runs don't pick up previously upgraded files).
+ *
+ * @param {string} dir - Directory to search.
+ * @param {string|null} skipSubdirectory - Name of a directory to exclude at
+ *   every level of the recursive search (e.g. the upgrade output directory so
+ *   that re-runs don't pick up already-upgraded files).
+ * @returns {string[]} Paths of all .zap files found.
  */
-function findZapFiles(dir) {
+function findZapFiles(dir, skipSubdirectory = null) {
```
brdandu, line 209: Have we investigated why this was not an issue before and is an issue now? We should not be hardcoding things like this. What is the underlying issue?

src-electron/main-process/startup.js
```diff
@@ -233,7 +236,9 @@ function findZapFiles(dir) {
  * @param {*} options
  */
 async function upgradeZapFile(argv, options) {
-  let zapFiles = findZapFiles(argv.d)
+  // TODO: make this configurable (e.g. via apack.json) instead of hardcoded
```
brdandu: Is this the solution? If so then why are we not doing this instead?

paulr34: we should

src-electron/main-process/startup.js
paulr34: I agree with you. We should meet with all teams involved and discuss the best path forward.

At the moment, I see two options:

1.) Merge this PR now to unblock progress, then follow up with a separate PR once we've aligned on the long-term solution.
2.) Hold off on merging, meet with the team to determine where the path should be defined, understand why this behavior started occurring, and then implement a fully agreed-upon solution.

Either way, this PR is available if timing becomes a factor and we need to unblock work quickly.

paulr34: Compromise: put the path in apack.json for now.

That avoids hardcoding it in ZAP and doesn't require SDK changes immediately. ZAP can consume the path from apack.json, and we can revisit the long-term ownership of the path later if needed.

paulr34:
closing this for now

Issue #1691, opened 2026-02-24
Allow additional qualities to be used in code generation


andy31415:
We have use cases where attribute qualities would be useful, like we can avoid some traffic if we know some attributes are fixed (i.e. we can skip sending them post priming)

Based on https://github.com/project-chip/connectedhomeip/blob/master/src/data-model-providers/codegen/CodegenDataModelProvider.cpp#L113 we are already aware that some flags are uknown like quieter reporting, fixed, source attribution, changes ommited.

We generally have qualities on attribute/event/cluster/commant etc, described in section 7.7 of the spec. Here is a list as of today:

| Quality Column | Name | Elements |
| :--- | :--- | :--- |
| A | Source Attribution | attribute, event |
| C | Changes Omitted | attribute data |
| F | Fixed | attribute data |
| I | Singleton | cluster |
| K | Diagnostics | cluster |
| L | Large Message | command |
| N | Non-Volatile | attribute data |
| Q | Quieter Reporting | attribute data |
| S | Scene | attribute |
| T | Atomic | attribute |
| X | Nullable | data fields |

Some of the above are already supported by zap, but many are not.
Beyond that we also have Fabric-scoped, fabric-sensitive and timed interaction access qualities.

PR #1367, opened 2024-07-24
Adding functionality for attribute qualities from the data model spec into ZAP:

brdandu:
- Is_scene_required was already part of the attribute table so updating that with the new xml format
- Also adding change_comitted and persistence quality fields to the attribute table
- Updating all relevant helpers and their associated queries with the above loaded information
- Updating dbmap for attribute and extendedEndpointTypeAttribute with above fields
- Adding tests to check the loading of qualities such as scene, changeOmitted and persistence
- Updating the schema
- Github: ZAP#980

PR #1720, opened 2026-07-01
Load a few more attribute qualities from ZCL XML

paulr34:
Parse the quieter reporting, source attribution and atomic qualities from the <quality> element, store them on the attribute, and add an if_attribute_quality helper so templates can read them. Defaults to false when the XML does not carry them.


src-shared/db-enum.js
```diff
@@ -199,6 +199,16 @@ const storagePolicy = {
 
 exports.storagePolicy = storagePolicy
 
+// Attribute persistence quality (Matter spec section 7.7).
+// Encodes both the "Fixed" (F) and "Non-Volatile" (N) qualities.
+const persistence = {
+  volatile: 'volatile',
+  nonVolatile: 'nonVolatile',
+  fixed: 'fixed'
+}
```
brdandu, line 208: Why is this needed?

src-electron/db/query-loader.js
```diff
@@ -182,13 +182,19 @@ INSERT INTO ATTRIBUTE (
   REMOVED_IN_REF,
   API_MATURITY,
   IS_CHANGE_OMITTED,
-  PERSISTENCE
+  PERSISTENCE,
```
brdandu, line 185: Why do we need persistence?

brdandu (commented):
Add the helper later The xml loader needs to be fixed as well

src-shared/db-enum.js
paulr34: Not new, we already load persistence from XML (#1367). This enum just avoids hard-coding 'fixed' / 'nonVolatile' in if_attribute_quality. Separate from storageOption (RAM/NVM/External).

src-electron/db/query-loader.js
paulr34: Persistence was already added in #1367. This PR just keeps that existing field and adds the new quality columns beside it.

src-shared/db-enum.js
paulr34: here is the ticket which explains https://github.com/project-chip/zap/issues/1691

src-electron/generator/helper-zcl.js
```diff
@@ -1943,6 +1946,72 @@ async function if_is_atomic(type, options) {
...
+ *
+ * @param {*} quality name of the quality to check for
+ * @param {*} options
+ * @returns rendered block content.
+ */
+function if_attribute_quality(quality, options) {
+  let hasQuality = false
+  if (this != null) {
+    switch (quality) {
+      case 'fixed':
```
brdandu, line 1982: Remove all hardcoded strings and extend existing enums or create a new one if necessary.

src-electron/generator/helper-zcl.js
```diff
@@ -1943,6 +1946,72 @@ async function if_is_atomic(type, options) {
...
+ *   {{else}}
+ *     {{label}} does not use quieter reporting
+ *   {{/if_attribute_quality}}
+ * {{/zcl_attributes}}
+ *
+ * @param {*} quality name of the quality to check for
+ * @param {*} options
+ * @returns rendered block content.
+ */
+function if_attribute_quality(quality, options) {
```
brdandu, line 1978: Don't believe a helper like this is needed. wondering if existing attribute helpers having these properties would be good enough.
@andreilitvin @bzbarsky-apple @hasty could you let @paulr34 know if you needed any specific helpers?

src-electron/db/db-mapping.js
```diff
@@ -133,6 +133,9 @@ exports.map = {
       apiMaturity: x.API_MATURITY,
       isChangeOmitted: dbApi.fromDbBool(x.IS_CHANGE_OMITTED),
       persistence: x.PERSISTENCE,
+      isQuieterReporting: dbApi.fromDbBool(x.IS_QUIETER_REPORTING),
+      isSourceAttribution: dbApi.fromDbBool(x.IS_SOURCE_ATTRIBUTION),
+      isAtomic: dbApi.fromDbBool(x.IS_ATOMIC),
```
brdandu, line 138: Add these for endpointTypeAttributeExtended as well

brdandu (changes requested):
You have made changes to the dbmap but you need to update the helper's sql queries to retrieve them and then pass to the dbmap or else they won't be usable

Issue #1712, opened 2026-05-14
Binary package of zap is complaining about electron and node versions


tecimovic:
0. Take binary package of zap.
1. Run zap.
2. See:

```
Expected node versions: v14.x.x,v16.x.x,v18.x.x,v20.x.x
Provided node version: v24.15.0
WARNING: you are using different node version than recommended.
Expected electron version: 17.4.x,18.x.x,24.x.x,27.x.x,31.x.x
Provided electron version: 41.5.1
WARNING: you are using different electron version that recommended.
```

PR #1722, opened 2026-07-16
Skip Node/Electron version warnings for packaged binaries.

paulr34:
Packaged Electron runtimes are fixed at build time, so those checks are not actionable; also allow Node 22/24 for source runs with Electron 41.

### DIFF
--- a/src-electron/util/env.js
+++ b/src-electron/util/env.js
@@ -583,14 +583,53 @@ function isMatchingVersion(versionsArray, providedVersion) {
 }
 
 /**
+ * True when running inside a packaged Electron binary.
+ * Node/Electron versions are fixed at build time there, so version
+ * warnings are not actionable for the user.
+ *
+ * @returns {boolean}
+ */
+function isPackagedElectronApp() {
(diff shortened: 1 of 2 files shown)


ethanzhouyc (approved):

src-electron/util/env.js
```diff
@@ -582,15 +582,47 @@ function isMatchingVersion(versionsArray, providedVersion) {
...
+ * @returns {boolean}
+ */
+function isPackagedElectronApp() {
+  if (process.versions.electron == null) {
+    return false
+  }
+  try {
+    // Under Electron this is the API object; under plain Node the electron
+    // package exports a path string to the binary.
+    const electron = require('electron')
```
brdandu: Move the requires to the top of the file.

brdandu (approved):
https://github.com/project-chip/zap/issues/1712
If the warnings here are no longer show up with the binary executable being opened up from this branch then feel free to merge

src-electron/util/env.js
paulr34: will do

PR #1723, opened 2026-07-20
Allow template control over endpoint_config attribute metadata.

paulr34:
**Background**
ZAP generates the endpoint configuration (attribute tables, cluster tables, etc.) that the Matter SDK compiles into firmware. Today, the generated attribute array comes from a single template helper:

#define GENERATED_ATTRIBUTES {{ endpoint_attribute_list order='default,id,size,type,mask'}}
That helper builds the entire C array as one big string inside JavaScript. Template authors can't reshape it, filter it, or leave parts out. This is a problem for clusters that are moving to code-driven implementations: those clusters manage their own attribute metadata in C++, so having ZAP also emit that same metadata is redundant and wastes flash. There was no way to tell ZAP "skip these clusters' attributes."

**What this PR does**
It gives templates real control over how attribute metadata is generated, without changing any existing behavior.

**1. Iterators for attributes and clusters**

New block helpers usable inside {{#endpoint_config}}:

{{#endpoint_attributes}} — iterate one block per attribute
{{#endpoint_clusters}} — iterate one block per cluster
Templates can now format each row themselves instead of receiving a pre-built string:

#define GENERATED_ATTRIBUTES { \
{{#endpoint_attributes}}
  { {{endpoint_attribute_default}}, {{id}}, {{size}}, {{type}}, {{endpoint_attribute_mask}} }, /* {{clusterName}}.{{name}} */ \
{{/endpoint_attributes}}
}
**2. Richer per-attribute data and formatting helpers**

Each attribute now exposes structured fields (clusterName, clusterId, endpointId, code, storage, isWritable, isNullable, and more), so templates can filter or branch by cluster. Two helpers, {{endpoint_attribute_mask}} and {{endpoint_attribute_default}}, produce the correctly formatted C tokens (the same logic the existing helper used, refactored so it can run per-row).

**3. Ability to omit attribute metadata per cluster**

A new option on {{#endpoint_config}}:

{{#endpoint_config omitAttributeMetadataClusters="Level Control,0x0006"}}
For the listed clusters, the attribute metadata is dropped (attributeCount becomes 0) while the cluster entry itself is preserved. This is the flash-saving path: code-driven clusters can own their attribute tables instead of ZAP duplicating them. Clusters may be matched by name or by code (codes are recommended for names containing /, such as On/Off, because Handlebars treats / specially inside block tags).

**Backward compatibility**
The existing endpoint_attribute_list helper is unchanged, so all current Matter and Zigbee templates continue to work exactly as before.
No new override framework was introduced. Integrators can already replace any built-in helper by declaring a same-named helper in gen-templates.json (helpers load after built-ins and take precedence). This is now documented.
Correctness note
The omit step runs while attributes are being loaded, before ZAP calculates array indexes and counts. This guarantees that the attribute array, its count, and the index/count pointers in the cluster array remain mutually consistent when a cluster's attributes are removed.

**Testing**
Added endpoint-config-control.zapt, a template that exercises the new iterators and verifies the omit option removes a cluster's attributes while keeping the cluster (attrs=0).
All existing byte-exact endpoint_config test suites pass (gen-matter-1, gen-matter-3-1, custom-xml-device-type, endpoint-config).
**Documentation**
docs/external-helpers.md now describes the iterators, the formatting helpers, the omitAttributeMetadataClusters option, and how to override built-in helpers. Generated API docs (docs/api.md, docs/helpers.md) are updated for the new helpers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ This software is licensed under [Apache 2.0 license](LICENSE.txt).
 - [Validating `.zap` files](docs/validating-zap-files.md)
 - [ZAP Template Helpers](docs/helpers.md)
 - [ZAP External Template Helpers](docs/external-helpers.md)
+- [Endpoint configuration generation](docs/endpoint-config-generation.md)
 - [ZAP file Extensions](docs/zap-file-extensions.md)
 - [Upgrading ZAP files with GSDK upgrades](docs/zap-file-upgrade.md)
 - [FAQ/Developer dependencies](docs/faq.md)
--- a/docs/api.md
+++ b/docs/api.md
@@ -109,6 +109,21 @@ across different query files.</p>
 <dt><a href="#module_DB API_ zcl database access">DB API: zcl database access</a></dt>
 <dd><p>This module provides queries for ZCL static queries.</p>
 </dd>
+<dt><a href="#Generator_ endpoint configuration formatters">Generator: endpoint configuration formatters</a></dt>
+<dd><p>Formatting of individual endpoint configuration rows.</p>
+<p>The endpoint configuration is produced in three layers:</p>
+<ol>
+<li>collection: <code>helper-endpointconfig.js</code> turns the session database into
+lists of plain objects (attributes, clusters, commands, ...).</li>
+<li>formatting: this module turns one such object into the C tokens that
+end up in the generated file.</li>
+<li>layout: templates decide where those tokens go, either through the
+aggregate helpers in <code>helper-endpointconfig.js</code> or through the
+iterators in <code>helper-endpointconfig-iterators.js</code>.</li>
+</ol>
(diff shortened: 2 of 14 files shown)


brdandu (changes requested):
This is not what is needed here. We need to completely get rid of this file and not use anything from here. Then create new helpers like the other helper files to produce the same content. It is a pretty complex task. Lets jump into a meeting sometime and we can go over what is required here.

paulr34:
> This is not what is needed here. We need to completely get rid of this file and not use anything from here. Then create new helpers like the other helper files to produce the same content. It is a pretty complex task. Lets jump into a meeting sometime and we can go over what is required here.

I will make a meeting for tomorrow to go over it

PR #1725, opened 2026-07-31
ZAP CLI

paulr34:


docs/make-zap-edit-architecture-pdf
brdandu, line 1: Is this file needed?

docs/zap-edit-architecture.md
```diff
@@ -0,0 +1,912 @@
...
+| See that a control is greyed out                    | Refuse the change in words, and say what the constraint is                |
+| Notice that a value it set came back different      | Never report a write that the next read undoes                            |
+| Know the vocabulary of a data model it has not read | Accept the names it plausibly has, and answer a miss with the near misses |
+| Read a manual page reliably                         | Describe its own surface in a form that parses                            |
+| Undo half of a batch                                | Apply all of a batch or none of it                                        |
+
+None of this is exotic. It is the same discipline good command line tools have
+always had. Agents just make the cost of getting it wrong immediate: a human who
+sees `storage=RAM` printed and `External` in the file will investigate; an agent
+will build ten more edits on top of the false belief.
```
brdandu, line 125: Seems like a lot of things can go wrong here for the agent. Could you point me to some example use cases which can be run through the cli which fail and which pass?

docs/zap-edit-architecture.md
```diff
@@ -0,0 +1,912 @@
...
+   Its storage is fixed at External, which is why the user interface
+   greys the choice out: the value lives in application code rather than in
+   an attribute store. Drop --storage, or pass --storage External.
+```
+
+Three properties make that message usable by a program as well as a person: it
+names the subject precisely, it gives the reason, and it ends with the flags that
+would work. [§10](#10-the-guardrails) lists the full set of rules. The same
+principle governs saving: an edit that could not survive a save is refused with
+the command that would make it possible, never reported as done.
```
brdandu, line 190: Could you give me the cli command to try this locally?

docs/zap-edit-architecture.md
```diff
@@ -0,0 +1,912 @@
...
+
+#### 3.4 Names as the caller has them
+
+An agent takes names from a specification, a header file, or another `.zap` file,
```
brdandu, line 194: This approach sounds very risky. The names and codes should match precisely. If we are using a multi-protocol app we do not want the agent to pick whatever cluster it wants. Even for a single protocol, semantic matching can lead to weird results on what a customer may want to do. We need to just throw an error and ask user the get correct name from the standard xml files.

docs/zap-edit-architecture.md
```diff
@@ -0,0 +1,912 @@
...
+where the same cluster appears as `On/Off`, `ON_OFF`, `onoff` or `0x0006`. All
+four resolve to the same cluster: matching ignores case and punctuation, and a
+number is read as a code. When nothing matches, the answer ranks the plausible
+candidates by edit distance rather than dumping the catalogue, and an ambiguous
+name is reported as ambiguous rather than resolved by luck
+([§9](#9-turning-words-into-rows)).
+
+#### 3.5 All or nothing
```
brdandu, line 202: Could you give me examples on how this works within slc and how this works standalone?
If user does this in a studio project then how does the cluster to component enablement/disablement work?

docs/make-zap-edit-architecture-pdf
paulr34: no ill remove it, thanks for catching that

docs/zap-edit-architecture.md
paulr34: Why do you think a lot could go wrong? What is your reasoning?

docs/zap-edit-architecture.md
paulr34: Just run the help command

paulr34: It is not made for humans. You really should just use an agent.

docs/zap-edit-architecture.md
paulr34: You are assuming a human is better than an agent at understanding ZAP files.

docs/zap-edit-architecture.md
paulr34: Run the help command

docs/zap-edit-architecture.md
paulr34: The GUI gives a human the platform to make ZAP files. The CLI which has the same restrictions as the GUI gives an agent the chance to make ZAP files. Anything that could go wrong would be a deficiency in the AI agent not in the CLI platform that enables it.

PR #1726, opened 2026-08-18
Feat/aai keep default flag

paulr34:
Let attributeAccessInterfaceAttributes entries opt in with { "name": "…", "keepDefault": true } so AAI/External attributes can keep an editable, generated default (e.g. FeatureMap) without taking attribute-store space.

Storage stays External; keepDefault is only a flag. Plain string entries are unchanged.

Wire the flag through load, config/import, validation, endpoint-config generation, and the attribute UI; add test/aai-keep-default.test.js.


src-electron/generator/helper-endpointconfig.js
```diff
@@ -869,6 +870,67 @@ async function determineAttributeDefaultValue(
   return '0x' + 'FF'.repeat(typeSize)
 }
 
+/**
```
brdandu, line 873: Why is helper-endpointConfig.js need a change. We do not need a any changes for generation right? My belief is generation remains the same and only UI needs to allow users to change the attribute when keepsDefault as true

src-electron/zcl/zcl-loader-silabs.js
```diff
@@ -42,6 +42,75 @@ const queryPackageNotification = require('../db/query-package-notification')
...
+ * implementation can read it out of the generated endpoint configuration
+ * instead of holding it itself.
+ *
+ * @param {*} attributeAccessInterfaceAttributes contents of the metadata section
+ * @returns the same clusters, with every entry turned into an object
+ */
+function normalizeAttributeAccessInterfaceAttributes(
+  attributeAccessInterfaceAttributes
+) {
+  let knownKeys = ['name', 'keepDefault']
```
brdandu, line 70: Use existing enums and avoid using explicit strings whenever possible.

src-electron/generator/helper-endpointconfig.js
paulr34: External attributes do not get a generated default today. Codegen throws it away on purpose, and the UI grays the field out so you could not usefully set it anyway. keepDefault is the switch that says keep it External, but still let us set the default and put it in the generated code.

src-electron/zcl/zcl-loader-silabs.js
paulr34: sure, will do

brdandu (approved):
Approving this given that it has been tested for UI and generation

ethanzhouyc (approved):

Issue #346, opened 2021-12-08
Need a way to force a particular storage strategy for some types


bzbarsky-apple:
Specifically:

1. The storage strategy value comes from some logic, not the ZAP UI.
2. The ZAP shows that value, but does not allow changing it.

Matter wants this for list-and-struct-typed attributes.

brdandu:
From @bzbarsky-apple : we did this for lists already, but for structs right now it's having to be done via manual annotations in Matter.  It would be good if ZAP handled this automatically for struct-typed attributes....  Nothing is blocked on this per se, but ergonomically it's pretty painful that your build will fail unless you edit some out-of-band JSON if you add an attribute that is struct-typed

PR #1729, opened 2026-09-14
Force External storage for Matter struct-typed attributes.

paulr34:
Mirror the existing list-type rule so Matter configs do not need a manual attributeAccessInterfaceAttributes entry for every struct.

https://github.com/project-chip/zap/issues/346


ethanzhouyc (approved):

test/resource/old-matter/zcl.json
```diff
@@ -99,6 +99,7 @@
     }
   },
   "listsUseAttributeAccessInterface": true,
+  "structsUseAttributeAccessInterface": true,
   "attributeAccessInterfaceAttributes": {
```
brdandu, line 103: Does this mean all list and struct type attributes are external?

test/zcl-loader.test.js
```diff
@@ -630,6 +630,48 @@ test(
...
+            "SELECT ATTRIBUTE_ID FROM ATTRIBUTE WHERE CLUSTER_REF = ? AND CODE = 0x0004 AND NAME = 'Application'",
+            [applicationBasic.id]
+          )
+        ).ATTRIBUTE_ID
+      )
+      expect(applicationAttr.type).toBe('ApplicationStruct')
+      expect(applicationAttr.isArray).toBe(0)
+      expect(applicationAttr.storagePolicy).toBe(
+        dbEnum.storagePolicy.attributeAccessInterface
+      )
```
brdandu, line 674: Would be great if we can add a cypress test which checks this attribute's storage type to be external in the UI.

brdandu (changes requested):

test/resource/old-matter/zcl.json
paulr34: yes

test/zcl-loader.test.js
paulr34: sure

PR #1730, opened 2026-09-14
Feat/nonvolatile nvm storage

paulr34:
https://github.com/project-chip/zap/issues/367


ethanzhouyc (approved):
Seems a few file changes also exist in #1729, like the files under zcl-builtin/matter. Are these 2 PRs related somehow?

paulr34:
> Seems a few file changes also exist in #1729, like the files under zcl-builtin/matter. Are these 2 PRs related somehow?

Yeah, they are both basically loading in Matter data and setting mandatory storage

brdandu:
Saw @bzbarsky-apple mention `For Matter, we'd want this information to come from the XML and default to NVM for such attributes,` in the issue. Do we need to bring some xml changes here and then use that to say the attribute cannot be RAM based?

paulr34:
> Saw @bzbarsky-apple mention `For Matter, we'd want this information to come from the XML and default to NVM for such attributes,` in the issue. Do we need to bring some xml changes here and then use that to say the attribute cannot be RAM based?

Good point, if for all structs then the boolean in the zcl.json would make sense to me especially since that is how it works for lists. What do you think @bzbarsky-apple ?

Issue #1695, opened 2026-03-10
Handle Obsolete conformance


paulr34:
1. Filter obsolete elements out of the GUI they shouldn't appear in the attribute/command/event configurator views (Client configuration is already disabled)
2. Add a template helper so codegen can skip obsolete elements cleanly:
`{{#attribute_block_helper}}  {{#if obsolete}}  {{else}}    // generate attribute code  {{/if}}{{/attribute_block_helper}}`
4. Add obsolete to the conformance enums in db-enum.js (currently missing from conformanceVal / conformanceTag)
5. Ensure ZCL loader parses obsolete conformance from the data model

brdandu:
Adding more information on this from another conversation. Need to also check the spec for a detailed explanation on this.

Can't any data model element be obsolete?
I would say yes, this can in practice be any element, a field, an attribute, a command, an event... Anything can be marked as obsolete.

Can an element be conditionally obsolete?
No (from a server perspective), if it is obsolete in a specific revision of the spec, it is removed and the ID can not longer be used. But with that said, a client still needs to be able to deal with the fact that older servers can include it.

PR #1731, opened 2026-09-15
Handle obsolete Matter conformance in loader, GUI, and codegen.

paulr34:
Parse <obsoleteConform/> as tag Z, hide obsolete attributes/commands/events from configurator views, and expose an obsolete flag so templates can skip them.

https://github.com/project-chip/zap/issues/1695


src-electron/db/db-mapping.js
```diff
@@ -26,6 +26,14 @@ const dbEnums = require('../../src-shared/db-enum.js')
...
 const conformEvaluator = require('../validation/conformance-expression-evaluator')
 
+/**
+ * True when the stored Matter conformance tag is obsolete (`Z` / `<obsoleteConform/>`).
+ * @param {*} conformance
+ */
+function isObsoleteConformance(conformance) {
+  return conformance === dbEnums.conformanceTag.obsolete
+}
+
```
brdandu, line 36: Don't put this function here. Add it to conformance-expression-evaluator.js

src-electron/generator/helper-zcl.js
```diff
@@ -626,6 +626,7 @@ async function zcl_clusters(options) {
  * - manufacturerCode
  * - mustUseTimedInvoke
  * - name
+ * - obsolete
```
brdandu, line 629: zcl-commands helper still does not have the obsolete access. You need to update update functions like query-command.js#selectAllCommands to get access to obsolete through zcl_commands and you need to do this for all the other helpers here as well. 
Your alternate option which is way better and a lot more scalable where you will not need to update so many of these helpers and use existing infrastructure we have for conformance every where is to add this <obsoleteConform/> to existing conformance and none of the helpers will need to be updated. Will save you a ton of changes here. @ethanzhouyc would that be the correct assumption?

src-electron/generator/helper-zcl.js
```diff
@@ -2039,6 +2042,32 @@ function isClient(side) {
...
+ *   {{#if obsolete}}
+ *   {{else}}
+ *     // generate attribute code
+ *   {{/if}}
+ * {{/zcl_attributes}}
+ *
+ * @param {*} conformance optional conformance string
+ * @returns boolean
+ */
+function isObsolete(conformance) {
```
brdandu, line 2061: This helper is not needed. Will be retrieved from existing helpers or conformance.

src-shared/db-enum.js
```diff
@@ -274,6 +274,7 @@ exports.conformanceTag = {
   provisional: 'P',
   disallowed: 'X',
   deprecated: 'D',
+  obsolete: 'Z',
```
brdandu, line 277: Is obsolete denoted as Z by the spec?

src-electron/generator/helper-zcl.js
brdandu: Functionality wise this should be very similar to disallowConformance

brdandu (changes requested):

src-shared/db-enum.js
paulr34: yes

src-electron/db/db-mapping.js
paulr34: Agreed

src-electron/generator/helper-zcl.js
paulr34: Agreed. Using the existing conformance infrastructure is cleaner and more scalable. <obsoleteConform/> is already parsed into the existing CONFORMANCE field as Z, and the conformance evaluator treats Z as notSupported, similar to disallowConform.

I’ll remove the separate obsolete property and use conformance as the source of truth.

One note: the main zcl_commands helper already has access to conformance because both selectAllCommands and selectCommandsByClusterId select COMMAND.CONFORMANCE. I’ll check the other helper query paths for consistency, but they should not need a separate obsolete field.

src-electron/generator/helper-zcl.js
paulr34: Understood. I’ll remove this helper and use the existing conformance value instead. Obsolete elements will remain represented as Z, and templates can check it using the existing string comparison helper.

PR #1732, opened 2026-09-17
Fix Studio cluster-to-component enablement for shared UC components and add cursor rules for agent to avoid code mashup

brdandu:
- Reload gen-template packages on session open so dependency JSON CRC changes are applied,
- Skip removing a component while another enabled cluster still maps to it.
- Add Cursor agent rules for reuse, layout, requires-at-top, and extending tests.
- JIRA: ZAPP-1742


.cursor/rules/js-requires-at-top.mdc
```diff
@@ -0,0 +1,33 @@
...
+description: Keep require/import statements at the top of JS files
+globs: "**/*.{js,cjs,mjs,vue}"
+alwaysApply: false
+---
+
+# Requires and imports stay at the top
+
+Never add `require(...)`, `import ...`, or dynamic `import()` for module loading in the middle of a function, test body, or after other statements.
+
+## Required pattern
```
paulr34, line 11: I like this, you might want to add it to the PR description/title because it is a separate task than clusters that share components

paulr34 (approved):
We need this..thank you!

.cursor/rules/js-requires-at-top.mdc
brdandu: Changed it to "Fix Studio cluster-to-component enablement for shared UC components and add cursor rules for agent to avoid code mashup- #1732"

.cursor/rules/extend-existing-tests.mdc
```diff
@@ -0,0 +1,32 @@
+---
+description: Extend existing ZAP tests before adding new suites; avoid redundant test code
+alwaysApply: true
```
ethanzhouyc: I like how these files regulate agent runs.

Is this maintainable as we add more .mdc files? With more alwaysApply: true, could every request eventually consume significant extra tokens?

Should we use alwaysApply: false with clear descriptions or globs instead?

ethanzhouyc (approved):

.cursor/rules/extend-existing-tests.mdc
brdandu: It actually depends on the issue at hand. If your prompt hits this and extends an existing test then you could consume less tokens because you may do this again as a separate prompt to fix it if you catch it. However, if your prompt does not really even lead to adding a test then it would look at this text and consume extra tokens for no reason. But good point, have updated the always apply true to false + globs.

